PR #14, opened 2021-04-01
Add esbuild-plugin

mattcompiles:
Current limitations:

- No automatic readable class names in dev
- The `projectRoot` plugin option must be set to get deterministic class name hashes between build systems

If `projectRoot` isn't set, fileScope will fallback to absolute paths. This is fine when building locally, but causes hash differences across different machines which would break caching.


README.md
```diff
@@ -159,6 +160,32 @@ module.exports = {
...
+```js
+const { vanillaExtractPlugin } = require('@vanilla-extract/esbuild-plugin');
+
+require('esbuild').buildSync({
+  entryPoints: ['app.ts'],
+  bundle: true,
+  plugins: [vanillaExtractPlugin({ projectRoot: '...' })],
+  outfile: 'out.js',
+})
+```
```
Mokshit06, line 187: According to [esbuild documentation](https://esbuild.github.io/api/#js-specific-details), we can't use plugins with the synchronous API (`buildSync`) since plugins are asynchronous. So instead of `buildSync`, I think we should use `build` for bundling the code

```suggestion
require('esbuild').build({
  entryPoints: ['app.ts'],
  bundle: true,
  plugins: [vanillaExtractPlugin({ projectRoot: '...' })],
  outfile: 'out.js',
}).catch(() => process.exit(1))
```

mattcompiles: Ahh good catch. I did know that but forgot when writing the docs.

README.md
```diff
@@ -159,6 +160,31 @@ module.exports = {
   ```
 </details>
 
+### esbuild
+
+Current limitations:
+
+- No automatic readable class names in dev
```
markdalgleish, line 167: What do you think about this?

```suggestion
- No automatic readable class names during development. However, you can still manually provide a debug ID as the last argument to functions that generate scoped styles, e.g. `export const className = style({ ... }, 'className');`
```

packages/esbuild-plugin/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@vanilla-extract/esbuild-plugin",
+  "version": "0.0.1",
```
markdalgleish, line 3: Given it hasn't been released yet, should this be v0.0.0?

packages/esbuild-plugin/src/index.ts
```diff
@@ -0,0 +1,201 @@
...
+      build.onLoad({ filter: /\.css\.(js|jsx|ts|tsx)$/ }, async ({ path }) => {
+        const result = await esbuild({
+          entryPoints: [path],
+          metafile: true,
+          bundle: true,
+          external: ['@vanilla-extract', ...externals],
+          platform: 'node',
+          write: false,
+          plugins: [vanillaExtractFilescopePlugin({ projectRoot })],
+          treeShaking: 'ignore-annotations',
```
markdalgleish, line 91: Do you know if this is still required?

markdalgleish:
Is this missing a changeset? Or are you planning to publish in a follow-up PR?

packages/esbuild-plugin/src/index.ts
mattcompiles: Ahh forgot about this. Will check.

mattcompiles:
Packages are auto published if they haven't been published before. So you don't need a changeset for the first release.

packages/esbuild-plugin/src/index.ts
mattcompiles: Removed

packages/esbuild-plugin/package.json
```diff
@@ -0,0 +1,32 @@
...
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/seek-oss/vanilla-extract.git",
+    "directory": "packages/esbuild-plugin"
+  },
+  "author": "SEEK",
+  "license": "MIT",
+  "peerDependencies": {
+    "esbuild": "^0.11.1"
```
markdalgleish, line 18: IMO restrictive peer dependency version ranges mostly just get in the way when dealing with third party libs. Thoughts on making this `>=0.11.1`? This way, if v0.12 comes out, the version range will still match. I realise that we're using experimental esbuild features, but we'll know pretty quickly if this doesn't work in practice, and we'll want to keep up to date anyway.

packages/esbuild-plugin/src/index.ts
```diff
@@ -0,0 +1,214 @@
...
+              cssByFileScope.set(fileScope, fileScopeCss);
+            }
+          },
+          registerClassName: (className) => {
+            localClassNames.add(className);
+          },
+          onEndFileScope: () => {},
+        };
+
+        setAdapter(cssAdapter);
```
markdalgleish, line 133: Do we need this line given we're setting the adapter in the string of code below?

packages/esbuild-plugin/src/index.ts
```diff
@@ -0,0 +1,214 @@
...
+        return {
+          contents,
+          loader: 'js',
+        };
+      });
+    },
+  };
+}
+
+const stringifyExports = (value: any) =>
```
markdalgleish, line 167: More just marking this for future conversation, but this makes me think it might be good to have a more first-class integration API so stuff like this is handled for you.

README.md
```diff
@@ -159,6 +160,31 @@ module.exports = {
...
+```js
+const { vanillaExtractPlugin } = require('@vanilla-extract/esbuild-plugin');
+
+require('esbuild').build({
+  entryPoints: ['app.ts'],
+  bundle: true,
+  plugins: [vanillaExtractPlugin({ projectRoot: '...' })],
+  outfile: 'out.js',
+}).catch(() => process.exit(1))
+
```
markdalgleish, line 187: Missing backticks here, it's messed up the readme formatting after this point.

markdalgleish (approved):
A few comments, but nothing to hold up an approval. This is super exciting 🔥

PR #23, opened 2021-04-10
Add note about CSS handling and URLs to treat migration guide

### LINKED ISSUE #22: Treat Migration Guide: issue with importing images
I just went ahead to migrate our 61 `.treat.js` files in our project to vanilla-extract and it all went really smoothly. Good job for making the migration so painless!
Of course we're using none of the css variables feature yet, but the nice thing is that we can now progressively migrate it piece by piece

One thing that broke was this though:

```js
import stepBgImg from "./imgs/steps-bg.svg";
const container = style({
  backgroundImage: `url("${stepBgImg}")`
})
```

it showed a rather long and unreadable error message before realising that vanilla-extract is probably more clever than I've expected. So I tried doing

```js
const container = style({
  backgroundImage: `url('./imgs/steps-bg.svg')`,
})
```

which worked!

So maybe this is worth mentioning in the migration guide?

Cheers!

markdalgleish:
Closes #22.

### DIFF
--- a/docs/treat-migration-guide.md
+++ b/docs/treat-migration-guide.md
@@ -41,24 +41,29 @@ module.exports = {
   module: {
     rules: [
       {
-        test: (file) => file.endsWith('.vanilla.css'),
-        use: [
-          MiniCssExtractPlugin.loader,
+        test: /\.css$/i,
+        oneOf: [
           {
-            loader: 'css-loader',
-            options: { url: false }
+            test: /\.vanilla\.css$/i,
+            use: [
+              MiniCssExtractPlugin.loader,
+              {
+                loader: 'css-loader',
+                options: { url: false }
+              }
+            ]
+          },
+          {
+            use: [MiniCssExtractPlugin.loader, 'css-loader']
           }
         ]
-      },
-      {
-        test: (file) =>
-          file.endsWith('.css') &&
-          !file.endsWith('.vanilla.css'),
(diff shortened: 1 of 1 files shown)


docs/treat-migration-guide.md
```diff
@@ -18,12 +18,50 @@ The Babel plugin was optional in treat, but it's required in vanilla-extract.
...
+In treat, we set css-loader's [`url` option](https://webpack.js.org/loaders/css-loader/#url) to `false`. This was to ensure that JavaScript import statements were always used for assets (e.g. `import logoUrl from './logo.png'`) rather than allowing the CSS to create implicit imports (e.g. `background: "url('./logo.png')"`).
+
+If you want to reinstate treat's approach to asset imports without affecting other CSS files, you can configure css-loader seprately for `*.vanilla.css` files. For example:
+
+```ts
+module.exports = {
+  module: {
+    rules: [
+      {
+        test: (file) => file.endsWith('.vanilla.css'),
```
mattcompiles, line 44: Any reason you're not using a regex? Also maybe a `oneOf` would be a better examples?

markdalgleish: Using `oneOf` is a good idea, simplifies things a lot 👍

markdalgleish: Pushed an update, let me know what you think.

mattcompiles (approved):

PR #24, opened 2021-04-10
Add note to readme about scoped selectors

markdalgleish:


README.md
```diff
@@ -255,6 +255,8 @@ export const childClass = style({
 });
 ```
 
+> 💡 Each `style` block can only target a single element. To enforce this, all selectors must target the `&` character which is a reference to the current element, e.g. `'&:hover:active'`. If you want to target nodes globally within the current element, you should use [`globalStyle`](#globalstyle) instead.
```
mattcompiles, line 258: Feel like we should provide reasoning for this?

markdalgleish: Added some more detail, let me know what you think.

mattcompiles (approved):

PR #39, opened 2021-04-22
Shorten identifiers by removing leading underscores

markdalgleish:
We achieve this by using alpha characters for the ref counter (e.g. `0` is `'a'`, `1` is `'b'`, `26` is `'aa'`) and placing it _before_ the hash, which means that a class like `_7skkcy0` can now be shortened to `a7skkcy`.


packages/css/src/fileScope.ts
```diff
@@ -29,6 +29,24 @@ export function getFileScope(): FileScope {
   return fileScopes[0];
 }
 
+const startingCharCode = 'a'.charCodeAt(0);
+
+export function refCountToAlpha(number: number) {
```
mattcompiles, line 34: Maybe move to a separate file

markdalgleish:
As we suspected after talking this through further, while this results in smaller _uncompressed_ CSS, it actually compresses _worse_ since the ref counter interrupts a repeating pattern.

PR #43, opened 2021-04-23
Theme tokens type update

huijiewei:
add number

### DIFF
--- a/packages/css/src/types.ts
+++ b/packages/css/src/types.ts
@@ -94,7 +94,7 @@ export interface Adapter {
 }
 
 export type Tokens = {
-  [key: string]: string | Tokens;
+  [key: string]: string | number | Tokens;
 };
 
 export type ThemeVars<ThemeContract extends Contract> = MapLeafNodes<


mattcompiles:
Hi, thanks for the PR. 

However, only allowing `string` is an explicit decision due to the nature of CSS Variables. Almost all of the time your variables will need a unit (e.g. `em` or `px`) to work as expected when consumed in styles. 

If you feel like there's something we're missing here feel free to open up a discussion topic to discuss it 😄

huijiewei:
Thanks, but fontWeight or lineHeight should set be number

markdalgleish:
Unitless numbers can still be provided as strings, e.g. `myFontWeight: '400'`. The idea is that it's forcing you to be explicit that it's unitless.

markdalgleish:
The reason we do this for variables and _don't_ do this for styles is that we have no idea where a variable will be used, so there's no way for us to infer whether it needs a unit applied or not.

maraisr:
The grid properties do not work w/o the literal number eg:

grid-column-start: "1"; ❌
grid-column-start: 1; ✅

cc @markdalgleish @mattcompiles

madou:
That's still `'1'` though.

If you wanted the quotes I'm sure you'd need to add them, like:

`"1"`

maraisr:
have you tried this? `'1'` doesnt work: https://codepen.io/marais/pen/wvgOwdx 

cc @madou

michaeltaranto:
To be clear, this discussion is about the input structure, which accepts css as an object with string values. The resulting values won't have any quotes.

For example, `{ gridColumnStart: '1' }` will be output as `grid-column-start: 1`. Same goes for  `{ display: 'block' }`, the quotes don't end up in the CSS.

maraisr:
seems fair! Buncha legends you folk.

huijiewei:
The point is I don't like it  ```fontWeight: (themeVars.fontWeights.medium as unknown) as number```

mattcompiles:
@huijiewei I agree with you that this is an issue. All properties should support CSS Variables as input without casting.

mattcompiles:
@huijiewei #50 Should resolve this issue 👍

PR #52, opened 2021-04-27
Rename 'createThemeVars' to 'createThemeContract'

markdalgleish:


README.md
```diff
@@ -381,19 +389,23 @@ export const themeVars = createGlobalTheme(':root', {
 
 > 💡 All theme variants must provide a value for every variable or it’s a type error.
 
-### createThemeVars
+### createThemeContract
+
+Creates a contract for themes to implement.
 
-Creates a collection of CSS Variables without coupling them to a specific theme variant.
+**This function must be called within a `.css.ts` context, otherwise variable names will be mismatched between themes.**
```
mattcompiles, line 396: Feel like this reads like this function is special and other don't need to be in a `css.ts` file. How about:
```suggestion
**Ensure this function is called within a `.css.ts` context, otherwise variable names will be mismatched between themes.**
```

mattcompiles (approved):

PR #62, opened 2021-04-29
Catch theme inconsistencies

loliver:
The error thrown when a theme doesn't quite abide by a contract doesn't really give any pointers to what is causing the issue(s).

Adding this shows you exactly where in the contract is wrong and it's simply a matter of figuring out which theme has the extra or missing tokens and correcting it.

Would be nice to be able to log the name of the theme here for more direct pointers, but unsure how to accomplish it.


mattcompiles:
Hey Bud.

I agree that the error experience needs some love here and it's something we're looking into. However I'm not sure just logging a warning and moving on is a good idea. This scenario is an error and should be handled. We'll aim to create a much more descriptive error here though.

loliver:
No worries just wanted to highlight was was effective for me in debugging my themes, the reason I chose logging over throwing was I had a few hundred of them and being able to see large chunks allowed me to more easily diagnose the larger issues.

mattcompiles:
Thanks for highlighting this. Closing this in favour of #84

PR #63, opened 2021-04-29
Sprinkles: Allow theme vars in atomic properties

fnky:
This change makes it possible to pass theme vars from `createThemeContract`/`createTheme` to be passed to atomic styles properties.

### DIFF
--- a/packages/sprinkles/src/index.ts
+++ b/packages/sprinkles/src/index.ts
@@ -1,10 +1,4 @@
-import { style } from '@vanilla-extract/css';
-import type { Contract, MapLeafNodes } from '@vanilla-extract/private';
-import type * as CSS from 'csstype';
-
-export type CSSVarFunction =
-  | `var(--${string})`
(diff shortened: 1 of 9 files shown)


packages/sprinkles/src/index.ts
```diff
@@ -1,6 +1,11 @@
 import { style } from '@vanilla-extract/css';
+import type { Contract, MapLeafNodes } from '@vanilla-extract/private';
 import type * as CSS from 'csstype';
 
+export type CSSVarFunction =
```
fnky, line 5: Not sure if this should be moved to the `private` package.

mattcompiles:
Hey, Thanks for the PR. We decided to export a first class `CSSProperties` type from `@vanilla-extract/css` and use that in sprinkles. Pushed some updates to your PR to match that.

markdalgleish (approved):

fnky:
@mattcompiles Yeah that makes more sense, thanks!

PR #67, opened 2021-04-30
Remove CodeSandbox links

markdalgleish:
Just until we get the type checking sorted out.


mattcompiles (approved):

markdalgleish:
Decided it's better than nothing, even if the editing experience currently isn't as good as it is locally.

PR #68, opened 2021-04-30
Compile .css.ts files with ts loader

jahredhope:
[By default](https://esbuild.github.io/content-types/#javascript) esbuild will use the `js` loader for any content returned from an `onLoad` callback.

This change attempts to detect whether the file being loaded ends in `.ts` or `.tsx` and switches to use the ts loader as documented here:

https://esbuild.github.io/content-types/#typescript

Files that are not detected as `.ts` files will use the current default behaviour of using the `js` loader.

**Example error:**
```
> src/App.css.ts:9:5: error: Expected ";" but found "Hello"
    9 │ type Hello = string;
      ╵      ~~~~~
```

### DIFF
--- a/.changeset/sweet-turkeys-rhyme.md
+++ b/.changeset/sweet-turkeys-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/esbuild-plugin': patch
+---
+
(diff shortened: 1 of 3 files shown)


mattcompiles (approved):
Nice one bud. We hit this the other day but didn't figure it out. 👍

PR #70, opened 2021-04-30
Add Vite plugin

mattcompiles:
Also create shared integration package for share code between bundler integrations.


markdalgleish (changes requested):
👏🎉

markdalgleish (approved):

fnky:
Wow, looks a lot better than my attempt. I'll try it out, since I already have a Vite project set up and currently missing the crucial CSS file for vars.

fnky:
Ok, I tested this out and it seems to work, albeit a few hiccups which are likely unrelated to Vite:

The approach as [documented in Sprinkles](https://github.com/seek-oss/vanilla-extract/tree/master/packages/sprinkles#setup) where `vars.css.ts` and `atoms.css.ts` both import each other (cyclic dependency) didn't work and resulted in the following errors:

```
[vite] Internal server error: atoms is not a function
```

and if `vars.css.ts` (where the first error originated from) is imported in a module (e.g. a component) the following error is displayed:

```
[vite] Internal server error: Cannot read property 'color' of undefined
```

The solution is to split it up further with e.g. a `style.css.ts` that imports for example `atoms.css.ts` or `vars.css.ts`, but not both at the same time to avoid the cyclic dependency.

---

Another issue is that the generated CSS atoms can be so large that the resulting base64 passed in the `source=` query param to the module, results in a `431: Request Header Fields Too Large` HTTP error.

This could be documented to avoid too many atoms being declared, but I also feel like it kind of defeats the purpose of the library, that it has this limitation.

---

These issues are likely not a Vite-specific problem and should probably be its own issue.

mattcompiles:
@fnky Thanks for the feedback. Which example are you referring to in sprinkles that requires cyclical dependencies? 

The URL too long issue is a problem though. Will look into another way to forward that information through.

fnky:
@mattcompiles Sorry, maybe I had it confused while reading the docs on that. Although, I think it's a common pitfall and would be great to document this, and perhaps show proper error message if there's any cyclic dependencies.

README.md
```diff
@@ -197,6 +198,29 @@ require('esbuild').build({
 
 > Please note: There are currently no automatic readable class names during development. However, you can still manually provide a debug ID as the last argument to functions that generate scoped styles, e.g. `export const className = style({ ... }, 'className');`
 
+### Vite
+
+> Warning: Currently the Vite plugin doesn't rebuild files when dependent files change, e.g. updating `theme.css.ts` should rebuild `styles.css.ts` which imports `theme.css.ts`. This is a limitation in the Vite Plugin API that will hopefully be resolved soon. You can track the Vite issue here. https://github.com/vitejs/vite/issues/3216
```
markdalgleish, line 203:
```suggestion
> Warning: Currently the Vite plugin doesn't rebuild files when dependent files change, e.g. updating `theme.css.ts` should rebuild `styles.css.ts` which imports `theme.css.ts`. This is a limitation in the Vite Plugin API that will hopefully be resolved soon. You can track the Vite issue here: https://github.com/vitejs/vite/issues/3216
```

PR #71, opened 2021-04-30
CSS: Accept number values for theme vars and contracts

fnky:
This change adds the ability to pass numeric values to theme variables/contracts.


mattcompiles:
As mentioned [here](https://github.com/seek-oss/vanilla-extract/pull/43#issuecomment-826427488), it is an explicit choice to not allow numbers. We probably need to document this better as it's come up a few times.

fnky:
@mattcompiles Alright, no problem. I expected it to work a similarly to other CSS-in-JS libraries but it makes sense since CSS variables is a first-class concept in vanilla-extract.

PR #110, opened 2021-05-09
Add snowpack-plugin

mxmul:
hey - I'm really liking this library so far!

I wanted to try it out in a site that's built with Snowpack, and ended up writing a basic Snowpack plugin to load `.css.ts` files. It's currently published as [@mxmul/vanilla-extract-snowpack-plugin](https://www.npmjs.com/package/@mxmul/vanilla-extract-snowpack-plugin), but I think it would be great to merge upstream.

There's a [demo app on code sandbox](https://codesandbox.io/s/github/mxmul/vanilla-extract/tree/snowpack-plugin-fork/examples/snowpack?file=/src/App.tsx) to demonstrate it working (note: HMR also works well locally but not in CSB, because its unable to open a websocket).


packages/snowpack-plugin/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+        addImportsToMap(filePath, dependency);
+      });
+
+      let css;
+      const js = processVanillaFile({
+        source,
+        filePath,
+        outputCss: !isSSR,
+        serializeVirtualCssPath({ base64Source, fileScope }) {
+          const cssUrl = `${path.join(cwd, fileScope.filePath)}.css`;
```
mattcompiles, line 57: What's then reasoning behind using `fileScope.filePath` over the `fileName` param that's passed? FileScope filePath's are relative to their package, ignoring that could lead to clashing file names. 

Also we always make CSS created by vanilla-extract end with `.vanilla.css`, for consistency and it allows you to target that CSS directly with bundler rules if need be. Understand that may not be that relevant for snowpack though.

packages/snowpack-plugin/src/index.ts
```diff
@@ -0,0 +1,84 @@
...
+    name: '@vanilla-extract/snowpack-plugin',
+    resolve: {
+      input: ['.css.ts'],
+      output: ['.js', '.css'],
+    },
+    onChange({ filePath }) {
+      if (importedByMap.has(filePath)) {
+        const importedBy = importedByMap.get(filePath);
+        importedByMap.delete(filePath);
+        for (const importerFilePath of importedBy) {
```
mattcompiles, line 33: `importedFilePath` ?

mattcompiles (commented):
Hey Bud 👋 
Thanks for the PR. You've done an amazing job without any docs to follow. Looks pretty good outside of a couple of things.
- Why the fileScope changes?
- Probably need to add some tests before releasing. 
Let me know if that's something you'd like to do or I could help out.

packages/snowpack-plugin/src/index.ts
mxmul: I'm using Snowpack's [multi-file building](https://www.snowpack.dev/guides/plugins#example%3A-multi-file-building) to emit a `.css` and `.js` file for each `.css.ts`; it seems like the idiomatic approach for this sort of transformation in Snowpack, but it doesn't give you control over the emitted file names. They're always emitted as `${sourceFileName}.js` and `${sourceFileName}.css`. This is why I avoided using `fileName` – the generated name doesn't align exactly with the filenames that Snowpack is emitting.

I initially tried to closely follow the approach of the Vite plugin, but Snowpack doesn't seem to support virtual files in the same way; it seems like the dev server can transform files on disk (and even emit multiple files from one source), but not serve a file that doesn't map to something on disk.

mxmul:
> * Why the fileScope changes?

I was testing this in a project that doesn't set a "name" field in package.json, and it was generating paths including a stringified `undefined`. Like "../../undefined/styles.vanilla.css"

packages/snowpack-plugin/src/index.ts
mattcompiles: Interesting. 🤔 

This may be an issue as two different `.css.ts` files may emit the same CSS. For example:
`theme.css.ts` -> `theme.vanilla.css`
`styles.css.ts` (which imports the theme contract) -> `styles.vanilla.css` & `theme.vanilla.css`

The bundler needs to dedupe the two `theme.vanilla.css` files to ensure the CSS doesn't get created twice and potentially introduce a CSS ordering bug. The idea is the CSS imports should be deterministic, no matter which `.css.ts` entrypoint they came from.

Any ideas on how to solve this from a snowpack persepctive?

mxmul: I don't think this is an issue. With this approach, `theme.vanilla.css` only gets created once when `theme.css.ts` is loaded (although it's named differently by snowpack: `theme.css.ts.css`).

I've tried building & bundling my example app, and it creates a single stylesheet with no duplication:

## Inputs
**App.tsx**
```js
import { themeClass } from './theme.css.js';
import { exampleStyle } from './styles.css.js';
```

**theme.css.ts**
```jsx
import { createTheme } from '@vanilla-extract/css';

export const [themeClass, vars] = createTheme({
  color: {
    brand: 'orange',
    blah: 'teal',
  },
  font: {
    body: 'arial',
  },
});
```

**theme.css.ts**
```jsx
import { vars } from './theme.css.js';

export const exampleStyle = style({
  backgroundColor: vars.color.blah,
  fontFamily: vars.font.body,
  color: 'white',
  padding: 10,
});
```

## Output
**build/dist/index.css**
```css
/* build/dist/styles.css.ts.css */
.g7vce90 {
  background-color: var(--zpioyc2);
  font-family: var(--zpioyc3);
  color: white;
  padding: 10px;
}

/* build/dist/theme.css.ts.css */
.zpioyc0 {
  --zpioyc1: orange;
  --zpioyc2: teal;
  --zpioyc3: arial;
}
```

mxmul: Although I did notice that these rules would be defined in a different order by the other integrations (e.g. theme.vanilla.css should be loaded before styles.vanilla.css). That seems tricky/impossible to avoid without a change to Snowpack, since [they inject the import a JS file's associated CSS at the top of the file](https://github.com/snowpackjs/snowpack/blob/680272eb19c43ef88f9d83e436e7131444c53407/snowpack/src/build/file-builder.ts#L125).

This didn't seem to matter in my little demo apps, but I could see this leading to weird specificity bugs.

mattcompiles: I still think there's an issue here. The `serializeVirtualCssPath` hook can run multiple times per `.css.ts` file as it runs once per fileScope. This implementation would end up overwriting the CSS with the last one to call it. In your example, what happens if you use `createGlobalTheme` and don't have to import the `themeClass` into your app.

mxmul: > The serializeVirtualCssPath hook can run multiple times per .css.ts file as it runs once per fileScope. This implementation would end up overwriting the CSS with the last one to call it.

It's true that the function is called multiple times per source file, but we shouldn't actually emit CSS multiple times. E.g. `serializeVirtualCssPath` will be called with`theme.css.ts` twice; once when the file is loaded and again when `style.css.ts` is loaded. The second call is only used to generate an import declaration though. With this conditional, we're only emitting CSS contents for the corresponding `.vanilla.css` file for the `.css.ts` file snowpack is loading (and each `.css.ts` file is only loaded once by snowpack):

```js
if (cssUrl === `${filePath}.css`) {
  css = getSourceFromVirtualCssFile(`?source=${base64Source}`).source;
  ...
}
```
> In your example, what happens if you use createGlobalTheme and don't have to import the themeClass into your app.

It seems to work fine. This is the bundled stylesheet:
```css
/* build/dist/styles.css.ts.css */
.g7vce90 {
  background-color: var(--zpioyc0);
  font-family: var(--zpioyc1);
  color: white;
  padding: 10px;
}

/* build/dist/theme.css.ts.css */
:root {
  --zpioyc0: rebeccapurple;
  --zpioyc1: arial;
}
```

mattcompiles: OK interesting. Snowpack is working quite differently here than I would expect. I wasn't expecting snowpack to load `.css.ts` files that are only used at build time. Seems that's not the case. Your approach seem valid considering the way it's working though. Sorry it took me a minute to catch-up 😅 . 

To get this merged we'll need to add tests so we know we're not breaking it going forward. Most of the work is getting a `snowpack` [startFixture](https://github.com/seek-oss/vanilla-extract/blob/master/test-helpers/src/startFixture/index.ts) script working so we can run our test suite against it. Also updating the README/site with the relevant setup info for snowpack.

packages/integration/src/processVanillaFile.ts
```diff
@@ -82,21 +83,21 @@ export function processVanillaFile({
   const cssImports = [];
 
   for (const [serialisedFileScope, fileScopeCss] of cssByFileScope) {
-    const filescope = parseFileScope(serialisedFileScope);
```
mattcompiles, line 85: You'll need to add a changeset (`yarn changeset`) to make sure the `integration` package is released with these changes.

mattcompiles:
@mxmul Just saw your comment about the M1 puppeteer issue. Happy to help out when I get a chance. However feel free to ignore the tests and just work on the `startFixture` script. If you get that working running `yarn start-fixture [fixture-name] --type=snowpack` should be able to start the fixtures locally. Test are easy if you get that sorted.

mattcompiles:
Hey @mxmul 👋,

I had a crack at getting the start-fixture script working. Managed to get snowpack started and the plugin registered but the app doesn't work. Not really sure what's wrong. Here's the code if it helps.

https://github.com/seek-oss/vanilla-extract/compare/snowpack-plugin?expand=1

Running `yarn start-fixture themed --type=snowpack` will start it up.

mxmul:
@mattcompiles The problem is in how we import the `.css.ts` files in the fixture. When Snowpack sees an import like:
```
import { theme, altTheme, responsiveTheme, vars } from './themes.css';
```

its [module resolution algorithm short-circuits](https://github.com/snowpackjs/snowpack/blob/7291daa9b237398ded16cf856111ece85aef2e6c/snowpack/src/build/import-resolver.ts#L36), and it assumes there must be a `themes.css` file on disk.

This is arguably a bug in Snowpack, but you can work around it by explicitly using a `.css.ts` in the import path  (e.g. ` ./themes.css.ts`). The approach works for Snowpack, but TS [considers it an error](https://github.com/Microsoft/TypeScript/issues/27481#issuecomment-745750281), so in my example app I had been writing the imports as `./themes.css.js`.

mxmul:
@mattcompiles this is ready for another look. My fix to snowpack has landed, and after bumping snowpack your fixtures are loading properly now.

mattcompiles (approved):
Massive effort getting this done mate. Even patching snowpack itself in the process! 🍻 👍

PR #121, opened 2021-05-11
Fix development identifiers for Windows paths

Brendan-csel:
Identifiers were not handling backslash as a path separator on Windows.

This was causing class names like `src\system\ContentBlock\ContentBlock_root__163xsf10` and preventing successful builds.


mattcompiles:
Hey Bud 👋 

Thanks for the PR. Just need to add a changeset before merging. You can do that by running `yarn changeset` and following the prompts.

Issue #122, opened 2021-05-11
Files starting with letter 't' breaks vite dev server on Windows


Saartje87:
### Describe the bug

Running vite on a windows machine and having a file starting with the letter 't' result in 404. In my example I'm using theme.css.ts, on mac this all works fine. However running the same example on Windows results in the following virtual file not found "http://localhost:3000/@id/project/srcheme.css.ts.vanilla.css?hash=xxx". The problem here is tab \t is somewhere being cast to a tab, thus the correct url should be `src/theme.css` not `srcheme.css`.  (Files starting with 's' don't cause any problems.)

I tried to debug (and fix) the issue, but thinking this is not a wonderfull solution 😅 

How I managed to fix it 🙈 

Replacing
https://github.com/seek-oss/vanilla-extract/blob/master/packages/vite-plugin/src/index.ts#L23
`const { fileName, source } = getSourceFromVirtualCssFile(id);`

With
`const { fileName, source } = getSourceFromVirtualCssFile(id.replace(/\t/g, '\/t');`

Happy to make a pr and follow advice.

### Link to reproduction
Got an example repo, run `yarn` then `yarn start` on windows platform.
https://bitbucket.org/Saartje87/vanilla-extract/src/master/

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

```node
  System:
    OS: Windows 10 10.0.19043
    CPU: (16) x64 Intel(R) Core(TM) i7-10870H CPU @ 2.20GHz
    Memory: 5.90 GB / 15.83 GB
  Binaries:
    Node: 14.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.10 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 7.7.5 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.964.0), Chromium (90.0.818.56)
    Internet Explorer: 11.0.19041.1
  npmPackages:
    @vanilla-extract/css: ^0.4.3 => 0.4.3
    @vanilla-extract/sprinkles: ^0.2.2 => 0.2.2
    @vanilla-extract/vite-plugin: ^0.1.1 => 0.1.1
[...]

PR #149, opened 2021-05-18
ci: run ci on windows agents as well

maraisr:
Might help catch issues like #122


markdalgleish:
We might come back to this when we do some more Windows dev work.

PR #152, opened 2021-05-19
Added ::-webkit-scrollbar simple pseudo

Saartje87:


markdalgleish:
Thanks for the PR! While you're in this area, would you be able to add the rest of the webkit scrollbar pseudos? https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar

Saartje87:
Added!

mattcompiles (approved):
Thanks for the PR

PR #155, opened 2021-05-21
Change sprinkles readme so example works together

samrobbins85:
The readme for sprinkles uses `blue50`, whereas the declaration uses `blue-50` and the gray declaration doesn't exist. While this is just an example so not too important, it helps with initial setup that you can copy the code provided and everything will work.


markdalgleish (approved):
Thank you!

PR #181, opened 2021-06-03
webpack-plugin: Skip child compile when within the child compiler

mattcompiles:


packages/webpack-plugin/src/compiler.ts
```diff
@@ -127,15 +122,15 @@ function compileVanillaSource(
           library: {
             type: 'commonjs2',
           },
-          import: [`!!${request}`],
+          import: [loader.resourcePath],
```
mattcompiles, line 125: This change is unrelated to the fix but is an improvement that we couldn't make before due to the old approach.

michaeltaranto (approved):

PR #183, opened 2021-06-03
Add processCss callback option for esbuild plugin

Brendan-csel:
Add `processCss` plugin option to allow further processing of css while bundling.

**Example for postcss with autoprefixer:**

```js
// build.js
const { vanillaExtractPlugin } = require('@vanilla-extract/esbuild-plugin');
const postcss = require('postcss');
const autoprefixer = require('autoprefixer');

require('esbuild').build({
  entryPoints: ['app.ts'],
  bundle: true,
  plugins: [vanillaExtractPlugin({

    processCss: async (css) => {
      return (await postcss([autoprefixer])
        .process(css, { from: undefined /* suppress source map warning */ })
      ).css
    }

  })],
  outfile: 'out.js',
}).catch(() => process.exit(1))
```

### DIFF
--- a/.changeset/curly-bats-allow.md
+++ b/.changeset/curly-bats-allow.md
@@ -1,5 +1,5 @@
 ---
-'@vanilla-extract/esbuild-plugin': patch
+'@vanilla-extract/esbuild-plugin': minor
 ---
 
 Add `processCss` plugin option to allow further processing of css while bundling.
--- a/.changeset/curly-bats-allow.md
+++ b/.changeset/curly-bats-allow.md
@@ -2,7 +2,7 @@
 '@vanilla-extract/esbuild-plugin': minor
 ---
 
-Add `processCss` plugin option to allow further processing of css while bundling.
+Add `processCss` plugin option to allow further processing of CSS while bundling.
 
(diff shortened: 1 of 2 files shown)


.changeset/curly-bats-allow.md
```diff
@@ -0,0 +1,28 @@
+---
+'@vanilla-extract/esbuild-plugin': patch
```
mattcompiles, line 2:
```suggestion
'@vanilla-extract/esbuild-plugin': minor
```

.changeset/curly-bats-allow.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+Add `processCss` plugin option to allow further processing of css while bundling.
```
mattcompiles, line 5:
```suggestion
Add `processCss` plugin option to allow further processing of CSS while bundling.
```

mattcompiles (commented):
Thanks for the PR mate. Agree this is a reasonable API seeing as `esbuild` doesn't support this kind of use-case ATM. Just a couple of small changes before merging.

mattcompiles (approved):

PR #185, opened 2021-06-04
esbuild-plugin: Document processCss option

mattcompiles:


.changeset/curly-bats-allow.md
```diff
@@ -7,22 +7,28 @@ Add `processCss` plugin option to allow further processing of CSS while bundling
...
-  entryPoints: ['app.ts'],
-  bundle: true,
-  plugins: [vanillaExtractPlugin({
+async function processCss(css) {
+    const result = await postcss([autoprefixer]).process(css, {
+        from: undefined /* suppress source map warning */,
+    });
+    
+    return result.css;
+}
```
michaeltaranto, line 20:
```suggestion
async function processCss(css) {
  const result = await postcss([autoprefixer]).process(css, {
      from: undefined /* suppress source map warning */,
  });

  return result.css;
}
```

michaeltaranto (approved):

site/docs/setup.md
```diff
@@ -93,6 +93,46 @@ require('esbuild')
 
 > Please note: There are currently no automatic readable class names during development. However, you can still manually provide a debug ID as the last argument to functions that generate scoped styles, e.g. `export const className = style({ ... }, 'className');`
 
+3/ Process CSS
```
markdalgleish, line 96: Should use dots for these headings too?

Brendan-csel:
Thanks for being so responsive folks - great teamwork! 😄

PR #192, opened 2021-06-11
Docs: Add note that `globalStyle()` only supports single class name

Brendan-csel:
Possibly suitable to address this issue https://github.com/seek-oss/vanilla-extract/issues/191


Brendan-csel:
Obsolete due to recent composeStyles change -thank you! :)

PR #198, opened 2021-06-17
Add note about extracting vendor chunk with webpack

kossnocorp:
I struggled to make vanilla-extract work with webpack because I had a configuration that extracts `vendor.js`, but I never included `vendor.css`. Also, extracting all vanilla-extract CSS into `vendor.css` doesn't make any sense,

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,29 @@ module.exports = {
   ```
 </details>
 
+<details>
+  <summary>If you extract vendor chunk using cacheGroups, make sure you exclude virtual-resource-loader and @vanilla-extract.</summary>
+
+  <br/>
+  For example:
+  
+  ```js
+  module.exports = {
+    optimization: {
+      splitChunks: {
+        cacheGroups: {
+          vendor: {
+            name: 'vendor',
+            chunks: 'initial',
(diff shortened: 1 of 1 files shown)


mattcompiles:
Hey Bud, Was having a look at this stuff the other day. I found the following worked well for splitting CSS into whatever bundle you want. 

```js
cacheGroups: {
  myCss: {
    test: (m) => {
      if (m.type === 'css/mini-extract') {
		const cssResource = m.issuer.matchResource; // -> 'MyApp/src/style.css.ts.vanilla.css'
        // Replace with whatever check you want
        return /some-test/.test(cssResource)
      }
    },
  }
}
```

Your scenario is more about excluding CSS from the cacheGroup. Not sure I'm comfortable using `virtual-resource-loader` as a way to target the CSS files as that's more of a private implementation detail. I haven't tested this but thought the better to way ignore CSS would be to just target the javascript. Something like:

```js
cacheGroups: {
  vendor: {
    name: 'vendor',
    type: 'javascript/auto',
    test: /node_modules/
  },
},
```

kossnocorp:
@mattcompiles I use webpack 4, and it doesn't work like that, but I thank you for the advice anyway.

PR #203, opened 2021-06-21
Resolve parent packageInfo when using with monorepos

samrose3:
When resolving styles from a different package in a monorepo, the `cwd` package info was being used instead of the package info relative to the styles being imported.

Below is just a simple structure example of the setup I am having issues with. When changing the packageInfo to resolve to the parent package.json relative to the styles being imported, this fixes the issue since. The `id` being passed to the `resolveId` and the `load` functions are now the same (in the Vite plugin).

```
packages
	@package/styles
		sprinkles.css.ts
site
	pages
		index.tsx
	vite.config.ts
```

### DIFF
--- a/.changeset/eight-tables-repair.md
+++ b/.changeset/eight-tables-repair.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/integration': minor
+---
+
+Improved support for resolving package info in monorepos
--- a/packages/integration/src/compile.ts
+++ b/packages/integration/src/compile.ts
@@ -9,9 +9,8 @@ import { getPackageInfo } from './packageInfo';
 export const vanillaExtractFilescopePlugin = (): Plugin => ({
   name: 'vanilla-extract-filescope',
   setup(build) {
-    const packageInfo = getPackageInfo(build.initialOptions.absWorkingDir);
-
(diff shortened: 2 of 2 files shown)


mattcompiles:
Hey 👋,

This was a conscience trade-off to prevent needing to resolve the packageInfo for every file in the project. I'm not sure it causes any issues in practice. Have you found any issues related to this?

samrose3:
I see now, yeah performing this for every file does seems very excessive (and expensive).

I am experiencing an issue trying to load `.css.ts` files across workspace packages. I have a monorepo setup with a Vite site and then a separate sibling packages directory. I want to be able to link the packages and load the files from outside website package during development. Vite seems to be handing the dep-scan step of finding the packages locally.

Currently, when the style files (`.css.ts`) are resolving, the path from the `resoveId` and `load` functions are different, causing the virtual transformed file to not return properly.

Open to alternative ideas or configurations if you suggestions on a better approach 😅

mattcompiles:
@samrose3 Interesting. Happy to take a look into this but I'll really need a reproduction of the issue. Would you be able to open an issue with a reproduction and we can work out next steps there?

samrose3:
Certainly, I'll setup a test repo to reproduce and demo 👍 Thanks Matt!

PR #213, opened 2021-06-30
Add TypeScript transformer

jimkyndemeyer:
Adds a TypeScript transformer for use with projects that use TypeScript instead of Babel for transpilation.

The implementation and tests are based on the babel-plugin with the only exception that require statements are currently not supported.

As for the docs at https://vanilla-extract.style/documentation/setup/ would be happy to contribute to that as well, but wanted to get input from the project maintainers beforehand. One option is to split the current "Webpack" section into separate "Webpack with Babel plugin" and "Webpack with TypeScript transformation" sections.

### DIFF
--- a/.changeset/serious-masks-kiss.md
+++ b/.changeset/serious-masks-kiss.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/typescript-transformer': major
+---
+
+Add TypeScript transformer for use with projects that use TypeScript instead of Babel for transpilation
--- a/packages/typescript-transformer/CHANGELOG.md
+++ b/packages/typescript-transformer/CHANGELOG.md
@@ -1,0 +1,1 @@
+# @vanilla-extract/typescript-transformer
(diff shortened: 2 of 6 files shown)


mattcompiles:
Hi 👋, Thanks for the interest in vanilla-extract, it looks like you've done a great job. However I'm concerned about supporting TypeScript transformers first-class due to lack of overall usage, seeing as TypeScript don't really support community plugins. Going forward this is a lot of complexity for us to support when it isn't a big priority for us and I feel we won't be able to do it justice. 

If this is something you would like to look after going forward we could consider adding a community managed integrations section to the docs and referencing your work there? Let me know if that's something you'd like to do.

jimkyndemeyer:
Hi Matt,

Makes sense. My initial thinking was that the API for debug ids and file scopes would be stable, but I get your point that since the transformer relies on the TypeScript AST and compiler API it would expand the maintenance scope of this project.

I'll publish the package from a separate repo, likely naming it something like `vanilla-extract-typescript-transformer`. People should be able to Google their way to it, so I don't think linking to it is needed.

Thanks again for this project, and all the other good work being shared by the Seek team.

PR #265, opened 2021-07-30
Fix fileScope error if `composeStyles` is called at runtime

mattcompiles:


tests/sprinkles/sprinkles.test.ts
```diff
@@ -2,7 +2,7 @@ import {
   createMapValueFn,
   createNormalizeValueFn,
 } from '@vanilla-extract/sprinkles';
-import { createAtomsFn } from '@vanilla-extract/sprinkles/createRuntimeAtomsFn';
+import { createAtomsFn } from '@vanilla-extract/sprinkles';
```
mattcompiles, line 5: This shows the fix working.

markdalgleish: This should be merged into the import statement above it.

markdalgleish (approved):
One minor comment, but not enough to hold back an approval.

Thanks for jumping on this 🙏

PR #270, opened 2021-08-03
vite-plugin: fix addWatchFiles call

yyx990803:
Should be iterating over the array values instead of indices.

Found out about this when testing the fix for https://github.com/vitejs/vite/issues/3216


mattcompiles (approved):
Thanks @yyx990803, I noticed this too when working on a Vite fix as well. Beat me to the punch.

PR #276, opened 2021-08-06
Replace Dynamic API with `assignInlineVars` and `setElementVars`

### LINKED ISSUE #245: Declarative way of setting inline variable
We have `setElementVar` to set a var on a DOM element. But how about a more declarative way to assign the var using inline styles? I tried to do something like

```
const currentValue = createVar();
const style = {[currentValue]: `${100 * percentage}%`};
return <div style={style} />
```

which obviously doesn't work because `currentValue` is `"var(--currentValue-xyz)"`. Is there another way to do this?

markdalgleish:
See changeset for details.

Closes #245.


packages/dynamic/src/assignInlineVars.ts
```diff
@@ -0,0 +1,45 @@
...
+
+type Styles = { [cssVarName: string]: string };
+
+export function assignInlineVars(vars: Record<string, string>): Styles;
+export function assignInlineVars<ThemeContract extends Contract>(
+  contract: ThemeContract,
+  tokens: MapLeafNodes<ThemeContract, string>,
+): Styles;
+export function assignInlineVars(varsOrContract: any, tokens?: any) {
+  const styles: { [cssVarName: string]: string } = {};
```
graup, line 17: You could remove the repeated type.
```suggestion
  const styles: Styles = {};
```

packages/dynamic/src/assignInlineVars.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (typeof tokens === 'object') {
+    const contract = varsOrContract;
+
+    walkObject(tokens, (value, path) => {
+      const varName = get(contract, path);
+
+      styles[varName.substring(4, varName.length - 1)] = String(value);
```
graup, line 25: Is the `substr` logic the same as what `getVarName` does? If so, maybe replace it with that.

graup (commented):
API looks good to me. Just left some tiny code comments.

markdalgleish:
@graup Thanks for the feedback 👍

packages/dynamic/src/assignInlineVars.ts
markdalgleish: The original idea here was that it's faster since we already know it's a vanilla-extract contract, but in hindsight this feels like an over-optimisation at the expense of consistent variable handling. If we update the `getVarName` logic in the future (e.g. adding support for vars with fallbacks as keys), this would need to keep up.

mattcompiles (approved):

PR #290, opened 2021-08-14
fix: normalize path(vitejs/vite#4556)

ygj6:
Fixes vitejs/vite#4556

See  [issue comment]( https://github.com/vitejs/vite/issues/4556#issuecomment-898862114)  for details

### DIFF
--- a/packages/vite-plugin/package.json
+++ b/packages/vite-plugin/package.json
@@ -19,5 +19,8 @@
   },
   "devDependencies": {
     "vite": "^2.2.3"
+  },
+  "peerDependencies": {
+    "vite": "^2.2.3"
   }
 }
--- a/.changeset/short-mayflies-reflect.md
+++ b/.changeset/short-mayflies-reflect.md
@@ -1,0 +1,5 @@
+---
+"@vanilla-extract/vite-plugin": patch
+---
+
+Normalize path of generated CSS files
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -1,4 +1,5 @@
 import type { Plugin, ResolvedConfig } from 'vite';
+import { normalizePath } from 'vite';
 import {
   cssFileFilter,
   virtualCssFileFilter,
@@ -24,7 +25,9 @@ export function vanillaExtractPlugin(): Plugin {
 
         // resolveId shouldn't really cause a side-effect however custom module meta isn't currently working
         // This is a hack work around until https://github.com/vitejs/vite/issues/3240 is resolved
-        const shortHashFileName = `${fileName}?hash=${hash(source)}`;
+        const shortHashFileName = normalizePath(
+          `${fileName}?hash=${hash(source)}`,
+        );
         cssMap.set(shortHashFileName, source);
 
         return shortHashFileName;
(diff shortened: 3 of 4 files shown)


packages/vite-plugin/package.json
```diff
@@ -15,9 +15,7 @@
   "author": "SEEK",
   "license": "MIT",
   "dependencies": {
-    "@vanilla-extract/integration": "^1.1.0"
-  },
-  "devDependencies": {
+    "@vanilla-extract/integration": "^1.1.0",
```
mattcompiles, line 18: Not sure why this was changed? Can you please revert this?

mattcompiles (commented):
Thanks for the PR. Just need you to remove your package.json change.

packages/vite-plugin/package.json
ygj6: https://github.com/seek-oss/vanilla-extract/blob/7c396ea58f5af60e65176875ae9c93764fe77fd2/packages/vite-plugin/src/index.ts#L2

Since we need to import the `normalizePath` method and use it, we need to adjust the `vite` dependencies to `dependencies` or `peerDependencies`, otherwise an exception will occur at build.
image

Maybe I should have adjusted it to `peerDependencies` more appropriately.

mattcompiles: Right. Sorry I misread this change. I think `peerDependencies` is the right call here.

ygj6: Done, thank you for the review.

mattcompiles: Sorry but you'll need to leave the dev dependency there as well, otherwise it may not be accessible locally.

ygj6: Sorry, I don't quite understand what you mean, because it is a plugin for `vite`, there is no way to run it alone, it can only work with `vite`, so `vite`must exist.

ygj6:
I noticed that the check failed and it says I can't modify the dependency, what should I do?
image

packages/vite-plugin/package.json
```diff
@@ -17,7 +17,7 @@
   "dependencies": {
     "@vanilla-extract/integration": "^1.1.0"
   },
-  "devDependencies": {
+  "peerDependencies": {
     "vite": "^2.2.3"
   }
```
mattcompiles, line 22: To fix CI commit this suggestion, then run a `yarn` locally to update the lockfile. 

```suggestion
  "devDependencies": {
    "vite": "^2.2.3"
  },
  "peerDependencies": {
    "vite": "^2.2.3"
  }
```

mattcompiles:
@ygj6 Hey Bud. This looks good but it's failing the format check. I tried to push the change to your branch but you haven't given me permission. Running `yarn format` and pushing the result should be enough to get CI green and then we can merge 👍

ygj6:
> @ygj6 Hey Bud. This looks good but it's failing the format check. I tried to push the change to your branch but you haven't given me permission. Running `yarn format` and pushing the result should be enough to get CI green and then we can merge 👍

An invitation has been extended and you need to agree.

ygj6:
Done, thanks for your patience in reviewing, I've done `yarn formatting` of the code.
Also, there were too many `commits` records and I did a `rebase`.

mattcompiles (approved):
Thanks for the fix mate 👍

PR #303, opened 2021-08-19
Sprinkles: Support passing style objects as property values

rtkaaho:
As part of some experimentation I have come across a use case that is desired in the sprinkles API. Currently when creating atomic styles, the values that can be passed must be either a `string | number`. This PR, provides users the ability to pass a custom `StyleRule` as a value as well. Please see the following example use case below:

Suppose I want to generate atom styles for a subset of colors. But, instead of being HEX values, they are rgb with an optional alpha override. This use case is very similar to the way tailwind has implemented the `bg-opacity` api. Where they define the color as `rgba(255, 0, 0, var(--tw-opacity))`. Currently, to my knowledge, sprinkles doesn't suppose this in an elegant way.

With the help of this PR, one could accomplish this behaviour by doing the following:

```ts
// atomic styles

const colorStyles = createAtomicStyles({
  properties: {
    bgOpacity: {
      none: {
        vars: {
          '--alpha': 0
        }
      },
      '10': {
        vars: {
          '--alpha': 0.1
        }
      }
    },
    background: {
      red: `rgb(255, 0, 0, var(--alpha, 1))`
    }
  },
});

// consuming in the atoms

const className = atoms({
  background: 'red',
  bgOpacity: '10'
});
```

Please let me know if there is anything I am missing in this PR. I am more so opening it as a way to prompt some conversation about the implementation. I am open to modifying it, or closing it if we think its a dumb idea :)


Bubz43:
I think you would need to reset the css var every time or it would cascade.

`<div class={}>
</div>`

Bubz43:
I think you would need to reset the css var inside the same color class or it will be affected by the cascade from parent components.

```jsx
<div class={atoms({ background: 'red', bgOpacity: 0.1 })}>
    <div class={atoms({ background: 'blue' })} /> // This would also have a background opacity of 0.1
</div>
```
This is what tailwind generates in this situation:
```css
.bg-red {
  --tw-bg-opacity: 1;
  background-color: rgba(255, 0, 0, var(--tw-bg-opacity));
}
```

rtkaaho:
> I think you would need to reset the css var inside the color class or it will be affected by the cascade from parent components.
> 
> ```js
> <div class={atoms({ background: 'red', bgOpacity: 0.1 })}>
>     <div class={atoms({ background: 'red' })} /> // This would also have a background opacity of 0.1
> </div>
> ```
> 
> This is what tailwind generates in this situation:
> 
> ```css
> .bg-red {
>   --tw-bg-opacity: 1;
>   background-color: rgba(255, 0, 0, var(--tw-bg-opacity));
> }
> ```

Ahh, very good point. I will need to think about this a little more to see what can be done to mitigate that.

rtkaaho:
So after playing around locally on my branch I realized that with this PR you can just update the example above to this:

```ts
const colorStyles = createAtomicStyles({
  properties: {
    bg: {
      red: {
        vars: {
          "--alpha": 1
        },
        background: `rgb(255, 0, 0, var(--alpha, 1))`
      },
    },
    bgOpacity: {
      none: {
        vars: {
          '--alpha': 0
        }
      },
      '10': {
        vars: {
          '--alpha': 0.1
        }
      }
    }
  },
});
```

Properties are evaluated in the order that they appear, so for specificity you need to put `bg` before `bgOpacity`.

fixtures/sprinkles/src/styles.css.ts
```diff
@@ -29,6 +30,26 @@ const responsiveStyles = createAtomicStyles({
...
+        vars: {
+          '--alpha': '1',
+        },
+      },
+      '10': {
+        vars: {
+          '--alpha': '0.1',
+        },
+      },
+    },
```
rtkaaho, line 52: This is an example of how I would like to be able to use this. The output here is:

```css
.background_red {
   --alpha: 1;
  background: rgba(255,0,0, var(--alpha, 1));
}

.bgOpacity_none {
  --alpha: 0;
}

.bgOpacity_10 {
   --alpha: 0.1;
}
```

This to me feels really powerful because the atoms no longer become a 1:1 relationship between the property and value. But instead become a combination of styles. This is just one immediate use case I have for this. But I can see more coming in the future.

markdalgleish:
Thanks so much for this PR! 🙏

I'm super excited about this. We've actually already spoken about adding this functionality to Sprinkles, and your example shows why this is such a powerful addition to the library. Really unlocks so much potential.

I've pushed some updates to your branch so I'll get another maintainer to have a look over it too, but this is looking pretty good to me.

markdalgleish (approved):

markdalgleish:
Out now in Sprinkles v1.1.0. Thank you!! 🙏

adamaho:
Thanks for your support! Im looking forward to taking advantage of this now :)

PR #307, opened 2021-08-20
Sprinkles: Fix style object bug when no conditions are present

rtkaaho:
I realized I missed the `else` block where we handle the case where no `conditions` are provided in the atomic styles. This caused the css to compile incorrectly. This PR addresses that bug.

Use case:

```ts
const styleWithNoConditions = createAtomicStyles({
  properties: {
    color: {
      red: {
        vars: {
          [alpha]: '1'
        },
        color: `rgba(255, 0, 0, ${alpha})`
      }
    }
  }
});
```

### DIFF
--- a/tests/E2E/__snapshots__/stylesheets.test.ts.snap
+++ b/tests/E2E/__snapshots__/stylesheets.test.ts.snap
@@ -1,11 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Stylesheet sprinkles - esbuild should create valid stylesheet 1`] = `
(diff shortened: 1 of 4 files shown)


rtkaaho:
Is there something special besides `yarn build` that i need to do in order for the tests to run? I keep running into these errors when I try to update the snapshots:

```
    /Users/<user>/<path>/vanilla-extract/babel-jest.config.js: Error while loading config - Unexpected token '?'

      at loadCjsDefault (node_modules/@babel/core/lib/config/files/module-types.js:77:18)
      at loadCjsOrMjsDefault (node_modules/@babel/core/lib/config/files/module-types.js:49:16)
          at loadCjsOrMjsDefault.next (<anonymous>)
```

markdalgleish:
Sounds like you might be on an old version of Node that doesn't support the `??` operator?

tests/E2E/__snapshots__/stylesheets.test.ts.snap
```diff
@@ -1,143 +1,94 @@
...
+  .styles_backgroundOpacity_0\\\\.2_tablet__3nw5tzz {
     --alpha__3nw5tz0: 0.2;
   }
-  .styles_backgroundOpacity_0\\\\.3_tablet__3nw5tz12 {
+  .styles_backgroundOpacity_0\\\\.3_tablet__3nw5tz13 {
     --alpha__3nw5tz0: 0.3;
   }
 }
 @media screen and (min-width: 1024px) {
-  .styles_display_flex_desktop__3nw5tz3 {
```
markdalgleish, line 79: I pushed an update to the snapshot, but it looks like it's removed some conditional styles?

mattcompiles:
Fixed the issue. Hoisting the `styleValue` caused the conditions to become accumulative and create invalid styles with media queries within media queries. Each condition scope has it's own `styleValue` now so should be good 👍

markdalgleish (approved):

markdalgleish:
Out now in Sprinkles v1.1.1 🎉

PR #310, opened 2021-08-22
Fix the Snowpack plugin to make it work with Astro + Astro example

kossnocorp:
The PR makes vanilla-extract work with Astro. I contacted the plugin author @mxmul (thanks 👋) and turned out that he already made the fix but couldn't find the time to open the PR.

Here's the video demonstrating the problem and the fix:

[](https://www.loom.com/share/3321d7bd0ec74c54bcad545119c8a793)

---

[The PR fixes](https://github.com/seek-oss/vanilla-extract/commit/395fe5442e967845f92c5efc342cab1da0fdca9c) the incorrectly used `isSSR` flag and makes the plugin produce CSS both in development and build environments.

The Snowpack documentation states:

> Some frameworks/languages (like React) run the same code on both the browser and the server. Others (like Svelte) will create different output for the server than the browser. In the example below, we use the isSSR flag to tell the Svelte compiler to generate server-optimized code when requested by Snowpack.

It doesn't applicable to vanilla-extract and hence `isSSR` should be used (cc @fredkschott, am I right?).

---

It also adds an Astro example demonstrating how to set up vanilla-extract with the site generator. For that, I ported the webpack/React example which also uses Sprinkles and Tailwind CSS' colors.

Currently, the example assumes that the PR is released as `1.1.1` (the latest version is `1.1.0`, so to make it work - [use my fork `@kossnocorp/vanilla-extract-snowpack-plugin` instead](https://github.com/kossnocorp/vanilla-extract/commit/e2a1e44c6178c31d9e33ba5b5e600306ce6dfc72).


examples/astro/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "name": "vanilla-extract-astro",
+  "description": "Example Astro project using vanilla-extract",
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "start": "astro dev",
+    "build": "astro build"
+  },
+  "devDependencies": {
+    "@vanilla-extract/snowpack-plugin": "^1.1.1",
```
mattcompiles, line 11: You'll need to reference `1.1.0` as `1.1.1` doesn't exist yet. Don't worry though as it will be automatically updated by changesets when it releases.

kossnocorp: Done!

markdalgleish:
I've been trying to get the example running locally on my machine. Fixed a couple of config issues, but now I'm getting the following error: `ReferenceError: module is not defined`. Not sure what's wrong.

Here are the steps I'm expecting to work on a fresh copy of your branch:

```
yarn
yarn build
cd examples/astro
yarn start
```

kossnocorp:
@markdalgleish sorry I never tried `yarn`, used only `npm`. Did https://github.com/seek-oss/vanilla-extract/pull/310/commits/1759a76119a872678fd9768984997dae85ab1fcc fix the issue?

markdalgleish:
Nope, that was just fixing a lint error from [manypkg](https://github.com/Thinkmill/manypkg) due to mismatched package versions.

kossnocorp:
@markdalgleish I haven't been able to make Snowpack work with yarn workspaces. I don't use Yarn workspaces and have no idea how it works, so I don't even know what could go wrong. If you try to run `npm i` and follow my instructions, you'll see that the fix works.

I'm going to extract the example in a separate PR to unblock the merge of the PR and try more or ask someone knowledgeable in Yarn or Snowpack for help.

markdalgleish:
If you update this PR to be just the Snowpack plugin fix, we'll be happy to merge it 🙂

mattcompiles:
Closing in favour of #335

Issue #309, opened 2021-08-21
Feature Request: Customize (or remove entirely) hash on css variables. 


nicksrandall:
Hello!

I thought I would document my use-case and see if there is any interest in supporting it within Vanilla Extract. If so, I'm happy to send in some PRs. 

I'm building a set a reusable components that will be consumed by several different applications. All of these components are using `vanilla-extract` for css and they pull variables in from a "global theme". Because all these components are using `vanilla-extact`, getting access to the css variable names is easy. 

However, the "host" applications that will be consuming these components have not migrated to using vanilla extract -- but they would still like to use some of the variables from the global theme in their hand-written, vanilla CSS. Because of the randomized hash that gets added to the CSS vars, the task of looking up these variable names is cumbersome. I wondering if it would be possible to create an API where the hash could be over-written or removed entirely.

What I'm thinking is that I would like to pass in an optional second param to `createThemeContract` that would define the hash that is to be used by all the variable names created by that contract. I could potentially pass `""` to denote that I would not like a hash at all.

Are the maintainers of this library open to this change? If so, any guidance on how the API should look?

P.S. I know about the DEBUG ids -- I think what I asking for is different but could potentially conflict with that feature.

PR #315, opened 2021-08-23
Expose `NullableTokens` and `ThemeVars` types

nicksrandall:
This facilitates a use case described in #309 where user wants to create their own version of `createThemeContract` to make var names predictable.


nicksrandall:
Closing in favor of #319

PR #335, opened 2021-09-03
Make Snowpack plugin output CSS

jonathantneal:
The commit makes vanilla-extract to work with Astro, preserving the great work by @kossnocorp in https://github.com/seek-oss/vanilla-extract/pull/310, as requested by [@markdalgleish](https://github.com/seek-oss/vanilla-extract/pull/310#issuecomment-912562258).

It fixes the incorrectly used `isSSR` flag and makes the plugin produce CSS both in development and build environments. The Snowpack documentation states:

> Some frameworks/languages (like React) run the same code on both the browser and the server. Others (like Svelte) will create different output for the server than the browser. In the example below, we use the isSSR flag to tell the Svelte compiler to generate server-optimized code when requested by Snowpack.

This is not applicable to `vanilla-extract`, hence `isSSR` should be used.


packages/snowpack-plugin/src/index.ts
```diff
@@ -57,10 +57,7 @@ export default function vanillaExtractPlugin(
       const js = processVanillaFile({
         source,
         filePath,
-        outputCss: !isSSR,
-        identOption:
```
mattcompiles, line 61: I think we should retain the identOption parameter as it's still valid.

packages/snowpack-plugin/src/index.ts
```diff
@@ -57,10 +57,7 @@ export default function vanillaExtractPlugin(
...
-          identifiers ??
-          (snowpackConfig.mode === 'production' ? 'short' : 'debug'),
+        outputCss: true,
```
mxdvl, line 60: It would be great to get this merged!

```suggestion
        outputCss: true,
        identOption:
          identifiers ??
          (snowpackConfig.mode === 'production' ? 'short' : 'debug'),
```

mattcompiles (approved):

kossnocorp:
@jonathantneal thank you for that!

PR #336, opened 2021-09-04
Add Next.js plugin

markdalgleish:


packages/next-plugin/src/index.js
```diff
@@ -0,0 +1,34 @@
...
+import { getGlobalCssLoader } from 'next/dist/build/webpack/config/blocks/css/loaders';
+
+export const createVanillaExtractPlugin =
+  (pluginOptions = {}) =>
+  (nextConfig = {}) =>
+    Object.assign({}, nextConfig, {
+      webpack(config, options) {
+        const { dev, isServer } = options;
+
+        config.module.rules[config.module.rules.length - 1].oneOf.unshift({
```
stefanprobst, line 11: i think  there's no guarantee the css related rules are last because other plugins might have pushed to `config.module.rules`

markdalgleish: Yep, that's a good point. I'll see it I can figure out a good way to detect which item in the array should be modified.

stefanprobst: maybe something like `config.module.rules.find(rule => Array.isArray(rule.oneOf) && rule.oneOf.some(r => r.test.test('test.css')))`

markdalgleish: Yep, I was thinking something like this.

markdalgleish: Pushed an update to make this work similar to how you suggested. Thanks for the feedback 👍

mattcompiles (approved):

PR #337, opened 2021-09-05
Move E2E tests to screenshots based tests

mattcompiles:
Primary changes:
- Migrate the "styles" tests which currently snapshot normalised CSS rules in jest-puppeteer to screenshots tests run by [playwright](https://playwright.dev/)
  - Note that the screenshots are saved per fixture. We assert that each integration should output an identical screenshot. 
- Setup new CI workflow to run screenshot diffs. It uses `mac-latest` to match with mac dev machines. PRs from non-mac environments will probably need to get maintainers to update the snapshots if required.
- Refactor "stylesheet" tests (which snapshot the output CSS file) into multiple files to allow for running in parallel. These are still run in jest.


benjervis (approved):
📸

PR #339, opened 2021-09-06
Cleanup adapter after processing CSS files

mattcompiles:
Also move stylesheet tests to playwright.


mattcompiles:
Closing to revisit test strategy

PR #343, opened 2021-09-06
Cleanup adapter after processing styles

mattcompiles:


.github/workflows/validate.yml
```diff
@@ -28,3 +28,34 @@ jobs:
 
       - name: Test
         run: yarn test:jest
+  playwright:
```
mattcompiles, line 31: Unrelated cleanup

packages/integration/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "1.2.0",
   "description": "Zero-runtime Stylesheets-in-TypeScript",
   "main": "dist/vanilla-extract-integration.cjs.js",
-  "module": "dist/vanilla-extract-integration.esm.js",
```
mattcompiles, line 6: Decided to only export a CJS version of the integration as there was no reason to have ESM

markdalgleish (approved):

PR #352, opened 2021-09-11
Use `transform` property in docs instead of `rotate`

domosedov:
at the moment the "rotate" css properties have little browser support.
https://caniuse.com/?search=rotate


markdalgleish (approved):

markdalgleish:
Thank you! 🙏

PR #376, opened 2021-09-21
Add RecipeVariants type to get variants from recipes

TheMightyPenguin:
# Description
Add `RecipeVariants` type to get variants from a recipe.

This type would help consumers easily get the variants type. Useful for typing their components:

```tsx
const textRecipes = recipe({
  variants: {
    size: {
      small: { fontSize: '12px' },
      medium: { fontSize: '16px' },
      large: { fontSize: '24px' },
    },
  },
});

type TextVariants = RecipeVariants<typeof textRecipes>;

type Props = {
  // ... other props
} & TextVariants;

function Text({ size, ...props }: Props) {
  return <div {...props} className={textRecipes({ size })} />;
}

<Text size="small">Awesome text</Text>;
```

For more context, these types are easy to get from outside the lib using `Parameters`, but having these utilities would make it easy and provide an official way to get these, plus if the way to get the type changes, consumers won't have to worry about it as the type abstracts this.

## Other Options

* Maybe exposing `recipe.variants` (similar to the `properties` property on the sprinkles returned function), then doing:
```ts
type Variants = typeof recipe.variants;
```

* Even if the type is something that ends up not being added, I think at least having some documentation into how to get these types is useful for library consumers.

If this ends up being useful, I can also add a similar type for sprinkles.

### DIFF
--- a/.changeset/rare-lemons-study.md
+++ b/.changeset/rare-lemons-study.md
@@ -11,7 +11,7 @@ This type would help consumers easily get the variants type. Useful for typing t
 import { recipe } from '@vanilla-extract/recipes';
 import { reset } from './reset.css.ts';
 import { sprinkles } from './sprinkles.css.ts';
-// 
+
 export const button = recipe({
   variants: {
     color: {
--- a/.changeset/rare-lemons-study.md
+++ b/.changeset/rare-lemons-study.md
@@ -4,13 +4,11 @@
 
 Add `RecipeVariants` type to get variants from a recipe.
 
-This type would help consumers easily get the variants type. Useful for consuming variants from a Component, functions, or other parts of the code:
+This type allows you to extract the variants from the recipe function. Useful for consuming variants from a componen.
 
 ```ts
 // button.css.ts
 import { recipe } from '@vanilla-extract/recipes';
-import { reset } from './reset.css.ts';
-import { sprinkles } from './sprinkles.css.ts';
 
 export const button = recipe({
   variants: {
--- a/.changeset/rare-lemons-study.md
+++ b/.changeset/rare-lemons-study.md
@@ -11,31 +11,19 @@ This type would help consumers easily get the variants type. Useful for typing t
 import { recipe } from '@vanilla-extract/recipes';
 import { reset } from './reset.css.ts';
 import { sprinkles } from './sprinkles.css.ts';
-
+// 
 export const button = recipe({
   variants: {
     color: {
-      neutral: {
-        background: 'neutral'
-      },
-      brand: {
-        background: 'brand'
-      },
-      accend: {
-        background: 'accent'
-      }
+      neutral: { background: 'whitesmoke' },
+      brand: { background: 'blueviolet' },
+      accent: { background: 'slateblue' }
     },
     size: {
-      small: {
-        padding: 'small'
-      },
-      medium: {
-        padding: 'medium'
-      },
-      large: {
-        padding: 'large'
-      }
-    }
+      small: { padding: 12 },
+      medium: { padding: 16 },
+      large: { padding: 24 }
+    },
   }
 });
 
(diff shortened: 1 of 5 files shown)


TheMightyPenguin:
@mattcompiles @michaeltaranto @markdalgleish  tagging for review 👀

site/docs/recipes-api.md
```diff
@@ -112,3 +112,40 @@ export const button = recipe({
...
+      medium: sprinkles({ padding: 'medium' }),
+      large: sprinkles({ padding: 'large' })
+    }
+  }
+});
+
+// Get the type 
+type ButtonVariants = RecipeVariants<typeof button>;
+
+// the above will result in a similar type to:
```
mattcompiles, line 146:
```suggestion
// the above will result in a type equivalent to:
```

site/docs/recipes-api.md
```diff
@@ -112,3 +112,40 @@ export const button = recipe({
...
+type ButtonVariants = {
+  color: 'neutral' | 'brand' | 'size';
+  size: 'small' | 'medium' | 'large';
```
mattcompiles, line 149: I believe `color` and `size` should be option here

site/docs/recipes-api.md
```diff
@@ -112,3 +112,40 @@ export const button = recipe({
...
+```ts
+// button.css.ts
+import { recipe } from '@vanilla-extract/recipes';
+import { reset } from './reset.css.ts';
+import { sprinkles } from './sprinkles.css.ts';
+
+export const button = recipe({
+  variants: {
+    color: {
+      neutral: sprinkles({ background: 'neutral' }),
```
mattcompiles, line 131: Think it's better to keep this example focussed on just the recipes API. Could we just use regular style blocks rather than sprinkles calls?

.changeset/rare-lemons-study.md
```diff
@@ -0,0 +1,31 @@
...
+
+type Props = {
+  // ... other props
+} & TextVariants;
+
+function Text({ size, ...props }: Props) {
+  return <div {...props} className={textRecipes({ size })} />;
+}
+
+<Text size="small">Awesome text</Text>;
```
mattcompiles, line 30: We try to avoid framework specific examples in docs and changesets. Would be great to just use the example that you wrote in the docs here.

.changeset/rare-lemons-study.md
```diff
@@ -0,0 +1,31 @@
+---
+'@vanilla-extract/recipes': patch
```
mattcompiles, line 2: I know it's small change but technically a minor as we're adding API. 
```suggestion
'@vanilla-extract/recipes': minor
```

tests/recipes/recipes-type-tests.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const invalidVariantValue: TextVariants = {
+    // @ts-expect-error Type '"extraLarge"' is not assignable to type '"small" | "large" | "medium" | undefined'.
+    size: 'extraLarge',
+  };
+
+  const invalidVariantName: TextVariants = {
+    // @ts-expect-error Property 'color' does not exist in type
+    color: 'brand',
+  };
```
mattcompiles, line 30: Could we add working example as well just to prove it works.

mattcompiles (commented):
Thanks for the PR. Was definitely keen to add this. Just a few comments around docs mostly then good to go 👍 🍻

site/docs/recipes-api.md
TheMightyPenguin: Yes good catch :raised_hands:

site/docs/recipes-api.md
TheMightyPenguin: Yep let me change this, makes sense!

tests/recipes/recipes-type-tests.ts
TheMightyPenguin: Yup 👍 added, thanks! I also added an extra type-level test to make sure the recipes API does return a string (let me know if this one was too much!)

TheMightyPenguin:
@mattcompiles attended the comments, thank you for the review :raised_hands:! let me know if more changes are needed 😄

mattcompiles (approved):

mattcompiles:
Thanks for the PR mate 👏 🍻

PR #381, opened 2021-09-24
Fix identifiers not respecting `NODE_ENV` for SSR builds

### LINKED ISSUE #372: Using short/debug identifiers outside of webpack
### Describe the bug
When using the webpack and babel plugin, I can configure `identifiers: isProduction ? 'short' : 'debug'` to control how the class names are generated inside webpack (or "client-side").

However, when doing SSR (code goes only through babel, not webpack, so "server-side"), the same cannot be configured, and it seems the generated class names are always of the type 'debug'.

It would be convenient if the babel plugin also contained the `identifiers` option, just like the webpack plugin:

```json
{
  "plugins": [
    ["@vanilla-extract/babel-plugin", { "identifiers": "short" }]
  ] 
}
```

### Link to reproduction
1. git clone https://github.com/iiroj/iiro.fi
1. run `npm ci`
1. run `npm run --workspace=site build`
1. notice how `site/dist/index.html` contains debug class names, as well as all CSS files
1. change the webpack plugin config at `site/webpack.config.babel.ts` to generate `short` identifiers and re-build
1. notice how the index.html file still contains debug class names

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

```node
  System:
    OS: macOS 11.6
    CPU: (8) arm64 Apple M1
    Memory: 260.52 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.9.1 - /opt/homebrew/bin/node
    Yarn: 1.22.11 - /opt/homebrew/bin/yarn
    npm: 7.24.0 - /opt/homebrew/bin/npm
  Browsers:
    Firefox: 92.0
    Safari: 14.1.2

```

mattcompiles:
Fixes #372

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -5,7 +5,7 @@ const config: PlaywrightTestConfig = {
   updateSnapshots: 'none',
   expect: {
     toMatchSnapshot: {
-      threshold: 0.1,
+      threshold: 0.2,
     },
   },
 
--- a/.changeset/rude-fireants-help.md
+++ b/.changeset/rude-fireants-help.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Fix identfiers not respecting `NODE_ENV` for SSR builds
--- a/fixtures/recipes/package.json
+++ b/fixtures/recipes/package.json
@@ -7,6 +7,6 @@
   "private": true,
   "dependencies": {
     "@vanilla-extract/css": "1.6.0",
-    "@vanilla-extract/recipes": "0.1.0"
+    "@vanilla-extract/recipes": "0.2.0"
   }
 }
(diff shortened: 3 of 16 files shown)


playwright.config.ts
```diff
@@ -5,7 +5,7 @@ const config: PlaywrightTestConfig = {
   updateSnapshots: 'none',
   expect: {
     toMatchSnapshot: {
-      threshold: 0.1,
+      threshold: 0.2,
```
mattcompiles, line 8: Making this more lenient to handle difference MacOS versions.

markdalgleish (approved):

iiroj:
Heh, I got totally bitten by this myself because I forgot about #372 and updated my patch-level dependencies. Since I had already configured the Webpack plugin to always use `debug` identifiers, this lead to a broken production built; the fix was to return to the default behavior (removing the override setting from the webpack config).

PR #400, opened 2021-10-07
Use a virtual prefix for Vite plugin

### LINKED ISSUE #397: Vite Plugin fails to rebuild after file changes in watch mode
### Describe the bug
When using the vite plugin along with vites library build mode and `--watch` enabled, there seems to be a caching issue that results in the following error:

```
[vite:load-fallback] Could not load vanilla-extract-rebuild-error/src/sprinkles.css.ts.vanilla.css?hash=d39ecf9e2120879a6ab8b71bd8033bc8 (imported by src/sprinkles.css.ts?used): ENOENT: no such file or directory, open 'vanilla-extract-rebuild-error/src/sprinkles.css.ts.vanilla.css'
```

### Link to reproduction
https://github.com/rtkaaho/vanilla-extract-rebuild-error

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

```node
  System:
    OS: macOS 11.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 615.15 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 12.19.0 - ~/.nvm/versions/node/v12.19.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v12.19.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.19.0/bin/npm
    Watchman: 4.9.0 - /opt/local/bin/watchman
  Browsers:
    Chrome: 93.0.4577.63
    Firefox: 92.0
    Safari: 15.0
```

fwouts:
This adds an explicit prefix to virtual modules generated by the Vite plugin.

It helps tell Vite that these files don't exist so it doesn't try to optimise them (it seems to fix #397) and fixes https://github.com/zenclabs/reactpreview/issues/52.

See https://vitejs.dev/guide/api-plugin.html#importing-a-virtual-file for more info. Note the `/` prefix is required or Vite automatically adds an additional `/@id/` to the module ID before `load()` gets invoked (I'm not sure why).

I've tested this change manually but I'm not 100% confident that I haven't broken anything. I was wondering if setting up some tests in the form of a simple app with associated Puppeteer tests would be beneficial too?

### DIFF
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -22,6 +22,9 @@ interface Options {
    */
   devStyleRuntime?: 'vite' | 'vanilla-extract';
 }
+
+const VIRTUAL_PREFIX = `/@vanilla-extract:`;
+
 export function vanillaExtractPlugin({
   identifiers,
   devStyleRuntime = 'vite',
@@ -52,14 +55,19 @@ export function vanillaExtractPlugin({
         );
         cssMap.set(shortHashFileName, source);
 
-        return shortHashFileName;
+        return VIRTUAL_PREFIX + shortHashFileName;
       }
     },
     load(id) {
-      if (cssMap.has(id)) {
-        const css = cssMap.get(id);
+      if (!id.startsWith(VIRTUAL_PREFIX)) {
+        return null;
+      }
 
-        cssMap.delete(id);
+      const shortHashFileName = id.substr(VIRTUAL_PREFIX.length);
+      if (cssMap.has(shortHashFileName)) {
+        const css = cssMap.get(shortHashFileName);
+
+        cssMap.delete(shortHashFileName);
 
         return css;
       }
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -75,12 +75,16 @@ export function vanillaExtractPlugin({
       return null;
     },
     async transform(code, id, ssr) {
-      if (!cssFileFilter.test(id)) {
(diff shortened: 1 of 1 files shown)


mattcompiles:
Hey Bud 👋

I'm not really sure what is being fixed here? The docs you pointed to are just an example, I don't see anywhere where it says all virtual modules must start with a special character. Can  we get a bit more context on what the actual problem is?

Also there is a playwright test suite already.

fwouts:
> Hey Bud 👋

Hi from Sydney!

> I'm not really sure what is being fixed here? The docs you pointed to are just an example, I don't see anywhere where it says all virtual modules must start with a special character. Can we get a bit more context on what the actual problem is?

I'll admit I don't understand all the details of how Vite works, but some parts of Vite will crash (as shown in #397) because they see a module ID that seems to correspond to a real file, and fail to access it from the file system. This prefix is common in Vite plugins, for example see https://github.com/vitejs/vite/issues/2380#issuecomment-796902541.

> 
> Also there is a playwright test suite already.

I meant a test specifically for the Vite plugin, I couldn't find one :)

mattcompiles:
Hey, I'm still not really sure what this PR addresses? #397 has already been resolved and the fix was unrelated. Going to close this for now but if there's something I've missed then let me know.

PR #409, opened 2021-10-11
Update browser runtime

benjervis:
Primarily moving to the new vanilla runtime.

Also includes a bunch of bugfixes especially around the vite plugin.


.changeset/polite-rings-nail.md
```diff
@@ -0,0 +1,9 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Update the vanilla runtime integration.
```
mattcompiles, line 5:
```suggestion
Update the "vanilla-extract" devStyleRuntime option
```

packages/vite-plugin/package.json
```diff
@@ -15,7 +15,8 @@
   "author": "SEEK",
   "license": "MIT",
   "dependencies": {
-    "@vanilla-extract/integration": "^1.4.2"
+    "@vanilla-extract/integration": "^1.4.2",
+    "outdent": "^0.8.0"
```
mattcompiles, line 19: Realised we're already using `dedent` in the css package so we should probably be consistent.

packages/integration/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.4.2",
   "description": "Zero-runtime Stylesheets-in-TypeScript",
   "main": "dist/vanilla-extract-integration.cjs.js",
+  "module": "dist/vanilla-extract-integration.esm.js",
```
mattcompiles, line 6: This isn't required anymore.

packages/integration/src/index.ts
```diff
@@ -1,4 +1,8 @@
-export { processVanillaFile } from './processVanillaFile';
+export {
+  processVanillaFile,
+  parseFileScope,
+  stringifyFileScope,
+} from './processVanillaFile';
```
mattcompiles, line 5: We need to add a changeset for the integration package

packages/vite-plugin/src/index.ts
```diff
@@ -31,37 +32,64 @@ export function vanillaExtractPlugin({
...
     },
     load(id) {
-      if (cssMap.has(id)) {
-        const css = cssMap.get(id);
+      const extensionIndex = id.indexOf(virtualExt);
+
+      if (extensionIndex > 0) {
+        const fileScopeId = id.substring(0, extensionIndex);
+
+        if (cssMap.has(fileScopeId)) {
```
mattcompiles, line 75: Wondering if we should throw an explicit error if this condition is false. It should never really occur yeah?

packages/vite-plugin/src/index.ts
```diff
@@ -71,32 +99,38 @@ export function vanillaExtractPlugin({
         return null;
       }
 
-      const usedIndex = id.indexOf('?used');
-      const fixedId = usedIndex > 0 ? id.substring(0, usedIndex) : id;
-
-      if (ssr || useRuntime) {
+      if (ssr) {
+        const validId = id.substring(0, id.indexOf('?'));
```
mattcompiles, line 103: This needs to be resilient to strings that don't have a query attached.

packages/vite-plugin/src/index.ts
```diff
@@ -71,32 +99,38 @@ export function vanillaExtractPlugin({
...
           source: code,
-          filePath: normalizePath(path.relative(packageInfo.dirname, fixedId)),
+          filePath: normalizePath(path.relative(packageInfo.dirname, validId)),
           packageInfo,
         }).source;
       }
 
       const { source, watchFiles } = await compile({
-        filePath: fixedId,
+        filePath: id,
```
mattcompiles, line 112: I think we should continue to strip the query if it has one, just to be sure there's no funny business.

packages/vite-plugin/src/index.ts
```diff
@@ -71,32 +99,38 @@ export function vanillaExtractPlugin({
...
         cwd: config.root,
       });
 
       for (const file of watchFiles) {
-        this.addWatchFile(file);
+        if (config.command === 'build' || file !== id) {
```
mattcompiles, line 117: Might be worth adding a comment as to why we have the `config.command === 'build'` check.

packages/vite-plugin/src/index.ts
```diff
@@ -71,32 +99,38 @@ export function vanillaExtractPlugin({
...
+          this.addWatchFile(file);
+        }
       }
 
       return processVanillaFile({
         source,
-        filePath: fixedId,
-        outputCss: !ssr,
+        filePath: id,
```
mattcompiles, line 124: Think we should continue to strip the query here as well.

packages/vite-plugin/src/index.ts
```diff
@@ -31,37 +32,66 @@ export function vanillaExtractPlugin({
...
 
-        return css;
+        const fileScope = parseFileScope(fileScopeId);
+
+        return dedent`
+          import { injectStyles } from '@vanilla-extract/css/injectStyles';
+          
+          injectStyles({
+            fileScope: ${JSON.stringify(fileScope)},
+            css: ${JSON.stringify(css)}
```
mattcompiles, line 92: This seems to be borking the screenshot tests due to the newline encoding

PR #414, opened 2021-10-14
Update `BooleanMap` type

swudged:
Updates the `BooleanMap` type to include both `true` and `false` cases and to map to `boolean` instead of `true`. Before this change a `Button` that had a `loading` variant (and therefore prop) would need to change write the following:

```
<Button loading={conditionalBooleanPropIsLoading as true | undefined}>
   etc
</Button>
```

With this change you will no longer need to do that and can more clearly write:

```
<Button loading={conditionalBooleanPropIsLoading}>
   etc
</Button>
```

This also aligns with stitches' behaviour.

### DIFF
--- a/packages/recipes/src/types.ts
+++ b/packages/recipes/src/types.ts
@@ -4,7 +4,7 @@ type RecipeStyleRule = ComplexStyleRule | string;
 
 export type VariantDefinitions = Record<string, RecipeStyleRule>;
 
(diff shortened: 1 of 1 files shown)


swudged:
Hey @benjervis! This is great, thanks for taking a look and getting a full PR in so quickly. I had only been testing `true` cases hence missing this 😅  Enjoying migrating over to vanilla-extract, thanks a lot!

PR #451, opened 2021-10-30
Add missing @vanilla-extract/css peer dependency

jdanil:
`@vanilla-extract/webpack-plugin` requires `@vanilla-extract/css` but it is not declared as a dependency. This leads to build errors when using strict node linkers like pnp or pnpm.


mattcompiles:
Closing in favour of #459

PR #469, opened 2021-11-13
Recipes: Add `requiredVariants` option

markdalgleish:
See changeset for details.

I know this is something we talked about ages ago, but https://github.com/seek-oss/vanilla-extract/discussions/468 reminded me so I thought I'd give it a go.

I updated TypeScript in this PR because there was a type error that I didn't pick up until it ran in CI because my editor was accidentally using a newer version of TypeScript. Figured it made more sense to upgrade than to figure out why the older version of TypeScript was unhappy.

### DIFF
--- a/.changeset/strong-poets-mate.md
+++ b/.changeset/strong-poets-mate.md
@@ -1,0 +1,25 @@
+---
+'@vanilla-extract/recipes': minor
+---
+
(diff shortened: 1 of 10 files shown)


mattcompiles:
Closing for now as we may go in a different direction.

PR #457, opened 2021-11-09
Vite plugin: Use normalized path as a key for storing processed vanilla css.

Katarina-UBCO:
Fixes #427

When used with Vite, if `*.css.ts` module is imported from _outside_ of the root directory, the following error occurs:

```
Unable to locate /path-to.css.ts$$$@scope/package-name in the CSS map.
```

This happens, because the path `../path-to.css.ts` gets overridden to `/../path-to.css.ts` (by `vite:import-analysis`) and then consequently to `/path-to.css.ts` (by `vite.normalizePath`). As a result, `cssMap` stores the source under a different key as it is being retrieved.

This is where the source is stored under `../path-to.css.ts$$$@scope/package-name`:
https://github.com/seek-oss/vanilla-extract/blob/076b02d91f22d0b61df2dd9edd875f86ac4a55ea/packages/vite-plugin/src/index.ts#L152

This is where the we are trying to retrieve the source by wrong key (`/path-to.css.ts$$$@scope/package-name`):
https://github.com/seek-oss/vanilla-extract/blob/076b02d91f22d0b61df2dd9edd875f86ac4a55ea/packages/vite-plugin/src/index.ts#L69

This PR should fix the problem, as it replaces the use of relative paths as keys with normalise paths.

PR #470, opened 2021-11-13
Vite: Mark `.vanilla.js` and `.vanilla.css` modules as virtual.

### LINKED ISSUE #427: `@vanilla-extract/vite-plugin@^3` + `vite-tsconfig-paths` = Unable to locate /abc/def/vanillaStyles.css.ts$$$xxx in the CSS map.
### Describe the bug
`@vanilla-extract/vite-plugin` @ `v3.0.0` broke compatibility with `vite-tsconfig-paths`. After upgrading to `^3` the following is seen in the console when trying to combine with `vite-tsconfig-paths`

`[vite] Internal server error: Unable to locate /src/default/vanillaStyles.css.ts$$$xxx in the CSS map.`

Downgrading to ^2 fixes the problem.

### Link to reproduction

https://github.com/kompot/vite-vanilla-extract-tsconfig-paths-bug

### System Info

```node
  System:
    OS: macOS 11.5.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 135.09 MB / 16.00 GB
    Shell: 3.3.1 - /usr/local/bin/fish
  Binaries:
    Node: 16.10.0 - /usr/local/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
    npm: 7.24.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 89.1.21.77
    Chrome: 95.0.4638.54
    Edge: 94.0.992.47
    Firefox: 68.0.2
    Firefox Developer Edition: 94.0
    Safari: 14.1.2
    Safari Technology Preview: 15.0
  npmPackages:
    @vanilla-extract/css: ^1 => 1.6.2
    @vanilla-extract/vite-plugin: ^3 => 3.0.0
    @vanilla-extract/webpack-plugin: ^2 => 2.1.0
    vite: ^2.6 => 2.6.10
```

Katarina-UBCO:
Fixes #427 

Also see a different approach in this PR #457 

### The problem

When used with Vite, if `*.css.ts` module is imported from _outside_ of the root directory, the following error occurs:

```
Unable to locate /path-to.css.ts$$$@scope/package-name in the CSS map.
```

This happens, because the path `../path-to.css.ts` gets overridden to `/../path-to.css.ts` (by `vite:import-analysis`) and then consequently to `/path-to.css.ts` (by `vite.normalizePath`). As a result, `cssMap` stores the source under a different key as it is being retrieved.

This is where the source is stored under `../path-to.css.ts$$$@scope/package-name`:
https://github.com/seek-oss/vanilla-extract/blob/076b02d91f22d0b61df2dd9edd875f86ac4a55ea/packages/vite-plugin/src/index.ts#L152

This is where the we are trying to retrieve the source by wrong key (`/path-to.css.ts$$$@scope/package-name`):
https://github.com/seek-oss/vanilla-extract/blob/076b02d91f22d0b61df2dd9edd875f86ac4a55ea/packages/vite-plugin/src/index.ts#L69

### Solution

This PR should fix the problem, as it prefixes the relative paths with `vanilla-extract:` keyword which preserves the paths.

See [Vite Plugin Conventions](https://vitejs.dev/guide/api-plugin.html#conventions) for `virtual:` prefix.

### DIFF
--- a/.changeset/calm-ravens-sit.md
+++ b/.changeset/calm-ravens-sit.md
@@ -2,4 +2,4 @@
 '@vanilla-extract/vite-plugin': patch
 ---
 
-Prefix `.vanilla.js` and `.vanilla.css` virtual modules with `vanilla-extract:`.
+Prefix `.vanilla.js` and `.vanilla.css` virtual modules with `virtual:vanilla-extract:`.
--- a/.changeset/calm-ravens-sit.md
+++ b/.changeset/calm-ravens-sit.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Prefix `.vanilla.js` and `.vanilla.css` virtual modules with `vanilla-extract:`.
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -66,7 +66,9 @@ export function vanillaExtractPlugin({ identifiers }: Options = {}): Plugin {
       const extensionIndex = id.indexOf(virtualExt);
 
       if (id.startsWith(virtualPrefix) && extensionIndex > 0) {
-        const fileScopeId = id.substring(0, extensionIndex).slice(virtualPrefix.length);
+        const fileScopeId = id
+          .substring(0, extensionIndex)
+          .slice(virtualPrefix.length);
 
         if (!cssMap.has(fileScopeId)) {
           throw new Error(`Unable to locate ${fileScopeId} in the CSS map.`);
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -58,17 +58,16 @@ export function vanillaExtractPlugin({ identifiers }: Options = {}): Plugin {
       packageInfo = getPackageInfo(config.root);
     },
     resolveId(id) {
-      if (id.startsWith(virtualPrefix) && id.indexOf(virtualExt) > 0) {
+      if (id.indexOf(virtualPrefix) === 0) {
         return id;
       }
     },
     load(id) {
-      const extensionIndex = id.indexOf(virtualExt);
-
-      if (id.startsWith(virtualPrefix) && extensionIndex > 0) {
-        const fileScopeId = id
-          .substring(0, extensionIndex)
-          .slice(virtualPrefix.length);
+      if (id.indexOf(virtualPrefix) === 0) {
+        const fileScopeId = id.slice(
+          virtualPrefix.length,
+          id.indexOf(virtualExt),
+        );
 
         if (!cssMap.has(fileScopeId)) {
           throw new Error(`Unable to locate ${fileScopeId} in the CSS map.`);
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -16,6 +16,8 @@ import {
 const styleUpdateEvent = (fileId: string) =>
   `vanilla-extract-style-update:${fileId}`;
 
+const virtualPrefix = 'vanilla-extract:';
+
 interface Options {
   identifiers?: IdentifierOption;
 }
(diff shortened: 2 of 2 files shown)


packages/vite-plugin/src/index.ts
```diff
@@ -16,6 +16,8 @@ import {
 const styleUpdateEvent = (fileId: string) =>
   `vanilla-extract-style-update:${fileId}`;
 
+const virtualPrefix = 'vanilla-extract:';
```
markdalgleish, line 19: Based on the plugin conventions you linked to, should this prefix be `virtual:vanilla-extract/`?

Katarina-UBCO: You're right, it looks like that it should be `virtual:vanilla-extract/` + path. (I first understood it as `virtual:` + path, and virtual could be replaced with the namespace, e.g. `vanilla-extract`). Anyway, it doesn't work with `virtual:vanilla-extract/` due to the `/` in the end. The relative path is being resolved incorrectly, e.g.

```
virtual:vanilla-extract/../design-system/src/theme.css.ts$$$@***/components.vanilla.js
```

is being resolved to 

```
/@id/design-system/src/theme.css.ts$$$@***/components.vanilla.js
```

which is resolved to

```
design-system/src/theme.css.ts$$$@***/components.vanilla.js
```

Which doesn't exist.

Removing `/` from `virtual:vanilla-extract/` (or replacing it with `:` or `-`) solves the problem. Maybe we should stick to something like `/@ve-css` (as described in https://github.com/seek-oss/vanilla-extract/pull/457) or use something like `/@vanilla-extract` ?

Thanks heaps for your help and love your work 🥇

packages/vite-plugin/src/index.ts
```diff
@@ -56,15 +58,17 @@ export function vanillaExtractPlugin({ identifiers }: Options = {}): Plugin {
       packageInfo = getPackageInfo(config.root);
     },
     resolveId(id) {
-      if (id.indexOf(virtualExt) > 0) {
+      if (id.startsWith(virtualPrefix) && id.indexOf(virtualExt) > 0) {
```
mattcompiles, line 61: Think we can drop the virtualExt check now as it's basically redundant. Also we prefer to use `indexOf` instead of `startsWith` as it seems to be quite a lot more performant.

packages/vite-plugin/src/index.ts
```diff
@@ -56,15 +58,17 @@ export function vanillaExtractPlugin({ identifiers }: Options = {}): Plugin {
...
         return id;
       }
     },
     load(id) {
       const extensionIndex = id.indexOf(virtualExt);
 
-      if (extensionIndex > 0) {
-        const fileScopeId = id.substring(0, extensionIndex);
+      if (id.startsWith(virtualPrefix) && extensionIndex > 0) {
```
mattcompiles, line 68: Think we can drop the `extensionIndex` check here as well as it's now redundant. This means we can pull the `extensionIndex` variable into the if block scope as well.

packages/vite-plugin/src/index.ts
```diff
@@ -56,15 +58,17 @@ export function vanillaExtractPlugin({ identifiers }: Options = {}): Plugin {
...
+        const fileScopeId = id
+          .substring(0, extensionIndex)
+          .slice(virtualPrefix.length);
```
mattcompiles, line 71: This can be done through one call to `slice`. e.g. `slice(virtualPrefix.length, extensionIndex)`

mattcompiles (commented):
This approach is seeming much cleaner. 👍  Just a couple of small implementation comments.

packages/vite-plugin/src/index.ts
mattcompiles: Could we make the prefix `virtual:vanilla-extract:` then?

Katarina-UBCO: @mattcompiles set to `virtual:vanilla-extract:` thank you

packages/vite-plugin/src/index.ts
Katarina-UBCO: Very good point. Done. Thank you

packages/vite-plugin/src/index.ts
Katarina-UBCO: Done these, too.

packages/vite-plugin/src/index.ts
Katarina-UBCO: Done these, too.

.changeset/calm-ravens-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Prefix `.vanilla.js` and `.vanilla.css` virtual modules with `vanilla-extract:`.
```
mattcompiles, line 5:
```suggestion
Prefix `.vanilla.js` and `.vanilla.css` virtual modules with `virtual:vanilla-extract:`.
```

mattcompiles (approved):
Awesome work. 👏  Thanks for all the effort.

Katarina-UBCO:
Thanks heaps @mattcompiles 🎉 Team work!

PR #490, opened 2021-11-23
Fix version script for yarn 3

mattcompiles:
Yarn 3 infers the `--immutable` flag in CI environments. This prevents the version command from updating the lockfile for the version packages PR.


benjervis (approved):
🏃‍♂️

Issue #411, opened 2021-10-12
Invalid className when issued from file with bracketed name


ksesong:
### Describe the bug
When setting styles on a `.css.ts` file that includes brackets in its name like `[test].css.ts`, the resulting styles from the output stylesheet have an invalid className, with a missing dot, such that the styles cannot be applied.

### Link to reproduction
This [sandbox](https://codesandbox.io/s/awesome-turing-3j88x) reproduces the issue. Both `divs` should be yellow, but the file whose name uses brackets gets processed incorrectly. 

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

```node
  System:
    OS: Linux 5.4 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 3.27 GB / 62.73 GB
    Container: Yes
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.6/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.17.6/bin/npm
  npmPackages:
    @vanilla-extract/css: 1.6.1 => 1.6.1
    @vanilla-extract/sprinkles: ^1.3.0 => 1.3.0
    @vanilla-extract/webpack-plugin: ^2.1.0 => 2.1.0
    webpack: ^5.36.1 => 5.58.1
```

PR #491, opened 2021-11-23
fix: classnames with brackets in files

Schubidu:
this fixes classnames with brackets from filename as in #411 mentioned. Improves better compatibility with nextjs


packages/css/src/identifier.ts
```diff
@@ -16,7 +16,7 @@ function getDevPrefix(debugId: string | undefined) {
     parts.unshift(file && file !== 'index' ? file : dir);
   }
 
-  return parts.join('_');
+  return parts.join('_').replace(/([^\w\d])/gm, '_');
```
mattcompiles, line 19: This fix just replaces all non word/decimal characters with `_`, even `-` characters which are commonly used. Instead of just replacing characters we've previously opted to escape them using [cssesc](https://www.npmjs.com/package/cssesc). 

Not sure why that's not working here but needs to be investigated.

mattcompiles:
Closing in favour of #504

PR #533, opened 2021-12-17
Added fallback properties usage example

goodwin64:


README.md
```diff
@@ -850,7 +850,16 @@ export const primaryColorVar = createVar();
 export const secondaryColorVar = createVar();
 
 export const exampleStyle = style({
-  color: fallbackVar(primaryColorVar, secondaryColorVar, 'blue');
+  color: fallbackVar(primaryColorVar, secondaryColorVar, 'blue'),
+});
+```
+
+For fallback styles you may simply pass an array of properties instead of a single prop.
```
mattcompiles, line 857: Feel like this part should be part of the the style API section. Maybe just before the "Multiple styles can be composed into a single rule by providing an array of styles" section?

mattcompiles (commented):
Thanks for this. These docs are currently replicated between the README and the site/docs folder which powers the website. Could you make the updates there as well?

README.md
goodwin64: Yup! No prob, doable!

Before this one
image

or this?
image

TheMightyPenguin: @mattcompiles @goodwin64 👀

mattcompiles: Sorry I missed this. I think the former option would be fine.

mattcompiles:
Hey @goodwin64, sorry but looks like the prettier check is failing. You just need to run a `format` and it should be good.

site/docs/styling-api.md
```diff
@@ -81,6 +81,15 @@ export const childClass = style({
...
 > If you want to globally target child nodes within the current element (e.g. `'& a[href]'`), you should use [`globalStyle`](#globalstyle) instead.
 
+For fallback styles you may simply pass an array of properties instead of a single prop.
+
+```ts
+export const exampleStyle = style({
+  // in Firefox and IE the "overflow: overlay" will be ignored and the "overflow: auto" will be applied
+  overflow: ['auto', 'overlay'],
+});
+```
```
michaeltaranto, line 91: Adding title to code block and import statement, along with formatting fix

```suggestion
```ts
// styles.css.ts

export const exampleStyle = style({
  // in Firefox and IE the "overflow: overlay" will be ignored and the "overflow: auto" will be applied
  overflow: ['auto', 'overlay']
});
```
```

michaeltaranto (approved):

PR #542, opened 2021-12-21
Add Sprinkles usage to docs site

mattcompiles:


site/docs/sprinkles-api.md
```diff
@@ -8,10 +8,234 @@ Generate a static set of custom utility classes and compose them either statical
...
+    flexDirection
+  })}">
+    ...
+  </section>
+`);
+```
+
+> 💡 Although you don’t need to use this library at runtime, it’s designed to be as small and performant as possible. The runtime is only used to look up pre-existing class names. All styles are still generated at build time!
+
+Within `.css.ts` files, combine with any custom styles by providing an array to vanilla-extract’s [`style`](https://vanilla-extract.style/documentation/styling-api/#style) function.
```
michaeltaranto, line 203:
```suggestion
Within `.css.ts` files, combine with any custom styles by providing an array to vanilla-extract’s [`style`](/documentation/styling-api/#style) function.
```

michaeltaranto (approved):
One minor tweak :rocket:

PR #607, opened 2022-03-03
docs(sprinkles): fix typo

ubmit:
Use "was" instead of "were"


benjervis:
Because this line describes an alternate way of thinking, rather than reality, it uses the subjunctive mood. "Were" is correct in this case.

https://www.grammarly.com/blog/was-vs-were/

PR #621, opened 2022-03-17
Compress CSS resources to keep the virtual size down

### LINKED ISSUE #374: [webpack.cache.PackFileCacheStrategy] Serializing big strings (110kiB) impacts deserialization performance
### Describe the bug

Strangely enough, adding more sprinkles (see [this commit](https://github.com/moroshko/webpack-cache-issue/commit/843ea3a26a4a36d92c5c851918414512081a56f8)) causing the following warning to appear in Node's console:

```
<w> [webpack.cache.PackFileCacheStrategy] Serializing big strings (110kiB) impacts deserialization 
performance (consider using Buffer instead and decode when needed)
```

### Link to reproduction

1. `git clone https://github.com/moroshko/webpack-cache-issue.git`
2. `cd webpack-cache-issue`
3. `yarn`
4. `yarn dev`
5. Observe the warning

Note: You might need to `rm -rf .next` to see the warning.

### System Info

```node
  System:
    OS: macOS 11.5.2
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 558.97 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.6 - ~/.nvm/versions/node/v14.17.6/bin/node
    Yarn: 1.22.11 - ~/.nvm/versions/node/v14.17.6/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.17.6/bin/npm
  Browsers:
    Chrome: 93.0.4577.82
    Firefox: 92.0
    Safari: 14.1.2
  npmPackages:
    @vanilla-extract/babel-plugin: 1.1.0 => 1.1.0 
    @vanilla-extract/css: 1.6.0 => 1.6.0 
    @vanilla-extract/next-plugin: 1.0.0 => 1.0.0 
    @vanilla-extract/recipes: 0.1.0 => 0.1.0 
    @vanilla-extract/sprinkles: 1.3.0 => 1.3.0 
```

nayaabkhan:
This also removes the dependency on the virtual-resource-loader and moves it internally for colocating compression code.

Resolves #374

### DIFF
--- a/.changeset/friendly-turkeys-play.md
+++ b/.changeset/friendly-turkeys-play.md
@@ -1,6 +1,6 @@
 ---
-'@vanilla-extract/integration': minor
-'@vanilla-extract/webpack-plugin': minor
+'@vanilla-extract/integration': patch
+'@vanilla-extract/webpack-plugin': patch
 ---
 
 Improve build performance when creating large CSS files
--- a/.changeset/friendly-turkeys-play.md
+++ b/.changeset/friendly-turkeys-play.md
@@ -1,0 +1,6 @@
+---
+'@vanilla-extract/integration': minor
+'@vanilla-extract/webpack-plugin': minor
+---
+
+Compress CSS resources to keep the virtual size down
--- a/.changeset/friendly-turkeys-play.md
+++ b/.changeset/friendly-turkeys-play.md
@@ -3,4 +3,4 @@
 '@vanilla-extract/webpack-plugin': minor
 ---
 
-Compress CSS resources to keep the virtual size down
+Improve build performance when creating large CSS files
--- a/packages/integration/src/virtualFile.ts
+++ b/packages/integration/src/virtualFile.ts
@@ -1,3 +1,5 @@
+import zlib from 'zlib';
+
 export function getSourceFromVirtualCssFile(id: string) {
   const match = id.match(/^(?<fileName>.*)\?source=(?<source>.*)$/) ?? [];
 
@@ -5,8 +7,12 @@ export function getSourceFromVirtualCssFile(id: string) {
     throw new Error('No source in vanilla CSS file');
   }
 
+  const decompressedSource = zlib.gunzipSync(
+    Buffer.from(match.groups.source, 'base64'),
+  );
+
   return {
     fileName: match.groups.fileName,
-    source: Buffer.from(match.groups.source, 'base64').toString('utf-8'),
+    source: decompressedSource.toString('utf-8'),
   };
 }
--- a/packages/webpack-plugin/package.json
+++ b/packages/webpack-plugin/package.json
@@ -17,19 +17,25 @@
     "./extracted": {
       "module": "./extracted/dist/vanilla-extract-webpack-plugin-extracted.esm.js",
       "default": "./extracted/dist/vanilla-extract-webpack-plugin-extracted.cjs.js"
+    },
+    "./virtual": {
+      "module": "./virtual/dist/vanilla-extract-webpack-plugin-virtual.esm.js",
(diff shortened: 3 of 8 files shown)


packages/webpack-plugin/package.json
```diff
@@ -17,19 +17,25 @@
     "./extracted": {
       "module": "./extracted/dist/vanilla-extract-webpack-plugin-extracted.esm.js",
       "default": "./extracted/dist/vanilla-extract-webpack-plugin-extracted.cjs.js"
+    },
```
nayaabkhan, line 20: Wondering if this is the [desired place](https://github.com/seek-oss/vanilla-extract/issues/374#issuecomment-1060046223) for having the virtual loader. I took a best guess to start out with.

.changeset/friendly-turkeys-play.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/integration': minor
+'@vanilla-extract/webpack-plugin': minor
```
mattcompiles, line 3: I think this should just be a patch release as it doesn't add any new API

.changeset/friendly-turkeys-play.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Compress CSS resources to keep the virtual size down
```
mattcompiles, line 6: I think this is misleading from a consumer of VE's perspective. How about:
```suggestion
Improve build performance when creating large CSS files
```

packages/webpack-plugin/package.json
mattcompiles: I agree it should live here but could we name it something a bit clearer. How about `virtualFileLoader`?

packages/integration/src/virtualFile.ts
```diff
@@ -1,12 +1,18 @@
+import zlib from 'zlib';
+
 export function getSourceFromVirtualCssFile(id: string) {
   const match = id.match(/^(?<fileName>.*)\?source=(?<source>.*)$/) ?? [];
 
   if (!match || !match.groups) {
     throw new Error('No source in vanilla CSS file');
   }
 
+  const decompressedSource = zlib.gunzipSync(
```
mattcompiles, line 10: Not a huge deal but maybe we could create some shared `compress` and `decompress` functions just to make sure all the places dealing with compression are using the same base64 and compression logic? You could export it from the integration package and use it in the virtual loader code.

packages/webpack-plugin/src/virtual.ts
```diff
@@ -0,0 +1,11 @@
+// @ts-expect-error
+import { getOptions } from 'loader-utils';
+import zlib from 'zlib';
+
+export default function (this: any) {
+  const { source } = getOptions(this);
+
+  let decompressedSource = zlib.gunzipSync(Buffer.from(source, 'base64'));
```
mattcompiles, line 8: Small one but can we use `const` just for consistency?

mattcompiles (commented):
A few comments mate but this is an excellent PR 👏 . 
Thanks so much for your help on this. 🍻

.changeset/friendly-turkeys-play.md
mattcompiles, line 3:
```suggestion
'@vanilla-extract/integration': patch
'@vanilla-extract/webpack-plugin': patch
```

mattcompiles (approved):
Thanks for this mate. Will fix up any remaining comments post merge. 👍

nayaabkhan:
Real glad to get this through! 🍻

PR #630, opened 2022-03-22
Add Rollup plugin

graup:
Rollup plugin that compiles vanilla extract files and emits separate CSS files. This does not do any CSS bundling by itself.

Based on a combination of vite-plugin and https://github.com/stropho/rollup-plugin-copy-imported-assets


packages/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+  serializeCss,
+} from '@vanilla-extract/integration';
+import { relative, normalize, dirname } from 'path';
+
+const virtualPrefix = 'virtual:vanilla-extract:';
+const vanillaSuffix = '.vanilla.css';
+
+interface Options {
+  identifiers?: IdentifierOption;
+  cwd?: string;
```
mattcompiles, line 17: Most other bundlers have a built-in idea of project root or cwd. Does rollup not?

mattcompiles (commented):
This looks pretty good. Made a few comments to start with. If we did look to merge this then I think we'd have to make the intent of the plugin pretty clear, in that it's only meant for package bundling.

packages/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+export function vanillaExtractPlugin({ identifiers, cwd = process.cwd() }: Options = {}): Plugin {
+  const emittedFiles = new Map<string, string>();
+  const isProduction = process.env.NODE_ENV === 'production';
+
+  return {
+    name: 'vanilla-extract',
+    outputOptions(outputOptions) {
+      if (outputOptions.format && !['es', 'esm', 'module'].includes(outputOptions.format)) {
+        this.warn('Only format \'esm\' is supported, others should NOT be used with this plugin');
```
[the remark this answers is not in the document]
graup: Hi, thanks for looking into this already! For some reason it doesn't work in CJS but I don't know why, will find out!

packages/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,96 @@
...
+      return processVanillaFile({
+        source,
+        filePath,
+        identOption: identifiers ?? (isProduction ? 'short' : 'debug'),
+        async serializeVirtualCssPath({ fileScope, source }) {
+          // This is almost identical to processVanillaFile's default behavior,
+          // minus the package name in the fileName, in order to
+          // place emitted assets at more sensible paths.
+          const serializedCss = await serializeCss(source);
+          const fileName = `${fileScope.filePath}${vanillaSuffix}`;
```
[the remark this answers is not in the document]
graup: I think because this puts the files at their original path it should work, but will test.

packages/rollup-plugin/src/index.ts
graup: I could not find anything related to this. But it seems to work fine with `process.cwd`. Not sure this option is needed.

packages/rollup-plugin/src/index.ts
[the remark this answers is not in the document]
graup: Tested again and seems there are no problems with CJS. Removed this check.

graup:
Hi @mattcompiles , I got this code working.

Here's an example monorepo with UI library and both nextjs and vite example apps: https://github.com/graup/vanilla-extract-rollup-example

Let me know if you get this running and what your thoughts are. Feel free to DM me on Discord (`@graycoding`).

packages/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,83 @@
...
+          name: fileName,
+          source,
+        });
+        emittedFiles.set(fileName, assetId);
+      }
+      const assetId = emittedFiles.get(fileName);
+
+      // Resolve to a temporary virtual external chunk
+      return {
+        id: `${virtualPrefix}${assetId}`,
```
mattcompiles, line 67: Could we just return the `fileName` here seeing as it's external? That would mean we could remove all the `virtualPrefix` stuff and just use the `emittedFiles` Map for all the `renderChunk` logic.

graup: Done!

JLee21:
Hey all! 
I'm just checking on the Review status for this to be merged...
Is there a ETA? 
My team is excited to use it!

graup:
@JLee21 I need to add tests and documentation, planning to get that done by early next week!

graup:
Hi @mattcompiles , I added snapshots tests for a few rollup configs as a fixture package. I also added some basic docs. The site version has some more details (wasn't sure if I should duplicate it all to the Readme or not).

packages/rollup-plugin/test/rollup-plugin.test.ts
```diff
@@ -0,0 +1,59 @@
+import { rollup, OutputOptions } from 'rollup';
+import esbuild from 'rollup-plugin-esbuild';
+import json from '@rollup/plugin-json';
+import path from 'path';
+
+import { vanillaExtractPlugin } from '..';
+
+async function buildAndMatchSnapshot(outputOptions: OutputOptions) {
```
mattcompiles, line 8: Rather than create a whole new fixture, I moved the tests to live with the plugin itself and use the existing themed fixture.

mattcompiles (approved):
Nice one pushing this through mate. 🍻 🎉

JLee21:
Woo woo 🎉

Issue #631, opened 2022-03-22
Multiple CSS files have the same name when using the Vite plugin with a scoped package


AndrewLeedham:
### Describe the bug
If the package name includes a scope e.g `@vanilla-extract/example` the outputted CSS files will all have the names `example.vanilla.css`, `example.vanilla2.css`, `example.vanilla3.css` etc. This seems to be because the [sanitise file scope function](https://github.com/seek-oss/vanilla-extract/blob/4d45d8209bd26a9b2a8f9b5b8dfee9d420c3e45f/packages/integration/src/processVanillaFile.ts#L18) adds the package name and when a scope is used it includes a `/` so it considers the last part as the file name. I would expect the file name to be included in the asset name, which it is when not using a scope.

### Link to reproduction
https://stackblitz.com/edit/vitejs-vite-gfnuxw?file=vite.config.ts
- Run `npm run build` in the above and it outputs:
  ```
  vite v2.8.4 building for production...
  ✓ 9 modules transformed.
  dist/main.js                0.25 KiB / gzip: 0.20 KiB
  dist/main.css.js            0.18 KiB / gzip: 0.16 KiB
  dist/Button.js              0.13 KiB / gzip: 0.13 KiB
  dist/Heading.js             0.14 KiB / gzip: 0.14 KiB
  dist/Button.css.js          0.18 KiB / gzip: 0.16 KiB
  dist/Heading.css.js         0.18 KiB / gzip: 0.16 KiB
  dist/example.vanilla.css    0.23 KiB / gzip: 0.16 KiB
  dist/example.vanilla2.css   0.12 KiB / gzip: 0.13 KiB
  dist/example.vanilla3.css   0.04 KiB / gzip: 0.06 KiB
  ```
- However, if you remove the scope from `package.json->name`. It outputs the following:
  ```
  dist/main.js                                0.25 KiB / gzip: 0.20 KiB
  dist/main.css.js                            0.19 KiB / gzip: 0.15 KiB
  dist/Button.js                              0.13 KiB / gzip: 0.13 KiB
  dist/Heading.js                             0.14 KiB / gzip: 0.14 KiB
  dist/Button.css.js                          0.20 KiB / gzip: 0.16 KiB
  dist/Heading.css.js                         0.20 KiB / gzip: 0.16 KiB
  dist/Button.css.ts___example.vanilla.css    0.23 KiB / gzip: 0.17 KiB
[...]

PR #632, opened 2022-03-22
vite: Sanitise virtual import path characters

AndrewLeedham:
fixes: #631 

Speaking with @graup this issue also occurs in the file path, when you have multiple style files called the same thing, for example `component1/style.css.ts` and `component2/style.css.ts`. So this sanitises both by replacing slashes and double dots with underscores. I moved the double dot fix from the last release up to `stringifyFileScope` as part of this.


AndrewLeedham:
After playing around with a patched version of this, it does not look like this works in all scenarios. It seems to break module graph stuff in server mode.

PR #638, opened 2022-03-24
Fix GitHub star count & broken twitter avatar

michaeltaranto:
~Also opting for `bigger` twitter avatars (73x73) given we only display at 60x60. Perf ftw~ 🏎️ 

Opting for 200x200 twitter avatars because high DPI. Still perf though 🐌

### DIFF
--- a/site/src/GitHubStars/GitHubStars.tsx
+++ b/site/src/GitHubStars/GitHubStars.tsx
@@ -25,14 +25,13 @@ export const GitHubStars = () => {
       );
       const { stargazers_count: count } = await res.json();
 
-      let starCount = null;
       if (typeof count === 'number') {
-        starCount =
+        setStars(
           Math.abs(count) > 999
             ? `${(Math.sign(count) * (Math.abs(count) / 1000)).toFixed(1)}k`
-            : `${Math.sign(count) * Math.abs(count)}`;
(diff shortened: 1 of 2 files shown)


site/src/GitHubStars/GitHubStars.tsx
```diff
@@ -25,11 +25,14 @@ export const GitHubStars = () => {
...
-          : `${Math.sign(count) * Math.abs(count)}`,
-      );
+      let starCount = null;
+      if (typeof count === 'number') {
+        starCount =
+          Math.abs(count) > 999
+            ? `${(Math.sign(count) * (Math.abs(count) / 1000)).toFixed(1)}k`
+            : `${Math.sign(count) * Math.abs(count)}`;
+      }
+      setStars(starCount);
```
mattcompiles, line 35: Should we just call `setStars` conditionally? No reason to call `setStars(null)` yeah?

mattcompiles (approved):
Made a comment but not gonna block over it.

site/src/GitHubStars/GitHubStars.tsx
michaeltaranto: Just realized we only call this the once on mount 🤦‍♂️, yep I'll update

PR #651, opened 2022-04-10
feat: validate media queries

amorriscode:
This PR adds some basic CSS media query linting (see: https://github.com/seek-oss/vanilla-extract/discussions/480).

At work, we've shipped a couple of bugs because an invalid media query wasn't caught during code review (eg. `@media (max-width: 600px`). CSSType doesn't catch this, nor does Vanilla Extract. It [looks like CSSType might tackle it in the future](https://github.com/frenic/csstype/issues/66#issuecomment-1004090875).

I wanted to see if I could put something quick together this weekend so this is that attempt. I also tried doing a RegEx attempt but that ended up getting pretty unwieldy (similar to this: https://github.com/niksy/regex-css-media-query).

Would love to hear if you had any ideas on how to tackle this!

### DIFF
--- a/.changeset/moody-poems-tan.md
+++ b/.changeset/moody-poems-tan.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': minor
+---
+
+Add CSS media query validation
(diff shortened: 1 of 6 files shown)


mattcompiles (approved):
This is a great PR mate 👏. 

Really appreciate the time taken to follow the design of the rest of the repo, specifically the error message styling. 🍻 🚀

PR #657, opened 2022-04-18
Permit custom identifier hashing

FrancoisChabot:
Simple custom identifier hashing. Not quite the exact same feature request, but related: https://github.com/seek-oss/vanilla-extract/discussions/313

This may not be the best possible API for this, but it has the benefit of being a backwards-compatible drop-in for any adapter/plugin that already exposes `IdentifierOption` directly.

For example,  with the vite plugin, one can do:

```
plugins: [vanillaExtractPlugin({
      identifiers: (scope, index) => { return 'insert_logic_here'; }
    })],
```

### DIFF
--- a/packages/css/src/identifier.ts
+++ b/packages/css/src/identifier.ts
@@ -19,24 +19,46 @@ function getDevPrefix(debugId: string | undefined) {
   return parts.join('_');
 }
 
-export function generateIdentifier(debugId: string | undefined) {
+function generateShortIdentifier(scope: string, index: number) {
   // Convert ref count to base 36 for optimal hash lengths
-  const refCount = getAndIncrementRefCounter().toString(36);
+  const refCountStr = index.toString(36);
+  const fileScopeHash = hash(scope);
+  return `${fileScopeHash}${refCountStr}`;
+}
+
+function generateDebugIdentifier(
+  scope: string,
+  index: number,
+  debugId: string | undefined,
+) {
+  let identifier = generateShortIdentifier(scope, index);
+  const devPrefix = getDevPrefix(debugId);
+  if (devPrefix) {
+    identifier = `${devPrefix}__${identifier}`;
+  }
+  return identifier;
+}
+
+export function generateIdentifier(debugId: string | undefined) {
+  const refCount = getAndIncrementRefCounter();
   const { filePath, packageName } = getFileScope();
+  const fileScopeStr = packageName ? `${packageName}${filePath}` : filePath;
+  const opt = getIdentOption();
 
-  const fileScopeHash = hash(
-    packageName ? `${packageName}${filePath}` : filePath,
-  );
+  let identifier: string;
+  if (opt === 'short') {
+    identifier = generateShortIdentifier(fileScopeStr, refCount);
+    identifier = identifier.match(/^[0-9]/) ? `_${identifier}` : identifier;
+  } else if (opt === 'debug') {
+    identifier = generateDebugIdentifier(fileScopeStr, refCount, debugId);
+    identifier = identifier.match(/^[0-9]/) ? `_${identifier}` : identifier;
+  } else {
+    identifier = opt(fileScopeStr, refCount, debugId);
 
-  let identifier = `${fileScopeHash}${refCount}`;
-
-  if (getIdentOption() === 'debug') {
-    const devPrefix = getDevPrefix(debugId);
-
-    if (devPrefix) {
-      identifier = `${devPrefix}__${identifier}`;
+    if (!identifier.match(/^[A-Z_][0-9A-Z_]+$/i)) {
+      throw new Error(`Invalid identifier:"${identifier}"`);
     }
   }
 
-  return identifier.match(/^[0-9]/) ? `_${identifier}` : identifier;
+  return identifier;
 }
--- a/packages/css/src/identifier.ts
+++ b/packages/css/src/identifier.ts
@@ -56,7 +56,7 @@ export function generateIdentifier(debugId: string | undefined) {
     identifier = opt(fileScopeStr, refCount, debugId);
 
     if (!identifier.match(/^[A-Z_][0-9A-Z_]+$/i)) {
-      throw new Error(`Invalid identifier:"${identifier}"`);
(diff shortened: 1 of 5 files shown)


packages/css/src/identifier.ts
```diff
@@ -19,24 +19,46 @@ function getDevPrefix(debugId: string | undefined) {
...
+  } else if (opt === 'debug') {
+    identifier = generateDebugIdentifier(fileScopeStr, refCount, debugId);
+    identifier = identifier.match(/^[0-9]/) ? `_${identifier}` : identifier;
+  } else {
+    identifier = opt(fileScopeStr, refCount, debugId);
 
-    if (devPrefix) {
-      identifier = `${devPrefix}__${identifier}`;
+    if (!identifier.match(/^[A-Z_][0-9A-Z_]+$/i)) {
+      throw new Error(`Invalid identifier:"${identifier}"`);
```
graup, line 59:
```suggestion
      throw new Error(`Identifier function returned invalid indentifier: "${identifier}"`);
```

packages/css/src/identifier.ts
```diff
@@ -19,24 +19,46 @@ function getDevPrefix(debugId: string | undefined) {
...
     }
   }
 
-  return identifier.match(/^[0-9]/) ? `_${identifier}` : identifier;
```
graup, line 41: Any reason why you chose to not apply this to user-defined identifiers? I think it'd be safe to always apply this.

FrancoisChabot: My reasoning was that only "fixing" a subset of invalid identifiers into valid ones would lead to surprises and confusion. That check is fine for the baked-in hash function because a numeric prefix just happens to be the only way the identifier can be broken, that's not the case for callback-driven ones.

I think the code should either convert any string into a valid identifier, or just check whether the produced identifier is valid or not. I went for the later.

mattcompiles:
Hi 👋 

What is the scenario in which we would suggest utilising this feature? I'm not completely against adding it, but I don't want to add features just because we can. In the past when this was suggested our team normally felt that other options would be a better solution to the problem.

douglaseggleton:
Might be a relevant use-case - we use storyshots and the ability to have a consistent readable debug instead of hash would be helpful.

douglaseggleton:
Might be a relevant use-case - we use storyshots and the ability to have a consistent readable debug instead of hash would be helpful.

plesiecki:
I suppose with that feature I would be able to migrate CSS Modules (with custom hashing fn) codebase to vanilla-extract 🤔

CroModder:
This one is a must have for us. The use case is that there are multiple parts on the page (micro frontends) that are using different versions of the shared lib that's using vanilla extract as a styling solution. And when multiple versions are being run simultaneously, it happens that classname hashes are not unique. In our old lib we used CSS modules and appended library version on the generated hash which assured us from hash clashes, but it is also very helpful when debugging...

dan-harris:
I'd also like to add a usecase for this PR - generating some stable classnames as part of a Design System implementation.

Currently exploring an exact approach, but would love to use vanilla extract for a range of usecases across 'ts-focussed' components, along with produce some 'template-focussed' utility css / shared classes. Having control of identifiers would give us the choice as to how we implement and where we buy in on the safety of the generated classnames.

mwood23:
+1 for this one. We embed widgets onto websites and need to add a unique identifier to ensure there will never be clashes on classnames since we don't own the environment.

juniorconquista:
Hi 👋
Do we have any news about this feature? He very helpful!

PatrickMennen:
@mattcompiles in issue #841 you mention that this goes against the philosophy of a solution like vanilla-extract. 

Does this mean this feature is not being considered whatsoever? This discussion illustrates that developers have a need to disable (or customize) the scoping of the output class names manually. 

Could we at least get an official response on this?

altrim:
Hi,

Any new updates on this one? 
We have a similar use case where we have a micro frontend solution and multiple views are using different versions of the shared library (built with vanilla extract). 
Having the option to customise the identifier would allow us to generate different styling on the version to avoid clashes.

bjgrosse:
I would really like to generate atomic class names with a hash of the selector+property name, etc. This is for deduplicating CSS atomic classes so that the last one applied takes effect.

tom-scott-carsales:
+ 1, the inability to customise the class name or hashing impedes our ability to use Vanilla Extract outside of Node or support tailwind style utility class names, conflicts can be managed outside of Vanilla Extract given the chance.

PatrickMennen:
just bumping this one ;-)

otaviomad:
This PR itself needs some updating. 
Might I suggest, if this ever goes forward, that we also get access to devPrefix in the identifier function?

mihkeleidast:
One other usage here: we'd like to keep className hashing as is, but disable the variable name hashing completely. We support using the variables in both regular CSS and vanilla extract, which means we need to provide and apply both the non-hashed variables and vanilla-extract variables (essentially duplicated code). Since we support regular CSS and using non-hashed variables there, we don't care much about scoping vanilla extract variables either, they are global anyway,.

greenskybluephish:
@mattcompiles Any chance you could outline or link to what those better alternatives would be? Similar to some commenters above, I'm on the cusp of releasing an internal component library, which will be used in a micro-frontend architecture. During our beta, we've hit issues with apps having differing versions of our package installed, as class names are the same but their properties can be different. 

With just the ability to add an additional versioning string to the class names, I believe we would eliminate the problem altogether, but as it stands we are looking at much less desirable alternatives, like needing to scope our CSS under version classes, i.e. 

```css
.v2-0-0 {
import '@component-lib/styles.css'
}
```

wesgro:
This would be an incredible addition for Design System authors that have very little control over where and what version of their css ends up on any given page.

jdecastroc:
@mattcompiles is there any update in regards this PR?

This still being a mandatory requirement to have vanilla-extract working properly in the scenario where several microfrontends using internally vanilla-extract, are renderer in the same page. The hash-clash is happening right now making the need to have custom hashes real.

Would love to see how this could progress or what is needed to be done :D

mattcompiles:
CLosing in favour of #1160

PR #661, opened 2022-04-20
Replace __dirname and __filename in .css.ts file source

AndrewLeedham:
Allows `__filename` and `__dirname` to be used in `.css.ts` files. Resolving this repro:  https://github.com/AndrewLeedham/vanilla-extract-filename

It uses a somewhat naive approach of regex replacement since that is what Vite does. The `'__' + 'filename'` instead of just `'__filename'` is intended to prevent Vite from replacing the replacement when using the plugin 🙄

### DIFF
--- a/.changeset/grumpy-feet-listen.md
+++ b/.changeset/grumpy-feet-listen.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/integration': patch
+---
+
+Replace **dirname and **filename in .css.ts file source
--- a/packages/integration/src/compile.ts
+++ b/packages/integration/src/compile.ts
@@ -60,7 +60,12 @@ export async function compile({
   }
 
   return {
-    source: outputFiles[0].text,
+    source: outputFiles[0].text
+      .replace(new RegExp('__' + 'filename', 'g'), JSON.stringify(filePath))
+      .replace(
+        new RegExp('__' + 'dirname', 'g'),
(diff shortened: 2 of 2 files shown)


mattcompiles:
The internal esbuild compile bundles all the internal files into a single output. This solution would lead to `__filename` and `__dirname` being incorrect for all but the entrypoint file. It would also lead to the same file having different values if it was imported in multiple places. To solve this you'd need to replace at the file level with an esbuild plugin that run for each internal file. 

Outside of the solution though, what scenario would having `__filename` and `__dirname` be useful for?

AndrewLeedham:
> Outside of the solution though, what scenario would having `__filename` and `__dirname` be useful for?

I am exporting the filename of each `.css.ts` file, so when I import it in each component I can reference it and use it in a React hook which handles including the resulting CSS file on the page. Which is handled by a path transform on said filename.

It is something along the lines of
```typescript
export const stylesheetPath = getStylesheetPath(__filename);
```

> The internal esbuild compile bundles all the internal files into a single output. This solution would lead to `__filename` and `__dirname` being incorrect for all but the entrypoint file. It would also lead to the same file having different values if it was imported in multiple places. To solve this you'd need to replace at the file level with an esbuild plugin that run for each internal file.

I created https://github.com/seek-oss/vanilla-extract/pull/668 for different reasons, but that should address the bundling/entry issue.

PR #647, opened 2022-04-05
Use realistic file paths in file scopes

mattcompiles:
This PR is the start of getting rid of package names in file scopes, with the intention of using more realistic file paths for the app being compiled. This hopefully solves (or helps to solve) a bunch of issues with file resolution and file chunk naming.

Fixes #462
Fixes #631

PR #662, opened 2022-04-21
Add `globalImport`

graup:
Based on the discussion at https://github.com/seek-oss/vanilla-extract/discussions/600

The rule format mirrors the CSS spec: https://developer.mozilla.org/en-US/docs/Web/CSS/@import

### DIFF
--- a/.changeset/great-bears-lay.md
+++ b/.changeset/great-bears-lay.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': minor
+---
+
+Added globalImport to add global import rules (e.g. font imports from Google Fonts) to CSS.
--- a/packages/css/src/style.ts
+++ b/packages/css/src/style.ts
@@ -9,6 +9,7 @@ import type {
   GlobalStyleRule,
   ClassNames,
   ComplexStyleRule,
+  GlobalImportRule,
 } from './types';
 import {
   registerClassName,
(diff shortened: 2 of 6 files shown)


mattcompiles:
Couple of concerns with adding an `@import` API: 

- You can use it to import other local CSS files which most bundlers will try to interpret. This leads to scenarios that will break as relative imports will not work currently. If we merge #647 this scenario may be able to be improved but not sure it'd ever be a good idea anyway. 
- In what scenario is using CSS import an ideal solution? As mentioned in the [linked discussion](https://github.com/seek-oss/vanilla-extract/discussions/600#discussioncomment-2227318), using CSS import leads to request waterfalls. Why wouldn't I just add the assets directly to the HTML?

graup:
* Yeah I'm also a bit concerned about local imports. That's a can of worms we likely don't want to open.
* If you link to the `woff2` files directly, I believe that has a couple disadvantages: I suspect Google is doing some magic to only deliver the file types needed for the user's browser. Also if they ever update the rules, you'll miss out on that.

The better workaround is importing the CSS using a `<link>` tag. That is fine, we just have to tell our consumers do that.

So yeah, this is not a must-have feature.

graup:
Since there hasn't been additional interest in this feature, I'm gonna close my PR!

SalahAdDin:
It would be a great feature man @graup !

Issue #665, opened 2022-04-22
Vite: is there a way to import files in `.css.ts` files?


dkzlv:
Currently importing an image fails with this error:

```
[plugin:vanilla-extract] Build failed with 1 error:
Check.css.ts:6:16: error: No loader is configured for ".png" files: pic.png
```

, even though it works fine in a regular `.ts` file.

wobsoriano:
`.css.ts` is working fine on my end with latest vite version. Try

```ts
export default defineConfig({
  assetsInclude: ['**/*.png'],
  ...
})
```

AndrewLeedham:
This is probably because [Vanilla Extract runs `.css.ts` files in esbuild](https://github.com/seek-oss/vanilla-extract/blob/master/packages/integration/src/compile.ts#L45) to generate the stylesheets, regardless of which plugin you use. There is currently no way to add loaders or configure esbuild here. Curious what your use case is?

AndrewLeedham:
@dkzlv Makes sense. I think having the ability to configure esbuild would be a good addition, it would also be a better way of doing this: https://github.com/seek-oss/vanilla-extract/pull/661

graup:
I believe this is also a problem with importing font files (.woff2) in your VE files. Maybe we can consider adding support for url imports (images, font files) to the default config.

willbattel:
Yeah I'm trying use static assets in my CSS, such as for a custom `<li>` bullet style, but we're facing the same limitation.

```
import { style } from "@vanilla-extract/css";

import myIcon from '../assets/icons/MyIcon.svg'

export const myListItem = style({
    listStyle: `url(${myIcon})`,
})
```

```
[plugin:vanilla-extract] Build failed with 1 error:
src/components/MyComponent.css.ts:6:25: error: No loader is configured for ".svg" files: src/assets/icons/MyIcon.svg
```

Is this something we can get working with some tweaking of esbuild?

johanneslumpe:
Is there something the community can help with here? It's not possible to use a file loader right now because it requires defining the `outDir` for esbuild, which isn't exposed in the current esbuild options.

wobsoriano:
For now, I am manually setting esbuild loader object

```ts
// vite.config.ts
import react from '@vitejs/plugin-react'
import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';

export default defineConfig({
  plugins: [
    vanillaExtractPlugin({
      esbuildOptions: {
[...]

PR #668, opened 2022-04-29
Add esbuild configuration

AndrewLeedham:
Adds an option for both the Vite and esbuild plugins to customize the internal esbuild process.

supersedes: #661


mattcompiles:
Hey @AndrewLeedham 👋 

I'm not against allowing customisation of the internal esbuild compile. However I'm not sure how this fixes #665? I'm assuming that issue refers mostly to the importing of assets like images. Which means just configuring loaders isn't enough to fix the problem. You'd also need to forward the assets to the parent build somehow. 

I've thought about this issue a lot and I've started to think using `url` imports in the CSS is a more ideal approach. That way your main bundler can use your main image loading pipeline when it interprets the CSS `url`, rather than having to setup one for internal VE builds as well. I'm hoping having merged #647 should allow better support for that scenario.

AndrewLeedham:
Hi @mattcompiles 👋 

I see your point. For #665 you would have to setup a separate pipeline to Vite, which is likely not ideal. I'll remove it from the description. It is not a use-case I particularly have just sparked the thought process for this PR which also fixes my `__filename` issue.

.changeset/clever-ghosts-roll.md
```diff
@@ -0,0 +1,8 @@
+---
+"@vanilla-extract/esbuild-plugin": patch
+"@vanilla-extract/integration": patch
+"@vanilla-extract/vite-plugin": patch
+"site": patch
```
mattcompiles, line 5: We don't need to version bump the site. You can remove this.

.changeset/clever-ghosts-roll.md
```diff
@@ -0,0 +1,8 @@
+---
+"@vanilla-extract/esbuild-plugin": patch
+"@vanilla-extract/integration": patch
+"@vanilla-extract/vite-plugin": patch
```
mattcompiles, line 4: These should all be `minor` as they add new API

README.md
```diff
@@ -393,6 +393,16 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
 
 Each integration will set a default value based on the configuration options passed to the bundler.
 
+### esbuildOptions (Vite)
+The Vite plugin uses esbuild internally to compile `.css.ts` files before running them to extract styles. You can pass additional options here to customize that process.
+
+Accepts an esbuild build options object: https://esbuild.github.io/api/#build-api which is deep merged with the default options.
+
+### internalEsbuildOptions (Vite)
```
mattcompiles, line 401: I can see why you made this one different but I feel like it might be easier to just name it `esbuildOptions` in all the plugins. Makes the docs much cleaner too.

README.md
```diff
@@ -393,6 +393,16 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
 
 Each integration will set a default value based on the configuration options passed to the bundler.
 
+### esbuildOptions (Vite)
```
mattcompiles, line 396: Maybe instead we could do something like:

```suggestion
### esbuildOptions

> Only for `esbuild`, `vite` and `rollup` plugins
```

packages/integration/package.json
```diff
@@ -20,7 +20,8 @@
     "find-up": "^5.0.0",
     "javascript-stringify": "^2.0.1",
     "lodash": "^4.17.21",
-    "outdent": "^0.8.0"
+    "outdent": "^0.8.0",
+    "ts-deepmerge": "^2.0.1"
```
mattcompiles, line 24: We already have a `deepmerge` dep in the core `@vanilla-extract/css` package. Could we just use the same one here?

packages/integration/src/compile.ts
```diff
@@ -46,7 +51,12 @@ export async function compile({
     write: false,
     plugins: [vanillaExtractFilescopePlugin()],
     absWorkingDir: cwd,
-  });
+  };
+
+
+  const resolvedOptions = merge(baseOptions, esbuildOptions ?? {});
```
mattcompiles, line 57: Could we only run the merge if there's anything to merge?

e.g. 
```suggestion
  const resolvedOptions = esbuildOptions ? merge(baseOptions, esbuildOptions) : baseOptions;
```

mattcompiles: Also just thinking about this, I'm wondering if exposing the whole set of options is a good idea. Outside of plugins, what can you add here that won't break vanilla-extract. Maybe we should just be exposing plugins? Thoughts?

mattcompiles (commented):
We'll want to add this to the `rollup-plugin` as well.

packages/integration/src/compile.ts
AndrewLeedham: Scoping is probably sensible. Perhaps these options would make sense:
`plugins`, `external`, `define`, `loader`?

packages/integration/package.json
AndrewLeedham: I ended up getting rid of the `deepmerge` altogether now there is a known set of options.

AndrewLeedham:
I have been experimenting with these changes in my project and have learnt a lot about how esbuild works. The main problem I am facing with rewriting `import.meta.url` or `__filename` is that esbuild does not chain plugins, so the `vanillaExtractFilescopePlugin` always runs first and loads the file meaning no subsequent load callbacks are run. The only way I can get it to work seems to be by replacing the `vanillaExtractFilescopePlugin` and adding my logic to it. Which does not seem ideal.

packages/integration/src/compile.ts
```diff
@@ -27,25 +27,29 @@ export const vanillaExtractFilescopePlugin = (): Plugin => ({
   },
 });
 
+
 interface CompileOptions {
   filePath: string;
   cwd?: string;
   externals?: Array<string>;
```
mattcompiles, line 34: This type needs to be removed as it's not longer honoured.

.changeset/clever-ghosts-roll.md
```diff
@@ -0,0 +1,7 @@
+---
+"@vanilla-extract/esbuild-plugin": minor
+"@vanilla-extract/integration": minor
```
mattcompiles, line 3: This is actually a major for the `integration` package as it contains a breaking change for how you pass externals.

README.md
```diff
@@ -393,6 +393,13 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
 
 Each integration will set a default value based on the configuration options passed to the bundler.
 
+### esbuildOptions
+> Only for `esbuild`, `vite` and `rollup` plugins
+
+The `vite`, `esbuild` and `rollup` plugins use esbuild internally to compile `.css.ts` files before running them to extract styles. You can pass additional options here to customize that process.
+
+Accepts an esbuild build options object: https://esbuild.github.io/api/#build-api which is deep merged with the default options.
```
mattcompiles, line 401: This isn't true anymore right? We only accept a select few of the options.

site/docs/setup.md
```diff
@@ -306,3 +306,10 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
 - `debug` identifiers contain human readable prefixes representing the owning filename and a potential rule level debug name. e.g. `myfile_mystyle_hnw5tz3`
 
 Each integration will set a default value based on the configuration options passed to the bundler.
+
+### esbuildOptions
+> Only for `esbuild`, `vite` and `rollup` plugins
+
+The `vite`, `esbuild` and `rollup` plugins use esbuild internally to compile `.css.ts` files before running them to extract styles. You can pass additional options here to customize that process.
+
+Accepts an esbuild build options object: https://esbuild.github.io/api/#build-api which is deep merged with the default options.
```
mattcompiles, line 315: Same here, needs to be updated to reflect the limited set of options we allow.

mattcompiles (commented):
Looks good mate, just a couple more comments. Sorry I've been slow to reply.

.changeset/clever-ghosts-roll.md
```diff
@@ -0,0 +1,7 @@
+---
+"@vanilla-extract/esbuild-plugin": minor
+"@vanilla-extract/integration": major
+"@vanilla-extract/vite-plugin": minor
```
mattcompiles, line 4:
```suggestion
"@vanilla-extract/vite-plugin": minor
"@vanilla-extract/rollup-plugin": minor
```

mattcompiles (approved):

PR #680, opened 2022-05-10
Feature/fillings

rafal2228:
As previously discussed with @mattcompiles @michaeltaranto and @okonet here's a PR for **Fillings**.

As mentioned in docs **Fillings** are meant to be "dynamic equivalent of **Sprinkles**, where user is free to specify any value to the properties" - thanks to the usage of CSS variables under the hood.


packages/fillings/package.json
```diff
@@ -0,0 +1,44 @@
...
+      "index.ts",
+      "createRuntimeFillings.ts"
+    ]
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/seek-oss/vanilla-extract.git",
+    "directory": "packages/fillings"
+  },
+  "author": "SEEK",
```
okonet, line 34: I think you should mention yourself as the author here.

okonet (commented):
Looking good! I will leave the proper review to maintainers.

rafal2228:
@graup I used to do something similar with css variables through a util:

```ts
export const createCSSVar = () => {
  const variable = createVar();
  const name = /var\((.*)\)/.exec(variable)?.[1] as string;

  return {
    variable,
    name,
  };
};
```

That way I was able to pass the `variable` part as a value to any `style`/`recipe` declaration and configure it later through `name`.  

```ts
export const color = createCSSVar();

export const menuRecipe = recipe({
  base: {
    color: color.variable,
  },
  variants: {
    size: {
      red: {
        [color.name]: vars.colors.red,
      },
      blue: {
        [color.name]: vars.colors.blue,
      },
      green: {
        [color.name]: vars.colors.green,
      },
    },
  },
});
```

But in theory, nothing would stop us from supporting `CSSVarFunction` types natively in fillings the way you described 👍

roginfarrer:
This looks neat! Exciting to see a first-party solution for dynamic Sprinkles. I'm curious if there was consideration to bake in support for predefined values or theme vars? I recently open-sourced [Rainbow Sprinkles](https://github.com/wayfair/rainbow-sprinkles), and it includes the option for providing vars. Is it possible to define properties in both sprinkles and fillings and use those functions together? (Some predefined utility classes from sprinkles, as well as accepting dynamic values from fillings)

rafal2228:
> This looks neat! Exciting to see a first-party solution for dynamic Sprinkles. I'm curious if there was consideration to bake in support for predefined values or theme vars? I recently open-sourced [Rainbow Sprinkles](https://github.com/wayfair/rainbow-sprinkles), and it includes the option for providing vars. Is it possible to define properties in both sprinkles and fillings and use those functions together? (Some predefined utility classes from sprinkles, as well as accepting dynamic values from fillings)

Interesting approach :) I wonder what use cases did you have that required specification of both pre-defined and dynamic values for css props?

> I'm curious if there was consideration to bake in support for predefined values or theme vars?

In general, the way we use it in our team (not VS/Seek) - when we need to pass one of predefined CSS variables/theme vars - to `fillings` based props is:

```tsx
import { vars } from 'styles-package';

// 'red.500' is one of the colors from theme
<Box color="red.500" maxWidth={vars.spacing.lg}>Content</Box>
```

_Note. `Box` is a component that utilizes both fillings and sprinkles for different props at once, where fe. `color` prop is sprinkle based and `maxWidth` is fillings based_

You essentially can pass one of the necessary theme vars directly to fillings based prop. That way we keep those APIs separate - sprinkles are really great for properties we want to keep as standard as possible (fe. color, padding, etc) - and only utilize `fillings` for those that are dynamic in their nature (maxWidth, gridTemplateAreas, etc)

okonet:
> as I was more focused on aligning with the styled-system API that uses string values. 

I think this is the most controversial decision of styled-system that leads to unpredictable results. 

One approach that worked for me previously was to separate styled props and custom (dynamic) CSS using a separate prop (`css`). This prop would allow passing any CSS object to the CSS-in-JS library of choice. I don't know if it's possible to do without the lib but as a last resort I can imagine `style` could work as this API is meant for one-off usages anyways. 

The approach with importing constants from theme seems like a better one to me since it's more explicit.

roginfarrer:
> I think this is the most controversial decision of styled-system that leads to unpredictable results.
> 
> One approach that worked for me previously was to separate styled props and custom (dynamic) CSS using a separate prop (css). This prop would allow passing any CSS object to the CSS-in-JS library of choice. I don't know if it's possible to do without the lib but as a last resort I can imagine style could work as this API is meant for one-off usages anyways.
> 
> The approach with importing constants from theme seems like a better one to me since it's more explicit.

Can you expound on what the unpredictable results are? I get how you could have name conflicts between named CSS colors and color variables, and that can be confusing for some users. Are you also thinking in terms of supporting non-standard values within a design system? (If so, I agree that this depends on the needs and guidelines of the design system and the needs of its users.) I also think it's interesting to think about moving custom values to a separate prop—this could be useful as a design system team to track custom values, and make it clear to consumers when they're deviating from the system.

@rafal2228 

> I would disagree here - realistically speaking you will end up using a small subset of all possible values - those could be handled with sprinkles. For the rest of snowflake occurances I would opt-in for either - native style API or creation of custom .css.ts files

> The difference in size doesn't really speak to me - it's a tradeoff of generating more classes vs passing values in javascript (which has to be done in both cases, but with classes you get less imports).

When it comes to size, I mean the overall number of bytes shipped in the CSS bundle vs the bytes shipped in the document for adding a CSS class and inline style for each property/value. 

From a design system team perspective (we built rainbow sprinkles for our design system), we were frequently getting questions about expanding the API for supported values, and were frequently running up against an ever-growing stylesheet. For example, we have design tokens for several dozen colors. A long while ago we had a Box component built with static CSS, and we were very conservative with which colors we supported in order to reduce the CSS bundle size. But then we consistently had to engage in conversations about adding support for new colors and weight the bundle size trade-offs for adding it. Being able to support all color tokens via a dynamic API could resolve this issue without having to change the API for consumers. They don't have to know whether a value is dynamic or not, the API remains the same (e.g., `color="blue200"` could be a static utility class or a dynamic value). 

> However I understand that you might prefer the API where a subset of theme is available to you via simple string and I believe this could be done on top of fillings. You can write a function (on the client side), that would inspect passed props and if value of any fillings based prop is a key of theme, you could simply pass the value from the theme.

This totally makes sense! And not something I had considered when building out rainbow sprinkles. I now see that it would make sense for an "official" library to be as simple as possible, and let this functionality be composed by the consumers that want it.

I appreciate both of your perspectives, and there's a lot here for me to consider!

packages/fillings/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/seek-oss/vanilla-extract.git",
+    "directory": "packages/fillings"
+  },
+  "author": "rafal2228",
+  "license": "MIT",
+  "peerDependencies": {
+    "@vanilla-extract/css": "^1.0.0",
+    "@vanilla-extract/dynamic": "2.0.2"
```
mattcompiles, line 38: I think `@vanilla-extract/dynamic` would be better off as a regular dependency.

packages/fillings/src/createRuntimeFillings.ts
```diff
@@ -0,0 +1,123 @@
...
+        classNames.push(definitions[prop].className);
+
+        setVars.push([definitions[prop].variable, props[prop] as string]);
+      }
+    }
+
+    return {
+      className: classNames.join(' '),
+      assignVars() {
+        return assignInlineVars(Object.fromEntries(setVars));
```
mattcompiles, line 119: I'd probably avoid `Object.fromEntries` as it was only supported in Safari 12+.

packages/fillings/src/createRuntimeFillings.ts
```diff
@@ -0,0 +1,123 @@
...
+  config: Config,
+): FillingsFn<Config>;
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function createRuntimeFillings(config: any): any {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  return (options: any) => {
+    if (typeof options !== 'object' || options === null) {
+      return {
+        className: '',
+        assignVars() {
```
mattcompiles, line 25: What's the reasoning for `assignVars` being a function? Seeing as it doesn't ever take any parameters I would've thought it'd be better represented as `style` seeing as it's effectively an inline style object.

rafal2228: Guess I wanted to mimic the API of `dynamic` library, but you're right we will be always calling this one at the end of the day. Will refactor it to a `style` object

packages/fillings/src/createRuntimeFillings.ts
rafal2228: If it will broaden supported browsers list for this package we'll update it :)

rafal2228:
> This is super awesome! Not only to see people creating libraries on top of vanilla-extract's core, but also seeing these great discussions on how best to approach CSS architecture in design systems.
> 
> Seeing @roginfarrer's [Rainbow Sprinkles](https://github.com/wayfair/rainbow-sprinkles) approaching this problem in a different way has got me thinking about how we handle external libraries in the ecosystem going forward. While I really want to highlight these libraries on the main vanilla site, I also don't want to cause confusion with users or play favourites with other libraries that are not promoted to the core repo.
> 
> I know we originally suggested moving fillings into the main repo, but after thinking about it more I feel like maybe the library would be better served living separately. That way you can iterate on the API as you see fit. I'm very happy to give feedback (I have added some on this PR) however I think this API is your vision and you're the best one to own it.
> 
> That still leaves the issue of promoting these external libraries though. I'm thinking we should add a community libraries section to the site to allow other libraries to showcase their APIs to vanilla-extract users.
> 
> @rafal2228 What are your thoughts? If you disagree that's fine, I'm happy to hear your thoughts on this. 😄

I think as long as others will be able to use `fillings` both approaches will be good :) Our main goal at the end of the day is to contribute back to the community that gave us such great tools like vanilla extract. That said - I totally understand there might be different approaches to creation of design systems and your reluctance to play favorites :) We'll continue to develop `fillings` as our internal project for now then - we're planning on open-sourcing a bit more things later this year - but for now we don't have a setup for package publishing in our repositories. 

It would be great to have fillings listed in 3rd party plugins list one day :)

rafal2228:
Closing in favor of releasing fillings as a standalone plugin for VS

PR #681, opened 2022-05-10
Fix rollup plugin tests and docs for preserve modules with adjusted asset names

graup:
- The package name is not needed anymore after https://github.com/seek-oss/vanilla-extract/pull/673
- The test actually wasn't testing what I intended (preserving modules with correct paths for both JS and CSS – `preserveModuelsRoot` is normally not required but required here because of the way the fixture is referenced)

This has no effect on any published package code, so I didn't add a changeset.

### DIFF
--- a/site/docs/setup.md
+++ b/site/docs/setup.md
@@ -262,12 +262,12 @@ This plugin works well with Rollup's `preserveModules`.
 
 Rollup by default places assets in an "assets" directory.
 You can configure [asset file names](https://rollupjs.org/guide/en/#outputassetfilenames)
-if you care about CSS assets being placed right next to source files.
+if you care about CSS assets being placed right next to the corresponding JS files.
 
 ```js
   preserveModules: true,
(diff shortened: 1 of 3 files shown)


site/docs/setup.md
```diff
@@ -262,12 +262,12 @@ This plugin works well with Rollup's `preserveModules`.
...
 Rollup by default places assets in an "assets" directory.
 You can configure [asset file names](https://rollupjs.org/guide/en/#outputassetfilenames)
-if you care about CSS assets being placed right next to source files.
+if you care about CSS assets being placed right next to the corresponding JS files.
 
 ```js
   preserveModules: true,
   assetFileNames({ name }) {
-    return name.replace(/^package-name\/src/, '');
+    return name?.replace(/^src\//, '') ?? '';
```
graup, line 270: not sure if TS or JS is better here

mattcompiles (approved):

PR #690, opened 2022-05-16
Add include & exclude options to Rollup plugin

roginfarrer:
It's common for Rollup plugins to have `include` and `exclude` options (so much so that they have official utils to create them). I'm working in a codebase that has `.css.js` files that predate our adoption of Vanilla Extract, and so we'd ideally be able to avoid processing them. 

This PR adds those options using the official Rollup utility, and makes the `cssFilterFile` function the default `include`. Also added tests.


graup:
Looks good to me!

pnpm-lock.yaml
```diff
@@ -585,6 +587,17 @@ packages:
...
+    resolution: {integrity: sha512-vKsoSQAyBmxS35JUOOt+07cLc6Nk/2ljLIHwmq2/NM6hdioUaqEXq/S+nXvbvXbZkNDlWOymPanJGOc4CBjSJA==}
+    engines: {node: '>=6.9.0'}
+    peerDependencies:
+      '@babel/core': ^7.0.0
+    dependencies:
+      '@babel/compat-data': 7.16.4
+      '@babel/helper-validator-option': 7.14.5
+      browserslist: 4.18.1
+      semver: 6.3.0
+
```
graup, line 600: There are a lot of changes in the lockfile that seem unnecessary, is that just due to the different lockfileVersion?

packages/rollup-plugin/test/rollup-plugin.test.ts
```diff
@@ -25,9 +25,6 @@ async function buildAndMatchSnapshot(outputOptions: OutputOptions) {
       chunkOrAsset.type === 'asset' ? chunkOrAsset.source : chunkOrAsset.code,
     ]),
   ).toMatchSnapshot();
-  if (bundle) {
-    await bundle.close();
-  }
```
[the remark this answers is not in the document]
roginfarrer: I actually didn't mean to remove it 😅 . I'm not sure what it's needed for, but I can remove the rest

pnpm-lock.yaml
roginfarrer: I think it's the result of installing `@rollup/pluginutils`

mattcompiles: This project hasn't been updated to pnpm 7 yet. That could be the issue?

roginfarrer: Oooh yup, that seemed to be it. Reinstalled with v6

graup:
@mattcompiles I'm good with this change. If you're happy with the lock file, lgtm!

packages/rollup-plugin/src/index.ts
```diff
@@ -8,24 +8,28 @@ import {
...
   const isProduction = process.env.NODE_ENV === 'production';
+  const filter = createFilter(include, exclude);
 
   return {
     name: 'vanilla-extract',
     async transform(_code, id) {
-      if (!cssFileFilter.test(id)) {
-        return null;
-      }
+      if (!filter(id)) return null;
```
mattcompiles, line 32: Just defaulting the `include` to `cssFileFilter` seems like a foot gun. It seems reasonable that a user would want to scope their rules to a folder for example. e.g. `include: ['vanilla-styles/**/*']`. But this would lead to plugin running on all files in that folder, not just the `.css.ts` files. I think the filter function should be combined with the regular `cssFileFilter` check as all files still need to have a `.css.ts/js` extension, that's not configurable.

.changeset/fuzzy-months-hear.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/rollup-plugin': minor
+'@fixtures/themed': patch
```
mattcompiles, line 3: No need to add this

```suggestion
```

mattcompiles (commented):
Looks good, just a comment about how the filter is being applied.

Issue #698, opened 2022-05-26
Vite dev doesn't get the correct style when refreshing


pakholeung37:
### Describe the bug
Vite dev doesn't get the correct style when refreshing. 

A source code like this
```javascript
// header.css.ts
export const header = style({
  display: 'flex'
})
```
generate code like this
```javascript
import { createHotContext as __vite__createHotContext } from "/@vite/client";import.meta.hot = __vite__createHotContext("/src/components/Header/header.css.ts.vanilla.js");import { injectStyles } from '/node_modules/.vite/deps/@vanilla-extract_css_injectStyles.js?v=bf7f4fdc';

const inject = (css) => injectStyles({
  fileScope: {"filePath":"src/components/Header/header.css.ts.vanilla.js"},
  css
});

inject(".header__16c76pf0 {\n  display: flex;");

import.meta.hot.on('vanilla-extract-style-update:src/components/Header/header.css.ts.vanilla.js', (css) => {
  inject(css);
});  
```
It seems that when changing the code, vite-plugin only runs `inject(css)` in `meta.hot.on` but does not change the file content. When refreshing the page, the original style will show up which was built in vite first startup.

```node
@vanilla-extract/vite-plugin@3.1.6

```

pakholeung37:
same as https://github.com/seek-oss/vanilla-extract/issues/578#issuecomment-1133020076

pakholeung37:
It turns out there is a bug in @vanillia-extract/vite-plugin 
https://github.com/seek-oss/vanilla-extract/blob/aebf9349371d631a38145df70c7de40643843554/packages/vite-plugin/src/index.ts#L166
The correct id should be `path.resolve('.', id)` (which is absolute path) not `id` (relative path like `src/components/...`). Not sure it is vite's bug or plugin's bug.

PR #707, opened 2022-06-01
vite: Absolute module id for invalidation

AndrewLeedham:
fixes: #698

@pakholeung37 I added you as a co-author on the commit since you essentially posted the fix in this issue 😃


mattcompiles (approved):
Thanks for the fix 👍 

Would love to get some tests for theses kinds of scenarios at some stage.

mattcompiles:
@AndrewLeedham Looks like you need to format the code. Run `pnpm format`.

PR #710, opened 2022-06-01
Normalize import path for emitted css files by rollup-plugin

### LINKED ISSUE #709: @vanilla-extract/rollup-plugin emit backslash in css imports on Windows system
### Describe the bug

On windows, while using @vanilla-extract/rollup-plugin, the relative imports added by vanilla extract have back slash on Windows.

This causes build errors because the paths cannot be resolved.

https://github.com/seek-oss/vanilla-extract/blob/ac8ad84310f4132029c32b2f46585b36a2662b8f/packages/rollup-plugin/src/index.ts#L85-L90

image

### Link to reproduction
image

https://github.com/riccardoperra/solid-vanilla-extract-repro/tree/rollup_plugin

### System Info
```node
  System:
    OS: Windows 10 10.0.25126
    CPU: (16) x64 AMD Ryzen 7 3700X 8-Core Processor
    Memory: 14.82 GB / 31.92 GB
  Binaries:
    Node: 16.13.2 - ~\AppData\Local\pnpm\node.EXE
    Yarn: 1.22.18 - ~\AppData\Local\pnpm\yarn.CMD
    npm: 8.1.2 - ~\AppData\Local\pnpm\npm.CMD
  Browsers:
    Edge: Spartan (44.25126.1000.0), Chromium (102.0.1245.30)
    Internet Explorer: 11.0.25126.1000
  npmPackages:
    @vanilla-extract/css: ^1.7.1 => 1.7.1
    @vanilla-extract/sprinkles: ^1.4.1 => 1.4.1
```

riccardoperra:
Windows systems default path separator is `backslash` which will not be replaced using `normalize`

Will fix #709


packages/rollup-plugin/src/index.ts
```diff
@@ -84,9 +84,8 @@ export function vanillaExtractPlugin({
       return importsToReplace.reduce((codeResult, importPath) => {
         const assetId = emittedFiles.get(importPath)!;
         const assetName = this.getFileName(assetId);
-        const fixedImportPath = `./${normalize(
-          relative(chunkPath, assetName),
-        )}`;
+        const normalizedImportPath = normalize(relative(chunkPath, assetName)).split(separator).join('/');
```
mattcompiles, line 87: Instead of trying to handle different OS file paths, I think we should just always use posix based paths? I assume Rollup is handling the OS normalization for us. 

You can force posix by instead importing the functions from `path.posix`. @graup thoughts on this?

packages/rollup-plugin/src/index.ts
```diff
@@ -7,7 +7,7 @@ import {
   getSourceFromVirtualCssFile,
   virtualCssFileFilter,
 } from '@vanilla-extract/integration';
-import { relative, normalize, dirname } from 'path';
+import { relative, normalize, dirname } from 'path/posix';
```
mattcompiles, line 10: @riccardoperra This seems to have broken the test suite as it's not importable for some reason. Maybe  try the following instead which we've done elsewhere in vanilla-extract.

```ts
import { posix } from 'path';

const { relative, normalize, dirname } = posix;
```

mattcompiles (approved):

PR #712, opened 2022-06-02
Add recipe variants to return value of `createRuntimeFn`

Mokshit06:
This PR adds the keys of variants passed to `createRuntimeFn` as a property on its return value. This allows for runtime reflection of variants that are available, which is useful in a few cases, like deciding which props should be passed to the `runtimeFn` and which to passed to the component

### DIFF
--- a/packages/recipes/src/createRuntimeFn.ts
+++ b/packages/recipes/src/createRuntimeFn.ts
@@ -19,11 +19,10 @@ const shouldApplyCompound = <Variants extends VariantGroups>(
   return true;
 };
 
-export const createRuntimeFn =
-  <Variants extends VariantGroups>(
-    config: PatternResult<Variants>,
-  ): RuntimeFn<Variants> =>
-  (options) => {
+export const createRuntimeFn = <Variants extends VariantGroups>(
+  config: PatternResult<Variants>,
+): RuntimeFn<Variants> => {
+  const runtimeFn: RuntimeFn<Variants> = (options) => {
     let className = config.defaultClassName;
 
     const selections: VariantSelection<Variants> = {
@@ -62,3 +61,8 @@ export const createRuntimeFn =
 
     return className;
   };
+
+  runtimeFn.variants = Object.keys(config.variantClassNames);
+
+  return runtimeFn;
+};
--- a/packages/recipes/src/types.ts
+++ b/packages/recipes/src/types.ts
@@ -32,9 +32,9 @@ export type PatternOptions<Variants extends VariantGroups> = {
   compoundVariants?: Array<CompoundVariant<Variants>>;
 };
 
-export type RuntimeFn<Variants extends VariantGroups> = (
+export type RuntimeFn<Variants extends VariantGroups> = ((
   options?: VariantSelection<Variants>,
-) => string;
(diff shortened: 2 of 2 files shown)


mattcompiles:
@Mokshit06 Could you provide some code examples of where you think this would be helpful?

Mokshit06:
@mattcompiles I was working on a css-in-ts library which provides a styled api wrapper over recipes API and allows to declare vanilla-extract styles in same file. In the styled API I would need some way to reliably figure out which props are to be passed through to the underlying component and which are variants, this API would make it easier

mattcompiles:
@Mokshit06 Sorry, it's still not super clear to me exactly what you're trying to achieve. Also, adding this data to all recipe calls is extra information that 99% of users don't require. Not sure I'm comfortable with adding it without having a clear use-case that's relevant to a fair share of users. 

If you're trying to build abstraction on top of recipes then you may be better off forking the code and internalising it into your own solution?

Mokshit06:
The abstraction already works without needing a fork, the thing that I want to add is:

```js
const styled = (Component, runtimeFn) => {
  return function StyledComponent(props) {
    const [variants, others] = splitProps(props, runtimeFn.variants)

    return <Component class={runtimeFn(variants)} {...others} />
  }
}

const button = recipe({ .. })
const StyledButton = styled("button", button)
```

here I want to be able to introspect the variants that a function returned by recipe has so that I can split the props accordingly. It is pretty small to add it to the core so I was looking if this could be added to the core without needing a fork of vanilla-extract/recipes

bjgrosse:
> The abstraction already works without needing a fork, the thing that I want to add is:
> 
> ```js
> const styled = (Component, runtimeFn) => {
>   return function StyledComponent(props) {
>     const [variants, others] = splitProps(props, runtimeFn.variants)
> 
>     return <Component class={runtimeFn(variants)} {...others} />
>   }
> }
> 
> const button = recipe({ .. })
> const StyledButton = styled("button", button)
> ```
> 
> here I want to be able to introspect the variants that a function returned by recipe has so that I can split the props accordingly. It is pretty small to add it to the core so I was looking if this could be added to the core without needing a fork of vanilla-extract/recipes

this is exactly my use case as well. Sprinkles nicely includes .properties and it seems like recipes doing the same would be very consistent.

packages/recipes/src/createRuntimeFn.ts
```diff
@@ -62,3 +61,8 @@ export const createRuntimeFn =
 
     return className;
   };
+
+  runtimeFn.variants = Object.keys(config.variantClassNames);
```
mattcompiles, line 65: How about we make `variants` a function instead? That way we only do this work when requested.

mszczepanczyk:
I need this too! I was actually about to create a PR but then I saw this one. Funnily I even implemented it exactly the same way byte for byte. @mattcompiles I can finish this PR if @Mokshit06 doesn't want to (make `variants` a function).

mattcompiles:
@mszczepanczyk Go for it bud 👍

mszczepanczyk:
@mattcompiles created a new PR #1044 as I've no right to push to this one

mattcompiles:
Closed in favour of #1044

PR #751, opened 2022-07-14
Support React Server Components in Vite plugin and fix HMR

frandiox:
Hi! We are trying to support `vanilla-extract` in [`@shopify/hydrogen`](https://github.com/shopify/hydrogen) but we've found that the Vite plugin in this repo is not compatible with React Server Components.

I'm adding here some minimal changes that enable RSC support with the [plugin we are making here](https://github.com/facebook/react/pull/22952). Forking your plugin is also possible, if for any reason you don't want to support RSC yet.

- The first commit is, to my understanding, a bug fix. Vite uses sometimes `?v=42`-like query strings, and I think the regex checked in https://github.com/seek-oss/vanilla-extract/commit/0bc2b9fd4bf96e9cfef87b656ed225f9dae2c815 should pass regardless.
- The second commit adds an environment variable that is truthy when we do an RSC build. The main problem is that Vite doesn't have a way to differentiate between pure SSR build vs RSC build. That's why for now we need to give it a manual hint. With this variable, we are able to collect styles that are imported in server components.
- The third commit should fix https://github.com/seek-oss/vanilla-extract/issues/735.

Alternatively, I think it's also possible to use a plugin option instead of an env variable but we'd need to add public interface and some extra code.

cc @benjaminsehl @juanpprieto

### DIFF
--- a/.changeset/soft-lobsters-promise.md
+++ b/.changeset/soft-lobsters-promise.md
@@ -2,4 +2,4 @@
 '@vanilla-extract/vite-plugin': patch
 ---
 
-Fix HMR and support React Server Components.
+Fix HMR
--- a/.changeset/soft-lobsters-promise.md
+++ b/.changeset/soft-lobsters-promise.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Fix HMR and support React Server Components.
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -68,7 +68,8 @@ export function vanillaExtractPlugin({
       virtualExt = `.vanilla.${config.command === 'serve' ? 'js' : 'css'}`;
     },
     resolveId(source) {
-      if (!source.endsWith(virtualExt)) {
+      const [validId, query] = source.split('?');
+      if (!validId.endsWith(virtualExt)) {
         return;
       }
 
@@ -76,21 +77,24 @@ export function vanillaExtractPlugin({
       // imported from outside the config root.
       const absoluteId = source.startsWith(config.root)
         ? source
-        : getAbsoluteVirtualFileId(source);
+        : getAbsoluteVirtualFileId(validId);
 
       // There should always be an entry in the `cssMap` here.
       // The only valid scenario for a missing one is if someone had written
       // a file in their app using the .vanilla.js/.vanilla.css extension
       if (cssMap.has(absoluteId)) {
-        return absoluteId;
+        // Keep the original query string for HMR.
+        return absoluteId + (query ? `?${query}` : '');
       }
     },
     load(id) {
-      if (!cssMap.has(id)) {
+      const [validId] = id.split('?');
+
+      if (!cssMap.has(validId)) {
         return;
       }
 
-      const css = cssMap.get(id);
+      const css = cssMap.get(validId);
 
       if (typeof css !== 'string') {
         return;
@@ -102,21 +106,23 @@ export function vanillaExtractPlugin({
 
       return outdent`
         import { injectStyles } from '@vanilla-extract/css/injectStyles';
-        
+
         const inject = (css) => injectStyles({
-          fileScope: ${JSON.stringify({ filePath: id })},
+          fileScope: ${JSON.stringify({ filePath: validId })},
           css
         });
 
         inject(${JSON.stringify(css)});
 
-        import.meta.hot.on('${styleUpdateEvent(id)}', (css) => {
+        import.meta.hot.on('${styleUpdateEvent(validId)}', (css) => {
           inject(css);
-        });   
+        });
       `;
     },
     async transform(code, id, ssrParam) {
-      if (!cssFileFilter.test(id.split('?')[0])) {
+      const [validId] = id.split('?');
+
+      if (!cssFileFilter.test(validId)) {
         return null;
       }
 
@@ -127,9 +133,6 @@ export function vanillaExtractPlugin({
       } else {
         ssr = ssrParam?.ssr;
       }
-
-      const index = id.indexOf('?');
-      const validId = index === -1 ? id : id.substring(0, index);
 
       if (ssr && !process.env.RSC_BUILD) {
         return addFileScope({
@@ -149,7 +152,7 @@ export function vanillaExtractPlugin({
       for (const file of watchFiles) {
         // In start mode, we need to prevent the file from rewatching itself.
         // If it's a `build --watch`, it needs to watch everything.
-        if (config.command === 'build' || file !== id) {
+        if (config.command === 'build' || file !== validId) {
           this.addWatchFile(file);
         }
       }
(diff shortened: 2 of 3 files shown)


juanpprieto:
Confirming that the updated plugin fixes HMR on this [starter](https://github.com/jonluca/vite-typescript-ssr-react)

benjaminsehl:
cc: @markdalgleish — could I get you to take a look at this? 🙏🙏🙏

packages/vite-plugin/src/index.ts
```diff
@@ -128,10 +134,7 @@ export function vanillaExtractPlugin({
         ssr = ssrParam?.ssr;
       }
 
-      const index = id.indexOf('?');
-      const validId = index === -1 ? id : id.substring(0, index);
-
-      if (ssr) {
+      if (ssr && !process.env.RSC_BUILD) {
```
markdalgleish, line 137: Rather than using an env var, I think this could use an `outputCss` option like the esbuild/webpack plugins, then this line would be `if (ssr || !outputCss) {`. The part I'm missing in this plugin is why it doesn't still go through `processVanillaFile` like the other plugins? cc @mattcompiles.

markdalgleish:
>the Vite plugin in this repo is not compatible with React Server Components.

Could you go into a bit more detail on this? How exactly does this plugin fit into a RSC setup? What is the current behaviour compared to your expectations?

>Alternatively, I think it's also possible to use a plugin option instead of an env variable but we'd need to add public interface and some extra code.

This would be my preference, and ideally it's an option describing the behaviour of the plugin rather than referencing RSC specifically. The answer to my previous question would help inform this.

frandiox:
@markdalgleish Thanks for checking this!

>Could you go into a bit more detail on this? How exactly does this plugin fit into an RSC setup? What is the current behaviour compared to your expectations?

To my understanding, in a traditional SSR setup you have client build + server/ssr build, and both resulting bundles are relatively similar because they include the same React components (one to do SSR, the other one to hydrate in the browser).

In RSC, however, these two builds are quite different: the client bundle only includes "client components", which come from files that are supposed to run in the browser. On the other hand, the server bundle also includes "server components" which are never leaked to the browser.

Our use-case is that we want to support importing vanilla-extract `*.css.ts` files in server components. Since these components are never discovered by the client build, their corresponding CSS is never generated. In the server build, the plugin logic assumes the CSS has already been generated by the client build and instead it goes to the `addFileScope` path.

This behavior works in traditional SSR builds. However, in RSC, Vite still enables the SSR flag but what we actually need is to generate the CSS instead of the JS version with `addFileScope`. Therefore, I'm trying to add here a way to signal the plugin to force emit CSS even on SSR builds.

>This would be my preference, and ideally it's an option describing the behaviour of the plugin rather than referencing RSC specifically. The answer to my previous question would help inform this.

I originally went with an environment variable because RSC is still experimental and things might change. This env var is added by the RSC Vite plugin to make other plugins aware of RSC builds.

However, we can instead add a public parameter `forceEmitCssInSsrBuild = false` or similar, and change the line to `if (ssr && (!forceEmitCssInSsrBuild || !config.build?.ssr)) {...}`. This way, we force generating CSS in SSR builds and keep the original behavior during dev and client builds.

markdalgleish:
Thanks for the detailed response!

Am I right in saying that your plugin also collects all styles generated across every build and then combines them into a single CSS file? If so, this makes me think that a public option on our Vite plugin like `forceEmitCssInSsrBuild` probably wouldn't be useful to anyone else because all it would do is generate duplicate styles. As a standalone setting on our plugin it wouldn't seem to make much sense without referencing your plugin anyway.

I feel like I've come around to the idea that this is handled behind the scenes so consumers can use our Vite plugins together and it just works, without having to set some weird flag in their config that nobody else in the world would ever want to use. If this approach takes off, it also means other plugins can silently hook into it to which would be nice. This is a potentially controversial call though, so I'll need to double check with the rest of the team before committing to this.

By the way, one thing I'm wondering is whether the environment variable from your plugin should be namespaced a bit more? e.g. `VITE_RSC_BUILD`?

frandiox:
@markdalgleish Thanks!

>Am I right in saying that your plugin also collects all styles generated across every build and then combines them into a single CSS file?

That's fairly accurate, yes. We currently ignore the client build CSS because it is a subset of what's generated in the RSC build (which includes both client and server components). Then, we emit a single CSS file asset when the RSC build is done.

>I feel like I've come around to the idea that this is handled behind the scenes so consumers can use our Vite plugins together and it just works, without having to set some weird flag in their config that nobody else in the world would ever want to use. If this approach takes off, it also means other plugins can silently hook into it to which would be nice.

Right, there's currently no easy way to allow communication across Vite plugins so the env variable seems to be the simplest solution. Plus, since things might still change, keeping it private and undocumented might be the right call.

> This is a potentially controversial call though, so I'll need to double check with the rest of the team before committing to this.

No problem, let us know!

>By the way, one thing I'm wondering is whether the environment variable from your plugin should be namespaced a bit more? e.g. VITE_RSC_BUILD?

Prefixing it with `VITE_` makes total sense, I'll update the RSC plugin and this PR 👍

packages/vite-plugin/src/index.ts
```diff
@@ -183,10 +186,16 @@ export function vanillaExtractPlugin({
             cssMap.get(absoluteId) !== source
           ) {
             const { moduleGraph } = server;
-            const module = moduleGraph.getModuleById(absoluteId);
+            const [module] = Array.from(
+              moduleGraph.getModulesByFile(absoluteId) || [],
```
mattcompiles, line 190: Not saying this is wrong, just for my own knowledge would like to understand this change more? 

Assuming the `absoluteId` doesn't match up with the vite id but does match a file?

.changeset/soft-lobsters-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Fix HMR and support React Server Components.
```
markdalgleish, line 5: Could you please split this into two changesets?

1) `patch` release: `Fix HMR`
2) `minor` release: `Add [experimental support for React Server Components](https://github.com/facebook/react/pull/22952)`

packages/vite-plugin/src/index.ts
frandiox: > Assuming the absoluteId doesn't match up with the vite id but does match a file?

That's right, according to my tests. A "module id" consists of a file path plus an optional query string that Vite adds for HMR or for other reasons. However, our `absoluteId` variable here is always created without a query string, so it sometimes doesn't match the module id. It always matches the file path, though.

.changeset/soft-lobsters-promise.md
frandiox: Done 👍 
If you want I can split the PRs as well (extract the env variable in a new PR).
Originally, I did it in one PR because at first I thought the HMR was only broken for RSC 😅

markdalgleish: As long as it's two changesets, I'm happy 😊

mattcompiles (approved):

juanpprieto:
Thanks team. Looking forward to putting together an hydrogen [example](https://github.com/Shopify/hydrogen/tree/v1.x-2022-07/examples/) very soon!
🙏🏼

Issue #746, opened 2022-07-07
Webpack-based HMR removes style-tag in monorepo setup


danielberndt:
### Describe the bug

Hi there!
After upgrading to create-react-app v5 which uses webpack 5.73.0, HMR is breaking after almost every save in our monorepo setup.

I finally was able to create a minimal repro-case here:
https://github.com/danielberndt/vanilla-extract-monorepo-issue

I'm attaching a recording of what is happening: When modifying a css file that is used in both its own package and another package within the monorepo, it's style tag gets removed:

![Screen Recording 2022-07-07 at 15 59 05](https://user-images.githubusercontent.com/856386/177793589-8f4d6b24-331c-4570-ad34-4c161dd36d2f.gif)

Interestingly it doesn't seem to happen if the contents of `frontend/src/App.js` get inlined into `frontend/src/index.js`

Let me know if you need any more details!

### Reproduction

https://github.com/danielberndt/vanilla-extract-monorepo-issue

### System Info

```shell
System:
    OS: macOS 11.6.5
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
    Memory: 277.28 MB / 20.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - /usr/local/bin/node
    Yarn: 1.22.10 - ~/.npm-global/bin/yarn
    npm: 8.1.0 - /usr/local/bin/npm
  Browsers:
    Brave Browser: 88.1.20.110
    Chrome: 103.0.5060.114
    Firefox: 101.0.1
    Safari: 15.5
```

### Used Package Manager

yarn

### Validations

- [X] Check that there isn't [already an issue](https://github.com/seek-oss/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

danielberndt:
Thanks for the pointer @kompot.
From my understanding, the config of the mono-repo should be sufficient. `create-react-app` already provides webpack with a rule for files ending in `.css`.
The provided [config](https://github.com/danielberndt/vanilla-extract-monorepo-issue/blob/main/frontend/config-overrides.js) works without issues in a non-mono-repo setup.

[...]

PR #755, opened 2022-07-15
[webpack-plugin] Allow `external` argument to accept both a string and array of strings

danielberndt:
To fix the issue described in #746, I need to specify externals of this form: 
```js
new VanillaExtractPlugin({externals: ["@name/pkg1/*", "@name/pkg1/**/*", "@name/pkg2/*", "@name/pkg2/**/*"]})
```
This PR allows `externals` to be either a single string or an array of strings.

I am aware that `externals` would also [accept](https://webpack.js.org/configuration/externals/) a regex or a function. But I haven't managed to make it work with either. I receive a `SyntaxError: Cannot use import statement outside a module` when using a regex, and the same error or
```
module.exports = @mono/ds/dsStyles.css;
                 ^
SyntaxError: Invalid or unexpected token
```
when using a function which returns different `type`s. So far only the array version of this PR worked reliably for me.

### DIFF
--- a/packages/webpack-plugin/src/compiler.ts
+++ b/packages/webpack-plugin/src/compiler.ts
@@ -114,7 +114,7 @@ function compileVanillaSource(
     new ExternalsPlugin('commonjs', [
       '@vanilla-extract/css',
       '@vanilla-extract/css/fileScope',
-      externals,
+      ...(Array.isArray(externals) ? externals : [externals]),
     ]).apply(childCompiler);
 
     let source: string;


danielberndt:
Feel free to close this PR as it's not fixing the underlying issue I've encountered. The PR might still be beneficial, so I'm not 100% sure whether to close it or not.

askoufis:
> Feel free to close this PR as it's not fixing the underlying issue I've encountered. The PR might still be beneficial, so I'm not 100% sure whether to close it or not.

Hey, thanks for the PR. I can't manage to reproduce this behaviour, the plugin seems to accept an array of externals just fine. Looking through the code for `ExternalsPlugin`, [it looks like it handles nested arrays](https://github.com/webpack/webpack/blob/b67626c7b4ffed8737d195b27c8cea1e68d58134/lib/ExternalModuleFactoryPlugin.js#L124-L150).

Going to close this PR, but feel free to open an issue if you run into this problem again.

PR #772, opened 2022-08-03
feat: provide a static set of conditions on the sprinkles function

astahmer:
Hey, first I want to thank you for this library and your work on it, it's great !

### Context:
This is both a feature request and a MVP as a PR.

Currently the sprinkles function only expose the `properties`, I'd like to be able to see (type-wise) / check (using as guards) the validity of the conditions. 
I was trying to implement [this feature request](https://github.com/TheMightyPenguin/dessert-box/issues/18) and realized I couldn't do it properly (no (easy?) way to infer all the conditions / I could check the validity of each props passed but couldn't check if the condition exist)

### Use-case
My use-case would be to implement properties grouped by conditions like:
`mobile: { color: "red", fontSize: "small" }`
rather than the opposite, having to specify conditions by property which is the current way of specifying conditions: 
`color: { mobile: "red" }, fontSize: { mobile: "red" }`

### Solution
For that I think a `conditions` property on the `createSprinkles` return is needed, just like the `properties` static attribute.
When I say "MVP", I mean that I guess my PR in the current state might not fit the repository, maybe you want some more tests or better typings, since I'm not really familiar with the repository I went for the quickest implementation 

Would love to hear your thought on this !

### DIFF
--- a/.changeset/cuddly-hounds-notice.md
+++ b/.changeset/cuddly-hounds-notice.md
@@ -2,4 +2,5 @@
 '@vanilla-extract/sprinkles': minor
 ---
 
-expose sprinkles conditions names
+Add conditions property to result of `createSprinkles`
+A `Set` of all the condition names assigned to the sprinkles function is now available via `sprinkles.conditions`.
--- a/.changeset/cuddly-hounds-notice.md
+++ b/.changeset/cuddly-hounds-notice.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/sprinkles': minor
+---
+
+expose sprinkles conditions names
--- a/packages/sprinkles/src/createSprinkles.ts
+++ b/packages/sprinkles/src/createSprinkles.ts
@@ -62,7 +62,10 @@ type SprinkleProps<Args extends ReadonlyArray<any>> = Args extends [
 
 export type SprinklesFn<Args extends ReadonlyArray<SprinklesProperties>> = ((
   props: SprinkleProps<Args>,
-) => string) & { properties: Set<keyof SprinkleProps<Args>> };
+) => string) & {
+  properties: Set<keyof SprinkleProps<Args>>;
+  conditions: Set<string>;
+};
 
 export const createSprinkles =
   <Args extends ReadonlyArray<SprinklesProperties>>(
@@ -70,6 +73,9 @@ export const createSprinkles =
   ) =>
   (...args: Args): SprinklesFn<Args> => {
     const sprinklesStyles = Object.assign({}, ...args.map((a) => a.styles));
+    const sprinklesConditions = new Set<any>(
+      args.flatMap((a) => a.conditions?.conditionNames).filter(Boolean),
+    );
     const sprinklesKeys = Object.keys(sprinklesStyles) as Array<
       keyof SprinkleProps<Args>
     >;
@@ -272,5 +278,6 @@ export const createSprinkles =
 
     return Object.assign(sprinklesFn, {
       properties: new Set(sprinklesKeys),
+      conditions: sprinklesConditions,
     });
   };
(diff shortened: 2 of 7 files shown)


packages/sprinkles/src/createSprinkles.ts
```diff
@@ -60,16 +60,35 @@ type SprinkleProps<Args extends ReadonlyArray<any>> = Args extends [
...
   : {};
 
+type SprinkleConditions<Args extends ReadonlyArray<any>> = Args extends [
+  infer L,
+  ...infer R
+]
+  ? L extends SprinklesProperties
+    ? [
+        keyof L['conditions'],
+        ...SprinkleConditions<[ChildSprinkleProps<L['styles']>]>,
```
mattcompiles, line 70: Not sure why we need to iterate over the `ChildSprinkleProps`? Think this line isn't required. However I could be wrong as I often forgot the underlying data structures in here.

packages/sprinkles/src/types.ts
```diff
@@ -1,3 +1,5 @@
+import type { Condition } from './';
```
mattcompiles, line 1: This is a circular import. Could we move the `Condition` definition in here and import it elsewhere?

packages/sprinkles/src/createSprinkles.ts
```diff
@@ -60,16 +60,35 @@ type SprinkleProps<Args extends ReadonlyArray<any>> = Args extends [
...
+      ]
+    : []
+  : [];
+
 export type SprinklesFn<Args extends ReadonlyArray<SprinklesProperties>> = ((
   props: SprinkleProps<Args>,
-) => string) & { properties: Set<keyof SprinkleProps<Args>> };
+) => string) & {
+  properties: Set<keyof SprinkleProps<Args>>;
+  conditions: Set<SprinkleConditions<Args>>;
```
mattcompiles, line 80: Could you please add a type test for this? You can see this existing ones in `tests/sprinkles/sprinkles-type-tests.ts`.

packages/sprinkles/src/createSprinkles.ts
astahmer: you are absolutely right there is no need for that, my bad

tests/sprinkles/sprinkles-type-tests.ts
```diff
@@ -283,4 +283,10 @@ const noop = (...args: Array<any>) => {};
   > = ['row'];
 
   noop(invalidRequiredValue);
+
+  conditionalProperties.conditions.conditionNames.includes('mobile');
+  conditionalProperties.conditions.conditionNames.includes('tablet');
+  conditionalProperties.conditions.conditionNames.includes('desktop');
+  // @ts-expect-error
+  conditionalProperties.conditions.conditionNames.includes('missingCondition');
```
mattcompiles, line 291: Could we change this to validate the public API (e.g. `sprinkles.conditions`) instead?

.changeset/cuddly-hounds-notice.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/sprinkles': minor
+---
+
+expose sprinkles conditions names
```
mattcompiles, line 5:
```suggestion
Add conditions property to result of `createSprinkles`

A `Set` of all the condition names assigned to the sprinkles function is now available via `sprinkles.conditions`.
```

mattcompiles (commented):
Just a couple of small things mate. Looking good 👍

astahmer:
(originally posted in discord but it will get lost quickly if not already)

I ended up also needing the shorthands map (to fix https://github.com/TheMightyPenguin/dessert-box/issues/20), should I submit another PR or do you want me to group it with the conditions PR ?

this would be another static property added to the sprinkles return, like so:
image

astahmer:
closing since it can be done in user-land as done [here](https://github.com/astahmer/box-extractor/blob/f0728ca0eb1b3332ce7218584497e9bb2a6f5cdf/packages/vanilla-extract/src/createBoxSprinklesInternal.ts)

PR #788, opened 2022-08-18
Add scale to the list of unitless properties

Dremora:
I've noticed the following CSS is being generated when I set `scale: 1`:

image

Scale [doesn't accept px](https://developer.mozilla.org/en-US/docs/Web/CSS/scale), so I think it should be added to the list of unitless properties.


mattcompiles (approved):
Thanks for taking the time to create the PR mate. Really appreciate it 👏 🍻

PR #793, opened 2022-08-22
Site redirects and bugs

michaeltaranto:
Some post-update site fixes:
- Add handling for legacy routes
- Hide `Show Output` button on mobile when no output to show
- Fix horizontal page scroll caused by code panels
- Make code panel max heights shorter on short screens

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ document.write(`
 
 ---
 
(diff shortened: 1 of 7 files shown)


mattcompiles (approved):
Nice mate 👏

PR #796, opened 2022-08-25
vite: Add experimental support for Astro

### LINKED ISSUE #795: Support for Astro
### Describe the bug

_This issue relates to this discussion: https://github.com/seek-oss/vanilla-extract/discussions/678_

Using Astro, which uses Vite under the hood, the styles are not applied.

The style is not at all imported, but the class names are correctly generated.
```html
<!-- from the reproduction example -->
<html>
   <!-- ... -->
  <body>
   <!-- ... -->
    <li class="Card__s70yhy1 astro-4Q53QFQW">
   <!-- ... -->
```

According to [this issue](https://github.com/withastro/astro/issues/2536#issuecomment-1122823724) (https://github.com/withastro/astro/issues/2536#issuecomment-1122823724):
> The vanilla-extract plugin works with Vite's default SPA templates, but doesn't work with Astro which renders purely server side. Not all Vite plugins work in Astro for this reason.

### Reproduction

https://stackblitz.com/edit/vanilla-extract-support-for-astro

### System Info

```shell
all
```

### Used Package Manager

npm

### Logs

_No response_

### Validations

- [X] Check that there isn't [already an issue](https://github.com/seek-oss/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.

---

### Edit (24/08/2022)

Using a React component with the `client:visible` directive works. On the screenshot, only the first card is a React component.

Capture d’écran 2022-08-24 à 21 10 34

mechairoi:
Generate CSS when in the Astro build and SSR flag is enabled, similar to the React Server Component (#751).

Will fix #795

[The reproduction of #795](https://stackblitz.com/edit/vanilla-extract-support-for-astro) seems to work correctly both of `dev` and `build`. (The build needs to move `pages/index.css.ts` outside of `pages/` so that it is not treated as an entry point.)

Screenshot 2022-08-25 22 49 18


arthur-fontaine:
I just updated and tested (with your pr) the example to make the `build` and `dev` scripts run without error and fix the style not fully applied on the cards.

packages/vite-plugin/src/index.ts
```diff
@@ -65,11 +69,15 @@ export function vanillaExtractPlugin({
         postCssConfig = await resolvePostcssConfig(config);
       }
 
-      virtualExt = `.vanilla.${config.command === 'serve' ? 'js' : 'css'}`;
+      virtualExt = config.command === 'serve' ? virtualExtJs : virtualExtCss;
```
mattcompiles, line 72: If `virtualExt` isn't the point of truth for what type of virtual file is loaded anymore then I'd prefer to remove it entirely and make the all the decisions inline.

packages/vite-plugin/src/index.ts
```diff
@@ -100,7 +108,11 @@ export function vanillaExtractPlugin({
         return;
       }
 
-      if (!server || server.config.isProduction) {
+      if (
+        !server ||
+        server.config.isProduction ||
+        validId.endsWith(virtualExtCss)
```
mattcompiles, line 114: I think the first two checks are pointless now. If we're loading a `.vanilla.css` file then it should always be a straight CSS return.

packages/vite-plugin/src/index.ts
```diff
@@ -165,7 +177,9 @@ export function vanillaExtractPlugin({
         identOption:
           identifiers ?? (config.mode === 'production' ? 'short' : 'debug'),
         serializeVirtualCssPath: async ({ fileScope, source }) => {
-          const rootRelativeId = `${fileScope.filePath}${virtualExt}`;
+          const rootRelativeId = `${fileScope.filePath}${
+            ssr && forceEmitCssInSsrBuild ? virtualExtCss : virtualExt
```
mattcompiles, line 181: Following on from before: 

`config.command === 'build' || (ssr && forceEmitCssInSsrBuild)` ? virtualExtCss : virtualExtJs`

packages/vite-plugin/src/index.ts
```diff
@@ -65,11 +69,15 @@ export function vanillaExtractPlugin({
         postCssConfig = await resolvePostcssConfig(config);
       }
 
-      virtualExt = `.vanilla.${config.command === 'serve' ? 'js' : 'css'}`;
+      virtualExt = config.command === 'serve' ? virtualExtJs : virtualExtCss;
+
+      if (config.plugins.find((p) => p.name === 'astro:build')) {
```
mattcompiles, line 74:
```suggestion
      if (config.plugins.some((p) => p.name === 'astro:build')) {
```

mattcompiles (commented):
Thanks for the PR. Changes makes sense. Just a few comments around keeping the plugin as simple as possible.

mechairoi:
Thanks for review. I have updated the PR.

mattcompiles (approved):
Happy with this as a code change. 

@mechairoi Is it worth adding some docs for Astro users? Is it understood that Vite plugins generally work in Astro builds?

mechairoi:
@mattcompiles I think it is better to add some docs. [The reference](https://docs.astro.build/en/reference/configuration-reference/#vite) is almost all the description of the Vite plugins in the Astro docs.

mattcompiles:
@mechairoi I'll merge this for now. If you want to contribute some Astro specific docs then feel free to open another PR 😄

PR #801, opened 2022-08-30
Fix Rollup/Vite generated exports

mrm007:
- https://rollupjs.org/guide/en/#outputgeneratedcode — see output.generatedCode.symbols
- https://github.com/vitejs/vite/pull/5018
- https://github.com/vitejs/vite/blob/757a92f1c7c4fa961ed963edd245df77382dfde6/packages/vite/src/node/build.ts#L466-L469


packages/integration/src/processVanillaFile.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    const value = {
+      string: `keep remove-me and-me dont-remove-me`,
+      other: `keep remove-me and-me`,
+    };
+
+    expect(
+      stringifyExports(recipeImports, value, /(remove-me|and-me)\s/g),
+    ).toMatchInlineSnapshot(
+      `"{string:'keep dont-remove-me',other:'keep and-me'}"`,
+      //                                            ^ this could be a problem
```
mrm007, line 51: This looks like a bug to me, although in practice it might not be a problem.

The regex was copied from here: https://github.com/seek-oss/vanilla-extract/blob/65b292d841b8843a38afa17bef5aabefc5dc8d2f/packages/integration/src/processVanillaFile.ts#L157

.prettierignore
```diff
@@ -6,6 +6,5 @@ CHANGELOG.md
 README.md
 .github/ISSUE_TEMPLATE
 site/docs-manifest.json
-.changeset
```
mrm007, line 9: I saw this was added [ages ago](https://github.com/seek-oss/vanilla-extract/pull/132), but no context other than "WIP".

packages/integration/src/processVanillaFile.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      other: `keep remove-me and-me`,
+      another: `keep remove-me dont-remove-me oh-no`,
+    };
+
+    expect(
+      stringifyExports(recipeImports, value, /(remove-me|and-me)\s/g),
+    ).toMatchInlineSnapshot(
+      `"{string:'keep dont-remove-me',other:'keep and-me',another:'keep dont-oh-no'}"`,
+      //                                            ^ these could be a problem ^
+    );
```
mrm007, line 53: These look like bugs to me, although in practice it might not be a problem.

The regex was copied from here: https://github.com/seek-oss/vanilla-extract/blob/65b292d841b8843a38afa17bef5aabefc5dc8d2f/packages/integration/src/processVanillaFile.ts#L157

.prettierignore
mattcompiles: I think we added it because it often prevents you from modifying them from the github UI, which we do frequently. Also there's not much point formatting them as they're removed and formatted via `Version Packages` anyway.

packages/integration/package.json
```diff
@@ -16,9 +16,9 @@
   "dependencies": {
     "@vanilla-extract/css": "^1.7.2",
     "esbuild": "^0.11.16",
-    "eval": "0.1.6",
+    "eval": "0.1.8",
```
mattcompiles, line 19: I remember having issues with bumping `eval`, hence why it's pinned. Not sure if those issues are resolved now but would prefer to change it in a separate PR.

packages/integration/package.json
```diff
@@ -16,9 +16,9 @@
...
     "find-up": "^5.0.0",
-    "javascript-stringify": "^2.0.1",
+    "javascript-stringify": "^2.1.0",
```
mattcompiles, line 21: Any reason for this bump? The old range still satisfies the latest version. Feel like this is unnecessary.

packages/integration/src/processVanillaFile.test.ts
mattcompiles: The assumption here is that composition classnames will never be the last one in the list. Could you update your test to reflect that?

packages/integration/package.json
mrm007: I thought the [latest release](https://github.com/blakeembrey/javascript-stringify/releases/tag/v2.1.0) was worth upgrading to. The old range matches the latest version, but it still needs a `pnpm update` to update the lockfile.

packages/integration/src/processVanillaFile.test.ts
```diff
@@ -0,0 +1,91 @@
...
+      styleWithComposition:
+        'features_styleWithComposition__1o6ek501 features_mergedStyle__1o6ek500',
+    };
+
+    const exports = stringifyExports(
+      recipeImports,
+      value,
+      /(features_compositionOnly__1o6ek504|features_styleCompositionInSelector__1o6ek507|features_styleVariantsCompositionInSelector_variant__1o6ek50a)\s/g,
+    );
+    const exportsObject = evalCode(`module.exports = ${exports}`, 'dummy.js');
```
mattcompiles, line 72: Why are we using eval here? Looks like we could just snapshot `exports` to the same effect?

mrm007: Yes, of course. I wanted to make it easier to read than

```js
    expect(exports).toMatchInlineSnapshot(
      `"{compositionOnly:'features_mergedStyle__1o6ek500 features_styleWithComposition__1o6ek501',mergedStyle:'features_mergedStyle__1o6ek500',styleCompositionInSelector:'features__1o6ek505 features__1o6ek506',styleVariantsCompositionInSelector:{variant:'features__1o6ek508 features__1o6ek509'},styleVariantsWithComposition:{variant:'features_styleVariantsWithComposition_variant__1o6ek502 features_mergedStyle__1o6ek500'},styleVariantsWithMappedComposition:{variant:'features_styleVariantsWithMappedComposition_variant__1o6ek503 features_mergedStyle__1o6ek500'},styleWithComposition:'features_styleWithComposition__1o6ek501 features_mergedStyle__1o6ek500'}"`,
    );
```

mattcompiles: Could we use JSOM.parse instead?

mrm007: Nope, not a JSON object (keys are not quoted).

.changeset/plenty-comics-compete.md
```diff
@@ -0,0 +1,11 @@
...
+'@vanilla-extract/integration': patch
+---
+
+Omit [`Symbol.toStringTag`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) when serializing module exports.
+
+Vite/Rollup sets this property to 'Module' which confuses `_.isPlainObject`
+
+- https://rollupjs.org/guide/en/#outputgeneratedcode (output.generatedCode.symbols)
+- https://github.com/vitejs/vite/pull/5018
+- https://github.com/vitejs/vite/blob/757a92f1c7c4fa961ed963edd245df77382dfde6/packages/vite/src/node/build.ts#L466-L469
```
mattcompiles, line 11: Let's remove this context from the changeset, changesets are user facing so this level of detail stands out a bit. 

```suggestion
```

mattcompiles (approved):
Looks good. Just a small tweak to the changeset.

PR #807, opened 2022-09-05
Add container query support

mattcompiles:
This PR adds support for container queries via the new `@container` key. It also adds a new top-level API called `createContainer` to allow for local scoping of container names. 

Example usage: 

```ts
import {
  style,
  createContainer
} from '@vanilla-extract/css';

const sidebar = createContainer();

const myStyle = style({
  containerName: sidebar,
  '@container': {
    [`${sidebar} (min-width: 768px)`]: {
      padding: 10
    }
  }
});
```


packages/css/src/types.ts
```diff
@@ -48,17 +49,42 @@ export interface FeatureQueries<StyleType> {
...
+    ContainerQueries<StyleType & FeatureQueries<StyleType>>
+> &
+  FeatureQueries<
+    StyleType &
+      MediaQueries<StyleType & ContainerQueries<StyleType>> &
+      ContainerQueries<StyleType & MediaQueries<StyleType>>
+  > &
+  ContainerQueries<
+    StyleType &
+      StyleType &
```
markdalgleish, line 70: This line is redundant.

packages/css/src/types.ts
```diff
@@ -1,15 +1,24 @@
 import type { MapLeafNodes, CSSVarFunction } from '@vanilla-extract/private';
-import type { PropertiesFallback, AtRule, Properties } from 'csstype';
+import type { Properties, AtRule } from 'csstype';
 
 import type { SimplePseudos } from './simplePseudos';
 
-type CSSTypeProperties = PropertiesFallback<number | (string & {})>;
+// csstype is yet to ship container property types as they are not in
+// the output MDN spec files yet. Remove this once that's done.
```
markdalgleish, line 7: It might be worth a link to the MDN docs here: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries

packages/css/src/types.ts
```diff
@@ -1,15 +1,24 @@
...
+interface ContainerProperties {
+  container?: string;
+  containerType?: 'size' | 'inline-size' | 'block-size' | (string & {});
+  containerName?: string;
+}
+
+type CSSTypeProperties = Properties<number | (string & {})> &
```
markdalgleish, line 14: Why is `PropertiesFallback` no longer used here?

site/docs/api/create-container.md
```diff
@@ -0,0 +1,43 @@
+---
+title: createContainer
+parent: api
+---
+
+# createContainer
+
+Creates a single scoped container for use with [CSS Container Queries]. This avoids potential naming collisions with other containers.
```
markdalgleish, line 8: Nit: Should it say "single scoped container _name_"? Would it be fair to say this function by itself doesn't create a container?

site/docs/api/create-container.md
```diff
@@ -0,0 +1,43 @@
...
+
+> 🚧 Ensure your target browsers [support container queries].
```
markdalgleish, line 10: Nice 👍

site/docs/overview/styling.md
```diff
@@ -237,6 +237,46 @@ globalStyle(`${parent} a[href]`, {
 });
 ```
 
+## Container Queries
+
+Container queries work the same as [Media queries] and are nested inside the `@container` key.
```
markdalgleish, line 242: Nit:
```suggestion
Container queries work the same as [media queries] and are nested inside the `@container` key.
```

packages/css/src/container.ts
```diff
@@ -0,0 +1,6 @@
+import { generateIdentifier } from './identifier';
+
+// createContainer is used for local scoping of CSS containers
+// For now it is mostly just an alias of generateIdentifier
```
markdalgleish, line 4: Nit: Why "for now"? Won't this always be the case?

site/docs/api/create-container.md
markdalgleish: This is also in the changeset if you think it's worth changing.

packages/css/src/types.ts
```diff
@@ -48,17 +49,42 @@ export interface FeatureQueries<StyleType> {
   };
 }
 
+export interface ContainerQueries<StyleType> {
+  '@container'?: {
+    [query: string]: StyleType;
+  };
+}
+
+export type WithQueries<StyleType> = MediaQueries<
```
markdalgleish, line 58: This is nice 👍

markdalgleish:
Should the blurb on the docs home page be updated? Here's a suggested change:

```diff
-Write maintainable CSS at scale without sacrificing platform features. Variables, selectors, pseudo‑classes, media/feature queries, keyframes, font‑face fules and global styles are all supported.
+Write maintainable CSS at scale without sacrificing platform features. Variables, selectors, pseudo‑classes, media/feature/container queries, keyframes, font‑face and global styles are all supported.
```

Note I changed `font-face rules` to `font-face` so it stays on 3 lines.

packages/css/src/types.ts
```diff
@@ -1,15 +1,24 @@
...
+import type { Properties, AtRule } from 'csstype';
 
 import type { SimplePseudos } from './simplePseudos';
 
-type CSSTypeProperties = PropertiesFallback<number | (string & {})>;
+// csstype is yet to ship container property types as they are not in
+// the output MDN spec files yet. Remove this once that's done.
+interface ContainerProperties {
+  container?: string;
+  containerType?: 'size' | 'inline-size' | 'block-size' | (string & {});
```
markdalgleish, line 10: I can't see `block-size` in the spec (https://www.w3.org/TR/css-contain-3/#container-type) and if I try using it in Chrome, the dev tools marks it as an error.

packages/css/src/container.ts
mattcompiles: I guess I wasn't sure if there would ever be more functionality. Potentially the container could provide an API for querying itself.

For example:

```ts
const container = createContainer();

style({
  '@container': {
    [container('(min-width: 400px)')]: ...
  }
})

Comment (markdalgleish) [reply packages/css/src/container.ts:4]: Oh, that's interesting.

File: packages/css/src/types.ts (hunk 9 above)
Comment (mattcompiles) [reply packages/css/src/types.ts:10]: Yeah I got it from MDN. I might remove it if it isn't in the spec. You can still override with string values, this is just for autocomplete anyway.

File: packages/css/src/types.ts (hunk 3 above)
Comment (mattcompiles) [reply packages/css/src/types.ts:14]: I was having type issues, but then realised we'd already created out own fallback types to better support our CSS variable type so it was redundant to use the built in one. 

Have a look at the CSSProperties type just below.

Comment (markdalgleish) [reply packages/css/src/types.ts:14]: Ah yeah, I didn't catch that. Nice one.

**Review by markdalgleish — APPROVED** [review]
Awesome work 🎉

File: packages/css/src/types.ts (hunk 3 above)
Comment (michaeltaranto) [reply packages/css/src/types.ts:15]: Is it worth adding a dummy type here with `@ts-expect-error` referencing `container` on `csstype`?

That way when we its available and we bump `csstype` we'll be notified to remove our version.

File: packages/css/src/transformCss.ts [hunk 10]
```diff
@@ -350,15 +354,49 @@ class Stylesheet {
...
+    if (rules) {
+      this.currConditionalRuleset?.addConditionPrecedence(
+        parentConditions,
+        Object.keys(rules).map((query) => `@container ${query}`),
+      );
+
+      forEach(rules, (containerRule, query) => {
+        const containerQuery = `@container ${query}`;
+
+        const conditions = [...parentConditions, containerQuery];
```
michaeltaranto, line 376: Minor: Noticed that `containerQuery` is not used, as a hangover from `transformMedia`. But given this is now a carbon copy of `transformSupports` should we look at sharing code here? Or are we deliberately preferring verbosity for readability?

site/docs/api/create-container.md
```diff
@@ -0,0 +1,43 @@
...
+import { sidebarContainer } from './sidebar.css.ts';
+
+export const navigation = style({
+  '@container': {
+    [`${sidebarContainer} (min-width: 400px)`]: {
+      display: 'flex'
+    }
+  }
+});
+```
```
michaeltaranto, line 40: 😘👌

site/docs/overview/styling.md
```diff
@@ -237,6 +237,46 @@ globalStyle(`${parent} a[href]`, {
 });
 ```
 
+## Container Queries
+
+Container queries work the same as [media queries] and are nested inside the `@container` key.
+
+> 🚧 Ensure your target browsers [support container queries]. Vanilla-extract supports the [container query syntax] but does not polyfill the feature in unsupported browsers.
```
michaeltaranto, line 244:
```suggestion
> 🚧&nbsp;&nbsp;Ensure your target browsers [support container queries]. Vanilla-extract supports the [container query syntax] but does not polyfill the feature in unsupported browsers.
```

(Blockquotes in the docs have been using double spacing after emojis)

site/docs/api/create-container.md
```diff
@@ -0,0 +1,43 @@
+---
+title: createContainer
+parent: api
+---
+
+# createContainer
+
+Creates a single scoped container name for use with [CSS Container Queries]. This avoids potential naming collisions with other containers.
+
+> 🚧 Ensure your target browsers [support container queries].
```
michaeltaranto, line 10:
```suggestion
> 🚧&nbsp;&nbsp;Ensure your target browsers [support container queries].
```

packages/css/src/transformCss.ts
mattcompiles: I think I like these being separate even though they are basically the same. Not sure there's much to gain by merging them outside of less lines of code.

michaeltaranto (approved):
Awesome stuff 🚀

PR #809, opened 2022-09-06
[Fix Document] Fix wrong props name

nkgrnkgr:
The name of the Container argument in the documentation seems to be wrong, so it has been corrected.


mattcompiles (approved):
Thanks for the PR 🍻

PR #810, opened 2022-09-06
Use standard identifier generation logic for variable names

markdalgleish:
I was getting errors during development when my variable names contained spaces. Turns out it's because we weren't handling variable identifiers the same as everything else, instead manually reimplementing the `generateIdentifier` logic but forgetting to replace spaces with underscores.

To fix this, I've refactored `createVar` to use `generateIdentifier`, but since var debug names previously didn't have the file name prefix applied, I've also added an additional option to `generateIdentifier` to support turning this off, which `createVar` then uses. This is public API so I've added a changeset for it.

NOTE: This PR now includes a bonus fix, ensuring debug file names are present for `.cjs` and `.mjs` files.

### DIFF
--- a/packages/css/src/identifier.test.ts
+++ b/packages/css/src/identifier.test.ts
@@ -3,26 +3,58 @@ import { generateIdentifier } from './identifier';
 
 describe('identifier', () => {
   beforeAll(() => {
-    setFileScope('test');
+    setFileScope('test.css.ts');
   });
 
   afterAll(() => {
     endFileScope();
   });
 
-  it(`should create a valid identifier`, () => {
+  it(`should ignore file scopes without a file extension when creating a path prefix`, () => {
+    setFileScope('test');
     expect(generateIdentifier(undefined)).toMatchInlineSnapshot(`"skkcyc0"`);
+    endFileScope();
+  });
+
+  it(`should create a valid identifier`, () => {
+    expect(generateIdentifier(undefined)).toMatchInlineSnapshot(
+      `"test__q11bah0"`,
+    );
   });
 
   it('should create a valid identifier with a debug id', () => {
     expect(generateIdentifier('debug')).toMatchInlineSnapshot(
-      `"debug__skkcyc1"`,
+      `"test_debug__q11bah1"`,
     );
   });
 
   it('should create a valid identifier with a debug id with whitespace', () => {
     expect(generateIdentifier('debug and more')).toMatchInlineSnapshot(
-      `"debug_and_more__skkcyc2"`,
+      `"test_debug_and_more__q11bah2"`,
     );
+  });
+
+  describe('options object', () => {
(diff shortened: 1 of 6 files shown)


packages/css/src/identifier.test.ts
```diff
@@ -3,26 +3,58 @@ import { generateIdentifier } from './identifier';
...
   });
 
   afterAll(() => {
     endFileScope();
   });
 
-  it(`should create a valid identifier`, () => {
+  it(`should ignore file scopes without a file extension when creating a path prefix`, () => {
+    setFileScope('test');
     expect(generateIdentifier(undefined)).toMatchInlineSnapshot(`"skkcyc0"`);
```
markdalgleish, line 15: This test is mostly here just to show that the logic is the same as before, we just couldn't see the usual output from this function because the file scope was invalid (`test` instead of `test.css.ts` or something similar).

packages/css/src/identifier.ts
```diff
@@ -3,23 +3,46 @@ import hash from '@emotion/hash';
...
-  const { filePath } = getFileScope();
 
-  const matches = filePath.match(
-    /(?<dir>[^\/\\]*)?[\/\\]?(?<file>[^\/\\]*)\.css\.(ts|js|tsx|jsx)$/,
-  );
+  if (debugFileName) {
+    const { filePath } = getFileScope();
+
+    const matches = filePath.match(
+      /(?<dir>[^\/\\]*)?[\/\\]?(?<file>[^\/\\]*)\.css\.(ts|js|tsx|jsx)$/,
```
mattcompiles, line 19: We hit and issues with this the other day where it didn't match on `.cjs` and `.mjs`. Could we add those while you're here?

packages/css/src/identifier.ts
```diff
@@ -3,23 +3,46 @@ import hash from '@emotion/hash';
...
 }
 
-export function generateIdentifier(debugId: string | undefined) {
+interface GenerateIdentifierOptions {
+  debugId?: string;
+  debugFileName?: boolean;
+}
+
+export function generateIdentifier(debugId?: string): string;
+export function generateIdentifier(options?: GenerateIdentifierOptions): string;
```
mattcompiles, line 37: I think you don't really need the overloads here as it always the same number of params. It's not wrong though. Upto you.

markdalgleish: True, but it's partly due to naming when consumers inspect the type, otherwise the type of `options` is also `string` which doesn't tell you it's a debug ID.

mattcompiles: Good point, happy to leave it then

packages/css/src/identifier.ts
markdalgleish: Added.

mattcompiles (approved):
Nice one 👏

PR #812, opened 2022-09-06
Sprinkles: Add support for container query conditions

markdalgleish:
This allows you to create utility classes scoped to container queries—pretty neat!

**Example usage**

```ts
import {
  createContainer,
  style
} from '@vanilla-extract/css';
import { defineProperties } from '@vanilla-extract/sprinkles';

const containerName = createContainer();

export const container = style({
  containerName,
  containerType: 'size'
});

const containerProperties = defineProperties({
  conditions: {
    smallContainer: {},
    mediumContainer: {
      '@container': `${containerName} (min-width: 768px)`
    },
    largeContainer: {
      '@container': `${containerName} (min-width: 1024px)`
    }
  },
  defaultCondition: 'smallContainer'
  // etc.
});
```

You can then consume them like this:

```ts
<div
  className={sprinkles({
    display: "flex",
    flexDirection: {
      smallContainer: "column",
      mediumContainer: "row",
    },
  })}
>
  ...
</div>
```

### DIFF
--- a/site/docs/packages/sprinkles.md
+++ b/site/docs/packages/sprinkles.md
@@ -772,3 +772,4 @@ const e: ResponsiveAlign = { desktop: 'center' };
 [tailwind]: https://tailwindcss.com
 [styled system]: https://styled-system.com
 [support container queries]: https://caniuse.com/css-container-queries
+[container query syntax]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries
--- a/site/docs/packages/sprinkles.md
+++ b/site/docs/packages/sprinkles.md
@@ -383,7 +383,9 @@ const responsiveProperties = defineProperties({
 
 ### conditions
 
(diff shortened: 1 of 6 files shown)


site/docs/packages/sprinkles.md
```diff
@@ -417,6 +419,40 @@ const properties = defineProperties({
 });
 ```
 
+Properties can also be scoped to container queries.
+
+> 🚧&nbsp;&nbsp;Ensure your target browsers [support container queries]. Vanilla-extract supports the [container query syntax] but does not polyfill the feature in unsupported browsers.
```
mattcompiles, line 424: Looks like you missed the markdown reference for this one

markdalgleish: Ah, nice catch.

mattcompiles (approved):
Nice 🍻

PR #836, opened 2022-09-19
Install plugins as dev dependencies in README

jgresham:
they're dev dependencies. also, some setups will try to bundle every normal dependency.


mattcompiles (commented):
I'm fine with this change but the README docs are about to be removed. The real docs are in `site/docs`.

markdalgleish (approved):

PR #837, opened 2022-09-20
Improve performance of transformCss with large amounts of class names

### LINKED ISSUE #771: NextJS compile especially slow on M1 Mac silicon
### Workarounds to try in your codebase!
It looks like this ticket isn't going to be addressed anytime soon, so I'll just share what we did in our codebase to massively mitigate this. Reduced compile time by 25x.
1. Move as many props as possible to unconditional `sprinkles` definitions (i.e. `defineProperties()` without `conditions` prop).
2. Patch VE to remove all the dev string concatenation. This is our patch using yarn. We left in the option to use the strings if an env is defined:
```diff
diff --git a/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.cjs.dev.js b/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.cjs.dev.js
index 6e40061..10283a2 100644
--- a/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.cjs.dev.js
+++ b/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.cjs.dev.js
@@ -177,11 +177,13 @@ function generateIdentifier(debugId) {
   var fileScopeHash = hash__default["default"](packageName ? "".concat(packageName).concat(filePath) : filePath);
   var identifier = "".concat(fileScopeHash).concat(refCount);
 
-  if (adapter_dist_vanillaExtractCssAdapter.getIdentOption() === 'debug') {
-    var devPrefix = getDevPrefix(debugId);
+  if (process.env.VANILLA_EXTRACT_DEV_PREFIX) {
+    if (adapter_dist_vanillaExtractCssAdapter.getIdentOption() === 'debug') {
+      var devPrefix = getDevPrefix(debugId);
 
-    if (devPrefix) {
-      identifier = "".concat(devPrefix, "__").concat(identifier);
+      if (devPrefix) {
+        identifier = "".concat(devPrefix, "__").concat(identifier);
+      }
     }
   }
```

### Describe the bug

We have a large NextJS codebase using VE that has progressively moved more and more of our CSS to VE over the past 2 months.

We've seen NextJS compile times on M1s get much slower as we adopted VE more (30s to 60s to 120s to 400s). Intel Macs and Windows do not have the issue - compiles are 30s for them.

We looked at 2 compile times, both of which got slower:
1. Starting up a next server and waiting for it to start serving pages
4. Going to a page that imports our VE components package and waiting until Next finishes the compile

FYI We have 74 `css.ts` files and several hundreds of elements with custom styles. And we use the standard `withVanillaExtract` plugin with no options.

Have you seen anything like this? Any tips to make this better for M1?

#### Workarounds attempted
In our NextJS page, removing the import to VE components eliminates the slowdown. It's around 3s after that. But of course we need the imports at some point.

I tried stubbing out all the `style()` and `recipe()` and other calls in our code to return empty strings. That brought the compile down to 150s. That is still much slower than the Intel compile.

After the stubbing, I removed the `withVanillaExtract` plugin and that brought the compile time down to 30s as expected. So I suspect it's something to do with the webpack stuff.

[See comments for more details on the following workarounds]

Installed an x64 version of node using Rosetta, and that brought compiles down to 50s.

Then I patched the VE webpack plugin to cache the result of processVanillaFile(), and that brought things down to 30s as expected.

### Reproduction

Can't provide since it's private code. The problem only seems to happen on non-minimal projects!

### System Info

```shell
System:
    OS: macOS 12.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 78.61 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.10.0 - ~/.nvm/versions/node/v16.10.0/bin/node
    Yarn: 3.2.0 - ~/.nvm/versions/node/v16.10.0/bin/yarn
    npm: 7.24.0 - ~/.nvm/versions/node/v16.10.0/bin/npm
  Browsers:
    Chrome: 103.0.5060.134
    Firefox: 102.0.1
    Safari: 15.6
  npmPackages:
    esbuild: 0.14.11 => 0.14.11 
    webpack: 5.73.0 => 5.73.0
```

### Used Package Manager

yarn

### Logs

_No response_

### Validations

- [X] Check that there isn't [already an issue](https://github.com/seek-oss/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.

mattcompiles:
When creating `.css.ts` files with large amounts of locally scoped classes, commonly encountered when using sprinkles, CSS output could be slow to create due to how selectors are transformed with large dynamic regular expressions. This issue was especially prevalent on M1 Macs. 

To resolve this, the old regular expression approach has been removed in favour of the following:
- In most cases, the selector being processed is just a single classname reference. In this case we don't need to search for all known local classnames and can be transformed whole after a single Map lookup. 
- In more complicated selectors, the RegExp approach has been removed in favour of using an Aho-Corasick based search, which is ideal for searching for multiple different keywords in text. 
  -  There aren't many libraries that implement this on NPM. `ahocorasick` seemed to be the best fit from a license, downloads and deps perspective. 

This change brings the [rainbowkit](https://github.com/rainbow-me/rainbowkit) start-up time down from ~180 seconds to ~12 seconds. 

Further optimisations can be done in future around caching CSS output more intelligently. However this will be explored in future PRs.  

Resolves #771

### DIFF
--- a/packages/css/src/ahocorasick.d.ts
+++ b/packages/css/src/ahocorasick.d.ts
@@ -1,0 +1,7 @@
+declare module 'ahocorasick' {
+  export default class Ahocorasick {
+    constructor(searchTerms: Array<string>);
+
+    search(input: string): Array<[endIndex: number, matches: Array<string>]>;
+  }
+}
--- a/packages/css/src/transformCss.test.ts
+++ b/packages/css/src/transformCss.test.ts
@@ -1564,4 +1564,30 @@ describe('transformCss', () => {
   });
 });
 
+it('should handle multiple references to the same locally scoped selector', () => {
+  expect(
+    transformCss({
+      composedClassLists: [],
+      localClassNames: ['testClass', 'parentClass'],
+      cssObjs: [
+        {
+          type: 'local',
+          selector: 'testClass',
+          rule: {
+            selectors: {
+              'parentClass &:before, parentClass &:after': {
+                background: 'black',
+              },
+            },
+          },
+        },
+      ],
+    }).join('\n'),
+  ).toMatchInlineSnapshot(`
+    ".parentClass .testClass:before, .parentClass .testClass:after {
+      background: black;
+    }"
+  `);
+});
+
 endFileScope();
--- a/packages/css/src/transformCss.test.ts
+++ b/packages/css/src/transformCss.test.ts
@@ -1571,7 +1571,7 @@ it('should handle multiple references to the same locally scoped selector', () =
   expect(
     transformCss({
       composedClassLists: [],
-      localClassNames: [style1, style2, '_1g1ptzo10', '_1g1ptzo1'],
+      localClassNames: [style1, style2, '_1g1ptzo1', '_1g1ptzo10'],
       cssObjs: [
         {
           type: 'local',
@@ -1585,6 +1585,10 @@ it('should handle multiple references to the same locally scoped selector', () =
               [`_1g1ptzo1_1g1ptzo10 ${style1}`]: {
                 background: 'blue',
               },
+
+              [`_1g1ptzo10_1g1ptzo1 ${style1}`]: {
+                background: 'blue',
+              },
             },
           },
         },
@@ -1595,6 +1599,9 @@ it('should handle multiple references to the same locally scoped selector', () =
       background: black;
     }
     ._1g1ptzo1._1g1ptzo10 .skkcyc1 {
+      background: blue;
+    }
+    ._1g1ptzo10._1g1ptzo1 .skkcyc1 {
       background: blue;
     }"
   `);
--- a/packages/css/src/transformCss.test.ts
+++ b/packages/css/src/transformCss.test.ts
@@ -1,10 +1,13 @@
 import { setFileScope, endFileScope } from './fileScope';
 import { createVar } from './vars';
 import { transformCss } from './transformCss';
+import { style } from './style';
 
 setFileScope('test');
 
 const testVar = createVar();
+const style1 = style({});
+const style2 = style({});
 
 describe('transformCss', () => {
   it('should escape class names', () => {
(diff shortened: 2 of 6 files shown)


markdalgleish (approved):
Nice one, this is massive.

packages/css/src/transformCss.ts
```diff
@@ -263,18 +278,43 @@ class Stylesheet {
...
+
+      // If class names already starts with a '.' then skip
+      if (transformedSelector[startIndex - 1] !== '.') {
+        transformedSelector = replaceBetweenIndexes(
+          transformedSelector,
+          startIndex,
+          endIndex + 1,
+          `.${cssesc(firstMatch, {
+            isIdentifier: true,
+          })}`,
```
mrm007, line 312:
```suggestion
          `.${cssesc(firstMatch, { isIdentifier: true, })}`,
```

Also, since this logic is now in [two places](https://github.com/seek-oss/vanilla-extract/blob/transform-perf/packages/css/src/transformCss.ts#L282) do you think it's worth creating an `escapeIdentifier` function to make sure they don't drift apart?

mrm007 (approved):

packages/css/src/transformCss.ts
```diff
@@ -111,10 +126,10 @@ class Stylesheet {
     this.conditionalRulesets = [new ConditionalRuleset()];
     this.fontFaceRules = [];
     this.keyframesRules = [];
-    this.localClassNameRegex =
-      localClassNames.length > 0
-        ? RegExp(`(${localClassNames.map(escapeStringRegexp).join('|')})`, 'g')
-        : null;
+    this.localClassNamesMap = new Map(
```
askoufis, line 129: If the key and value are the same, and the classname is just being used for a lookup, could this not be a `Set` instead?

askoufis (approved):

packages/css/src/transformCss.ts
askoufis: Answer from Matt: `Map` is faster in this case

packages/css/src/ahocorasick.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module 'ahocorasick' {
+  export default class Ahocorasick {
+    constructor(searchTerms: Array<string>);
+
+    search(input: string): Array<[endIndex: number, matches: Array<string>]>;
+  }
+}
```
momin-tf, line 7: Could submit this to DefinitelyTyped

packages/css/src/transformCss.test.ts
```diff
@@ -1564,4 +1567,44 @@ describe('transformCss', () => {
...
+        {
+          type: 'local',
+          selector: style1,
+          rule: {
+            selectors: {
+              [`${style2} &:before, ${style2} &:after`]: {
+                background: 'black',
+              },
+
+              [`_1g1ptzo1_1g1ptzo10 ${style1}`]: {
```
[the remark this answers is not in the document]
mattcompiles: These are simulating locally scoped styles. They get processed to have the class delimiter. That's what the test is for.

PR #847, opened 2022-09-26
Add Angular sample

sebtoombs:
# What

This PR adds a simple Angular app to the examples directory. The example closely copies the webpack-react example. The implementation should probably be considered as a reasonably robust proof of concept - it has not yet been used in production and has only been "tested" by one author (me).

# Why

For better or worse, some of us need to use Angular. Most existing Angular component libraries are poor quality. We are building an Angular component library for internal use, and I wanted to explore if it was possible to use a best-in-class styling solution such as vanilla-extract.

# How

The readme & webpack config cover most of the how, but tl;dr - follow the vanilla-extract install instructions for webpack, but make a slight modification to the Angular webpack config to avoid letting the angular webpack loader load `.css.ts` files. It's possible this could be done by modifying the Angular webpack loader to add an "emitter" for these files (if one knew how...).


mattcompiles (commented):
Hey Bud, Thanks for the PR. This repo uses is pnpm for package management. Could we drop the `package-lock.json` file?

sebtoombs:
> Hey Bud, Thanks for the PR. This repo uses is pnpm for package management. Could we drop the `package-lock.json` file?

Thanks Matt 🙏 sorry I wasn't sure whether I should update the root lock, as Angular often doesn't play well with others in this scenario (in my experience), and then I forgot to do anything about it.

I've updated to remove the `package-lock.json` and re run pnpm install to update the lockfile. I'm not sure what the best way to ensure this creates no issues with other packages is in this project.

askoufis (commented):
Just some small comments, primarily to do with trimming down the boilerplate/generated code to keep the example as small as possible.

askoufis:
@sebtoombs Are you still interested in getting this PR through?

askoufis:
@sebtoombs Closing this for now. Feel free to re-open it if you're still interested. However, now that Angular provides first-class Vite support, I don't think there's as much value in having a webpack + angular example in the repo. Setting up a Vite + angular codebase should be as simple as consuming the Vanilla Extract Vite plugin, and if it's not, please raise an issue!

PR #851, opened 2022-09-28
vite: Fix RSC integration

frandiox:
It looks like the RSC integration broke in https://github.com/vanilla-extract-css/vanilla-extract/pull/796 (release 3.4.0). The reason is that the code changed from `!process.env.VITE_RSC_BUILD` to `!!process.env.VITE_RSC_BUILD`.

This change fixes it. I've added it to the `else` part of the Astro/Solid integration to make sure we don't break those frameworks. It would be good to check that everything works there regardless -- cc @mechairoi @markdalgleish

### DIFF
--- a/.changeset/stupid-keys-notice.md
+++ b/.changeset/stupid-keys-notice.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Fix React Server Components integration at build time.
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -33,7 +33,7 @@ export function vanillaExtractPlugin({
   let postCssConfig: PostCSSConfigResult | null;
   const cssMap = new Map<string, string>();
 
-  let forceEmitCssInSsrBuild: boolean = !!process.env.VITE_RSC_BUILD;
+  let forceEmitCssInSsrBuild: boolean;
   let packageName: string;
 
   const getAbsoluteVirtualFileId = (source: string) =>
@@ -74,6 +74,8 @@ export function vanillaExtractPlugin({
         )
       ) {
         forceEmitCssInSsrBuild = true;
+      } else {
+        forceEmitCssInSsrBuild = !process.env.VITE_RSC_BUILD;
       }
     },
     resolveId(source) {


markdalgleish:
Is this logic correct?

The original diff you introduced was this, to skip the early return in SSR if `process.env.VITE_RSC_BUILD` is truthy:

```diff
-      if (ssr) {
+      if (ssr && !process.env.VITE_RSC_BUILD) {
        return addFileScope({
          source: code,
          filePath: normalizePath(validId),
          rootPath: config.root,
          packageName,
        });
      }
```

Since then it's been generalised to support other tools with a similar requirement for CSS output during SSR:

```diff
-      if (ssr && !process.env.VITE_RSC_BUILD) {
+      if (ssr && !forceEmitCssInSsrBuild) {
        return addFileScope({
          source: code,
          filePath: normalizePath(validId),
          rootPath: config.root,
          packageName,
        });
      }
```

So shouldn't `forceEmitCssInSsrBuild` be `true` if `process.env.VITE_RSC_BUILD` is truthy? This PR seems to do the opposite: `forceEmitCssInSsrBuild = !process.env.VITE_RSC_BUILD`

markdalgleish:
I'm wondering, is this a timing issue? The main change related to `process.env.VITE_RSC` seems to be that its value is now retrieved once up front and stored over the lifetime of the build, whereas your original PR looks up the value dynamically every time `transform` is called.

markdalgleish:
I ran the repro from the original issue and noticed a few things:

- The version of `@vanilla-extract/vite-plugin` in `package-lock.json` is v3.3.1 which predates the Astro update in v3.4: https://github.com/typeandpixel/demo-store/blob/issue/vanilla-extract-server-components/package-lock.json#L2642
- I've confirmed that the pre-Astro version of the Vite plugin is in node_modules while reproducing the issue.
- Rolling the Vite plugin back to the original v3.3.0 RSC release doesn't fix the issue.
- I added some console logs into the Vite plugin and noticed that `process.env.VITE_RSC_BUILD` is always `undefined`.

mechairoi:
I too think it may be a timing issue https://github.com/vanilla-extract-css/vanilla-extract/pull/851#issuecomment-1261458510. Sorry for the careless change. I tried to look into it, but couldn't make a working sample with `@shopify/hydrogen` and `@vanilla-extract/vite-plugin@3.3.0`.

michaeltaranto:
For now, we might close this one in favour of the [new `emitCssInSsr` option]. That should allow testing different logic external to the plugin, and if you find that the `VITE_RSC_BUILD` env var is not being respected correctly by the plugin we can look at a fix.

[new `emitCssInSsr` option]: https://github.com/vanilla-extract-css/vanilla-extract/pull/989

PR #865, opened 2022-10-05
Add onFileCompiled callback

DanielHeath:
Per https://github.com/vanilla-extract-css/vanilla-extract/discussions/864

This hook lets users supply the missing piece for integration with non-javascript backends - recording the class-name mapping used by vanilla-extract.

### DIFF
--- a/packages/esbuild-plugin/src/index.ts
+++ b/packages/esbuild-plugin/src/index.ts
@@ -22,11 +22,13 @@ interface VanillaExtractPluginOptions {
   externals?: Array<string>;
   runtime?: boolean;
   processCss?: (css: string) => Promise<string>;
+  onFileCompiled?: (path: string, identifiers: Record<string, string>) => void;
   identifiers?: IdentifierOption;
   esbuildOptions?: CompileOptions['esbuildOptions'];
 }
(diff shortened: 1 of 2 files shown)


DanielHeath:
The primary limitation of this callback as it stands is that it's called multiple times per build for the same file (possibly because I have multiple entrypoints importing the same component). 

If the classname mapping generation wasn't stable, it wouldn't be clear which one to use unless the current entrypoint was also passed in, which (AFAICT) esbuild doesn't provide to plugins.

mattcompiles:
Hey @DanielHeath 👋 

I assume what you ultimately want to produce is some metafile containing all the Vanilla Extract exports per file? This callback seems like a very low-level way to solve that issue. It's also potentially out of sync as more export processing occurs in the `stringifyExports` function. 

I want to run this by the rest of the team first, but would you be open to a different solution?

DanielHeath:
Yeah absolutely - this was a more of a quick hack to check I could do it.

DanielHeath:
Just wanted to check in about the timeline here - if it's likely to be more than a few months I might forge ahead and publish a plugin, but if it's being discussed currently & there's likely to be progress made I'd prefer to wait and see.

PR #869, opened 2022-10-05
Use async Babel transform in `integration/transform.ts`

mrm007:
Currently both transforms are synchronous because the Babel `transform` method used is just an alias for `transformSync` — see [here][1] and [here][2].

[2]: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/babel__core/index.d.ts#L398-L401
[1]: https://github.com/babel/babel/blob/main/packages/babel-core/src/transform.ts#L63-L72

### DIFF
--- a/.changeset/lovely-lies-hide.md
+++ b/.changeset/lovely-lies-hide.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/babel-plugin-debug-ids': patch
+---
+
+Use correct async Babel transform method in integration transform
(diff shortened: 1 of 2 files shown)


mattcompiles (commented):
Will need a changeset. LGTM otherwise

.changeset/lovely-lies-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/babel-plugin-debug-ids': patch
```
mattcompiles, line 2:
```suggestion
'@vanilla-extract/integration': patch
```

mrm007: Oops

mattcompiles (approved):

PR #879, opened 2022-10-13
Fix `normalizeResponsiveValue` example in sprinkles docs

reiss-d:
- Fixed incorrect file reference
- Fixed return values of `normalizeResponsiveValue`


michaeltaranto (approved):
Thank you 👏

PR #890, opened 2022-10-26
Add RuntimeFn as an export to recipes index

otaviomad:
Closes #739

"Partial" builds (simple tsc builds, no bundling) will emit wrong declaration when using recipes. This allows it to use the correct import.

Issue #739, opened 2022-06-27
ESM TypeScript Module Resolution breaks sprinkles type definitions


davidhousedev:
### Describe the bug

We are building a component library using Vanilla Extract and Sprinkles. We have been using CommonJS module resolution via `moduleResolution: 'node'` in `tsconfig.json`. We would like to migrate to ES Module resolution via `moduleResolution: 'node16'` in `tsconfig.json`. We are encountering issues with the Sprinkles package where the types generated via ES Module resolution are not specific, when they are when using CommonJS module resolution.

When using `moduleResolution: 'node'`, our type definitions look something like this:

*dist/sprinkles.css.d.ts*
```ts
export declare const sprinkles: import("@vanilla-extract/sprinkles/dist/declarations/src/createSprinkles").SprinklesFn<[{
    conditions: {
        defaultCondition: "mobile";
        conditionNames: ("mobile" | "tablet" | "desktop")[];
    };
    styles: {
        display: {
            values: {
                block: {
                    defaultClass: string;
                    conditions: {
                        mobile: string;
                        tablet: string;
                        desktop: string;
                    };
                };
                flex: {
                    defaultClass: string;
                    conditions: {
                        mobile: string;
                        tablet: string;
                        desktop: string;
                    };
                };
                none: {
                    defaultClass: string;
                    conditions: {
                        mobile: string;
                        tablet: string;
                        desktop: string;
                    };
                };
            };
        };
    // ...
}
```

However, when we use `moduleResolution: 'node16'`, the type definition looks like this:

*dist/sprinkles.css.d.ts*
```ts
export declare const sprinkles: any;
```

### Reproduction

https://github.com/davidhousedev/vanilla-extract-esm-issue-reproduction

### System Info

```shell
System:
    OS: macOS 11.6.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 311.74 MB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.15.0 - ~/.nvm/versions/node/v16.15.0/bin/node
    Yarn: 1.22.18 - ~/.nvm/versions/node/v16.15.0/bin/yarn
    npm: 8.5.5 - ~/.nvm/versions/node/v16.15.0/bin/npm
  Browsers:
    Chrome: 102.0.5005.115
    Firefox: 101.0
    Safari: 15.5
  npmPackages:
[...]

PR #907, opened 2022-11-09
export types needed for typescript inference with esm

pkieltyka:
additional fixes for https://github.com/vanilla-extract-css/vanilla-extract/issues/739 similar to https://github.com/vanilla-extract-css/vanilla-extract/pull/890 which was missing a few types necessary for export

### DIFF
--- a/.changeset/tasty-olives-tell.md
+++ b/.changeset/tasty-olives-tell.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/recipes': minor
+---
+
+Add RuntimeFn as an export to recipes index
--- a/packages/css/src/vars.ts
+++ b/packages/css/src/vars.ts
@@ -11,6 +11,8 @@ import { Tokens, NullableTokens, ThemeVars } from './types';
 import { validateContract } from './validateContract';
 import { generateIdentifier } from './identifier';
 
+export type { Contract, MapLeafNodes, CSSVarFunction } from '@vanilla-extract/private';
+
 export function createVar(debugId?: string): CSSVarFunction {
   const cssVarName = cssesc(
     generateIdentifier({
--- a/packages/recipes/src/index.ts
+++ b/packages/recipes/src/index.ts
@@ -10,7 +10,7 @@ import type {
   VariantSelection,
 } from './types';
 
-export type { RecipeVariants } from './types';
+export type { RecipeVariants, RuntimeFn } from './types';
 
 function mapValues<Input extends Record<string, any>, OutputValue>(
   input: Input,
--- a/packages/recipes/src/index.ts
+++ b/packages/recipes/src/index.ts
@@ -10,7 +10,7 @@ import type {
   VariantSelection,
 } from './types';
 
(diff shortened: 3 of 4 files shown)


johndarmstrong:
bump - could someone take a look at exporting these additional types? This is currently a blocker from using our component library in modern tooling, without having to patch vanilla extract in our build pipeline.

pkieltyka:
Strange a trivial fix like this takes so much time to review / merge. Where the maintainers at?

awhiteside1:
Just ran into this issue as well - I think I was able to workaround it by basically isolating the theme contract  and disabling emitting declarations for that package, but definitely not ideal for something  fairly trivial to handle properly.

awhiteside1 (approved):

awhiteside1:
Might have been a lack of changeset, added here: https://github.com/pkieltyka/vanilla-extract/pull/1

michaeltaranto:
Hey, we were discussing this one today and want to understand the use case a bit more clearly before we take on the increase to the public API footprint.

Can you provide a simple reproduction example demoing the use case?

jungpaeng:
@michaeltaranto Hi, I'm using Node16 moduleResolution.

Here's the problem I'm currently having, I'll write it briefly to clarify the error.

  ```ts
// Button.css.ts
import { recipe, RecipeVariants } from '@vanilla-extract/recipes';

export type ButtonContainerVariants = RecipeVariants<typeof buttonContainer>;

export const buttonContainer = recipe({
  variants: {
    test: {
      test1: {
        display: 'flex',
      },
    },
  },
});
```

```tsx
// Button.tsx
import React from 'react';

import { ButtonContainerVariants } from './Button.css.js';

type Props = ButtonContainerVariants & {
  test?: 'test1';
  children?: React.ReactNode;
};
```

```tsx
export function Button({ ...props }: Props) {
  return <button {...props} />;
}

// Boo.tsx
function Boo() {
  return <Button />;
}
```

There's an issue with the test's type inference on the Button inside the Boo component. Here's a picture of it
image

The reason for this issue is as follows.

Node16 moduleResolution and above can only read exported types, so the `VariantGroups` I return as a result of `RecipeVariants` are not exported, causing this issue.

This is causing me a lot of inconvenience. Is there something wrong with my explanation?

mrm007:
Hi @jungpaeng & @pkieltyka ,

I was not able to reproduce the issue. I used the code from this comment https://github.com/vanilla-extract-css/vanilla-extract/pull/907#issuecomment-1474265207 and type inference on the `Button` inside the `Boo` component worked as expected:

image

In my attempt to reproduce the issue, I have tested with:

- `pnpm@7.30.0` with default config
- `typescript@5.0.4` or `typescript@4.9.5` installed
- `"type": "module"` in `package.json`
- `"module": "ESNext"` (or `NodeNext`) and `"moduleResolution": "node16"` (or `nodenext`) in `tsconfig.json`

Let me know if I missed something. A repo with more specific code and steps to reproduce would be helpful. We'll close the issue for now, but if you can provide more details, we'll be happy to reopen it.

PR #912, opened 2022-11-12
next-plugin: add types to plugin config

nix6839:
Next plugin doesn't currently have `d.ts` file. So user can't use intellisense for options. This PR provides that.

### DIFF
--- a/.changeset/good-cats-boil.md
+++ b/.changeset/good-cats-boil.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/next-plugin': patch
+---
+
+Provide intellisense for users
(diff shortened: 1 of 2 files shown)


graup:
Hi @nix6839!

Is adding `@vanilla-extract/integration` as a dependency just for the type really necessary? Maybe it would it be possible to pull out the type for the plugin options from the plugin, e.g. something like

```ts
type PluginOptions = ConstructorParameters<VanillaExtractPlugin>[0];
```

This way, we don't create another type to maintain. Could you test if that works?

.changeset/good-cats-boil.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/next-plugin': patch
+---
+
+Provide intellisense for users
```
graup, line 5:
```suggestion
Add types to Nextjs plugin configuration
```

graup (approved):
LGTM

mattcompiles (approved):

PR #925, opened 2022-11-22
vite: support HMR

wmertens:
With this change, hot reload works in Qwik and presumably anything else using Vite.

The problem was that a change in css.ts didn't force a regen of the CSS.

### DIFF
--- a/.changeset/slimy-llamas-dream.md
+++ b/.changeset/slimy-llamas-dream.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': minor
+---
+
+Hot module reloading now works for Qwik
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -19,6 +19,7 @@ const styleUpdateEvent = (fileId: string) =>
 
 const virtualExtCss = '.vanilla.css';
 const virtualExtJs = '.vanilla.js';
+const virtualRE = /.vanilla.(css|js)$/;
 
 interface Options {
   identifiers?: IdentifierOption;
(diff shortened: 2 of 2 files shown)


endigma:
What's pending here? This would majorly help us.

mattcompiles:
Hey @wmertens 👋 

This PR implies that we don't currently support HMR in Vite but that's not true. Are you able to provide a reproduction of the issue that's being resolved here? When testing locally, HMR seems to work fine so I'm not sure why it doesn't work in your scenario?

Also sorry, I know this PR has been open for a while.

wsow4:
@mattcompiles 
I can confirm the bug exists. HMR works in general, but not in every case. 
`server.moduleGraph.getModulesByFile(virtualId)` sometimes returns more than one file and each of that file should be invalidated, but VE plugin invalidates only the first one.

MicMonen:
Waiting on this as well.

mattcompiles:
@wsow4 Thanks for jumping in, I don't doubt you. Just without a reproduction to test against how can we say something has been resolved? If someone provides a reproduction of this issue we will definitely try get it resolved ASAP.

wmertens:
@mattcompiles Try changing the text color in https://stackblitz.com/edit/qwik-starter-r2tt19?file=src/routes/styles.css.ts - it only works when you restart the server

Note that it's not doing hot reload, just reload.

mattcompiles:
@wmertens Thanks bud. I'll take a look.

askoufis:
> @mattcompiles Try changing the text color in https://stackblitz.com/edit/qwik-starter-r2tt19?file=src/routes/styles.css.ts - it only works when you restart the server
> 
> Note that it's not doing hot reload, just reload.

@wmertens Implementing the fix mentioned by @wsow4 where we invalidate all modules returned by `server.moduleGraph.getModulesByFile(virtualId)` results in at least a page reload when updating a style in your reproduction. I'll make a separate PR for that change.

The crux of the issue here is that the Vanilla Extract vite plugin handles HMR for vanilla modules, but it _does not_ handle HMR for plain `.css` files, which are what's being emitted due to [logic within the plugin that detects other plugins](https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/vite-plugin/src/index.ts#L82) (namely the qwik plugin).

If I'm understanding it correctly, you've effectively implemented HMR for emitted CSS. IMO this shouldn't be the responsibility of the Vanilla Extract vite plugin, but rather the qwik vite plugin. In fact, I'm assuming it already has some kind of support for CSS HMR already, so maybe what's needed is some extra logic to handle CSS emitted by vanilla.

wmertens:
@askoufis no, this PR doesn't implement CSS hot reload, it only re-parses the .css.ts file when it should be, but your solution is more elegant. Closing.

PR #931, opened 2022-11-28
fix typo

terribIecoder:


acf77 (approved):

acf77 (approved):

site/docs/overview/getting-started.md
```diff
@@ -39,10 +39,10 @@ export const container = style({
 });
 ```
 
-Importantly, this is does two things:
+Importantly, this does two things:
 
-1. Creates a locally scoped class,
-2. Exports the generated class name,
+1. Creates a locally scoped class.
+2. Exports the generated class name.
```
askoufis, line 45: I'd argue that because these items form part of a sentence, but are not sentences themselves, only the last item should end with a period, and they should not start with capital letters. I think this list is best represented as a numbered fragment list, as described [here](https://www.stylemanual.gov.au/structuring-content/lists#choose_a_type_of_list).

```suggestion
1. creates a locally scoped class.
2. exports the generated class name.
```

michaeltaranto (approved):

Issue #941, opened 2022-12-09
Module declaration bundling causing downstream TypeScript error with `ahocorasick`


AndrewLeedham:
### Describe the bug

The `.d.ts` file for `ahocorasick` seems to be bundled into the `transformCss` typings which causes a downstream error as `declare module '...'` cannot be used in a module: https://github.com/microsoft/TypeScript/issues/9748#issuecomment-232822688

Reproduction below requires running `npm run types` in the console.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-sxrsgv?file=src/main.ts

### System Info

```shell
  System:
    OS: macOS 12.4
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 814.13 MB / 16.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
  npmPackages:
    @vanilla-extract/esbuild-plugin: 2.2.0 => 2.2.0 
    @vanilla-extract/vite-plugin: 3.7.0 => 3.7.0 
    esbuild: 0.14.50 => 0.14.50 
    rollup: 2.77.0 => 2.77.0 
    vite: 3.1.3 => 3.1.3
```

### Used Package Manager

yarn

### Logs

```shell
../node_modules/@vanilla-extract/css/transformCss/dist/vanilla-extract-css-transformCss.cjs.d.ts(3,16): error TS2665: Invalid module name in augmentation. Module 'ahocorasick' resolves to an untyped module at '../node_modules/ahocorasick/src/main.js', which cannot be augmented.
```

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

AndrewLeedham:
@mattcompiles Seems to me like the [`.d.ts` shouldn't be imported](https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/css/src/transformCss.ts#L1) and should just be copied across to the build outpuut

PR #943, opened 2022-12-09
Remove ahocorasick.d.ts import

AndrewLeedham:
fixes: #941 

Looks like it doesn't need to be copied across because `ahocorasick` is not referenced in the outputted types 👍


.changeset/many-doors-prove.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Removed unused ahocorasick declaration reference
```
askoufis, line 5: This is describing the solution rather than the problem if fixes. Could be reworded to something like:
```suggestion
Fix a `tsc` error caused by a dependency's type declarations being unnecessarily bundled
```

michaeltaranto (approved):

PR #942, opened 2022-12-09
feat: integration callbacks

astahmer:
hey, made a vite plugin that allows purging any unused style generated by vanilla-extract/sprinkles

https://twitter.com/astahmer_dev/status/1601244606133567488

it needs some minor changes (mostly onEvaluated+serializeVanillaModule callbacks for the integration package + forceEmitCssInSsrBuild option for the vite plugin)

this PR adds:
- a `onEvaluated` callback to `packages/integration/src/processVanillaFile.ts` (very similar to what was proposed [here](https://github.com/vanilla-extract-css/vanilla-extract/pull/865))
- a `serializeVanillaModule` callback to `packages/integration/src/processVanillaFile.ts`, pretty much like the current `serializeVirtualCssPath` callback, it allows customizing the behaviour
- integrates the HMR fix from [this PR](https://github.com/vanilla-extract-css/vanilla-extract/pull/925) 
- exports the default implementation of `serializeVanillaModule`
- makes changes to the plugins integration (vite/webpack/esbuild/rollup) by allowing to pass those new callbacks to the 
`processVanillaFile` call
- exports the `ProcessVanillaFileOptions` & `SprinklesProperties` type
- creates an `AdapterContext` (which is just a reference to the result of the adapter after the `evalCode` of `processVanillaFile` so anyone can retrieve the generated class names / CSS rules by file scopes
- allows customizing the `forceEmitCssInSsrBuild` option for the vite plugin (needed to make the HMR work with vite-plugin-ssr), would probably allow removing the hard-coded checks (`['astro:build', 'solid-start-server', 'vite-plugin-qwik'].includes()`) 

let me know if there's any questions or things you'd like to see

PR #952, opened 2022-12-21
vite-plugin: Whitelist `vite-plugin-svelte`

### LINKED ISSUE #937: Incorrect classnames during SSR
### Describe the bug

During server rendering, the target div is rendered with class `routes__1ry5ddd0`, which has no corresponding styles. It hydrates to `_1ry5ddd0`, which has the correct styles. 

The provided example is nothing but the SvelteKit starter (pnpm create svelte@latest) with the vite plugin and a basic style added. It can be easily observed by comparing the hydrated DOM with "view page source" etc. You can also observe the lack of styling using an addon like uBlock Origin or uMatrix or NoScript etc to block JS.

You can clone the repo, run `pnpm install`, `pnpm dev` to see the issue. 

The most outwardly visible symptom is a FOUC, you may or may not be able to see it depending on your setup. 

The reproduce conditions are a little complicated, so I've created a table to help. There may be other reproduce conditions I'm not aware of yet, but these work on the repo I provided.

| Running Mode                 | `identifiers` | Reproduce? |
| ---------------------------- | ------------- | ---------- |
| `vite dev`                  | `"short"`     | Yes        |
| `vite dev`                   | `"debug"`     | Yes        |
| `vite build && vite preview` | `"short"`     | Yes        |
| `vite build && vite preview` | `"debug"`     | No         | 

### Reproduction

https://github.com/endigma/vanilla-extract-ssr-bug

### System Info

```shell
System:
    OS: Linux 6.0 Void
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 6.65 GB / 31.26 GB
    Container: Yes
    Shell: 3.5.1 - /usr/bin/fish
  Binaries:
    Node: 16.15.1 - ~/.asdf/shims/node
    Yarn: 1.22.18 - /usr/bin/yarn
    npm: 8.11.0 - ~/.asdf/shims/npm
  Browsers:
    Chromium: 107.0.5304.87
    Firefox: 107.0
  npmPackages:
    @vanilla-extract/css: ^1.9.2 => 1.9.2 
    @vanilla-extract/vite-plugin: ^3.7.0 => 3.7.0 
    vite: ^3.2.4 => 3.2.4
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

endigma:
- Add explicit exception for when plugins includes `vite-plugin-svelte`.
- Add a user configurable override

Resolves #937

I'm not sure how this plays into #851, that PR seems dead/invalid?

This fixes SSR support in sveltekit, at least, and provides an override for other framework users to use before built-in support is added.


astahmer:
FWIW I also added an option for `forceEmitCssiInSsrBuild` [here](https://github.com/vanilla-extract-css/vanilla-extract/pull/942), which show it's needed

endigma:
> FWIW I also added an option for `forceEmitCssiInSsrBuild` [here](https://github.com/vanilla-extract-css/vanilla-extract/pull/942), which show it's needed

Nice, I do disagree with you about "would probably allow removing the hard-coded checks", the point of these is to reduce friction in cases where it's definitely required. This is good DX.

Let's home something gets merged soon.

astahmer:
ah yes you're right, it makes sense for the DX !

endigma:
What does this need to be merged?

endigma:
@michaeltaranto @mattcompiles @markdalgleish

What needs to be done here? This fixes VE usage for SvelteKit and probably others. VE is practically unusable on non-whitelisted frameworks right now as SSR FOUC is a problem.

michaeltaranto:
Discussed this one at the maintainers meeting this morning and we are going to split this into two so we can get sveltekit out the door. We will look at adding the forceSSR flag in a separate PR along with documentation.

michaeltaranto (approved):

PR #954, opened 2023-01-03
Upgrade to Jest 29

mrm007:
Jest 29 comes with [new snapshot format defaults](https://jestjs.io/blog/2022/08/25/jest-29).

I upgraded Jest in the layers PR but thought I'd make it a separate PR.


askoufis (approved):
🎪

PR #955, opened 2023-01-03
Support `@layer`

mrm007:
Added 2 new top-level APIs to declare layers (`layer` and `globalLayer`), and a way to attach style rules to a layer with the `@layer` key.

```ts
const base = globalLayer('base'); // uses the provided name
const lib = globalLayer({ parent: base }, 'lib'); // with parent

const utils = layer({ parent: lib }); // creates a hashed name
const reset = layer({ parent: lib }, 'reset'); // with parent + debug id
const typography = layer('typography'); // with debug id

// attach to a layer
const text = style({
  '@layer': {
    [typography]: {
      backgroundColor: 'indigo'
    }
  }
});
// also works with global styles
globalStyle('p', {
  '@layer': {
    [lib]: {
      color: 'orange'
    }
  }
});
```

See [tests], [layers fixture][fixture] and [resulting CSS][css] for details.

[tests]: https://github.com/vanilla-extract-css/vanilla-extract/blob/layers/packages/css/src/transformCss.test.ts#L1070
[fixture]: https://github.com/vanilla-extract-css/vanilla-extract/blob/layers/fixtures/layers/src/styles.css.ts
[css]: https://github.com/vanilla-extract-css/vanilla-extract/blob/layers/tests/stylesheets/__snapshots__/layers.test.ts.snap

TODO:

- [x] Some tests are failing because I bumped Jest but there are still many snapshots which need to be updated. Merge #954 first.
- [x] `layer` needs a fix for automatic debug id insertion. Merge #956 first.
- [x] Docs
	- Styling > Layers https://layers--vanilla-extract-css.netlify.app/documentation/styling/#layers
	- `layer` API https://layers--vanilla-extract-css.netlify.app/documentation/api/layer/
	- `globalLayer` API https://layers--vanilla-extract-css.netlify.app/documentation/global-api/global-layer/
- [x] Detailed changeset

PR #827, opened 2022-09-14
Remove filescoping from babel-plugin, Add transform integration APIs

mattcompiles:
This PR does the following things:

- Removes filescope appending from babel plugin
- Adds a new package (`@vanilla-extract/jest`) to be used instead of using the babel plugin
- Adds new `transform` and `transformSync` APIs that handle filescoping and automatic debug Ids
- Updates all the existing integrations to use the new transform APIs internally
  - This results in OOTB automatic debug Id support in all integrations

PR #956, opened 2023-01-04
Fix auto debugId for `styleVariants`

mrm007:
Looks like a bug slipped through in #827:

https://github.com/vanilla-extract-css/vanilla-extract/blob/c70aaaf1fa3e31e9a27f04fdcc4bac2b93c1641c/packages/babel-plugin-debug-ids/src/index.test.ts#L413-L417

This fix is also needed in #955 where `layer()` can have up to 2 arguments:

```ts
const lib = layer();
const utils = layer({ parent: lib });
const reset = layer({ parent: lib }, 'my-reset');
const other = layer('my-other'); // don't add a debug id
```


jahredhope (approved):
Makes sense. Nice work.

.changeset/green-brooms-sniff.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/babel-plugin-debug-ids': patch
+'@vanilla-extract/integration': patch
+---
+
+Don't crash on [`satisfies` operator](https://devblogs.microsoft.com/typescript/announcing-typescript-4-9/#satisfies)
```
michaeltaranto, line 6:
```suggestion
Ensure support for new TypeScript language features by bumping the minimum required Babel dependency versions.
```

michaeltaranto (approved):

PR #967, opened 2023-01-16
feat: It is meaningless to use different values for the same key name

idler8:
OK?
What might be the impact?

### DIFF
--- a/.changeset/gold-rice-approve.md
+++ b/.changeset/gold-rice-approve.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': major
+---
+
+Changing the output of array parameters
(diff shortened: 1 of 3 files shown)


graup:
```
color: blue;
color: red;
color: yellow;
```
This example may look meaningless, but the feature actually is not. There are cases where you want to provide fallback values for different browsers, for example
```
const layoutStyles = defineProperties({
  properties: {
    overflow: {
      hidden: 'hidden',
      auto: {
        overflow: 'auto',
      },
      overlay: {
        overflow: ['auto', 'overlay'], // in Firefox and IE `overlay` will be ignored and `auto` will be applied
      },
  },
```
Now `atoms({ overflow: 'overlay' })` creates the CSS `{ overflow: auto; overflow: overlay }`, which uses overlay in supported browsers and auto otherwise.

idler8:
👌

Issue #963, opened 2023-01-11
Cannot import from *.css.js files: module has no exports (Reintroduced bug)


roginfarrer:
### Describe the bug

I believe [this bug](https://github.com/vanilla-extract-css/vanilla-extract/issues/118), which was fixed when the `@vanilla-extract/babel-plugin` was required to use the webpack plugin, was reintroduced in `@vanilla-extract/webpack-plugin@2.2.0`. I was able to reproduce the issue in the codesandbox provided in that issue just by updating the webpack config to use the latest version of the webpack plugin and omitting the babel plugin.

I could resolve the issue by reintroducing the babel plugin and downgrading `@vanilla-extract/webpack-plugin` to `2.1.12`.

### Reproduction

https://codesandbox.io/s/cannot-import-from-css-js-files-module-has-no-exports-ct6ryh?file=/src/test.css.js:0-208

### System Info

```shell
System:
    OS: Linux 5.4 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (16) x64 Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz
    Memory: 1.93 GB / 62.72 GB
    Container: Yes
    Shell: 5.0.3 - /bin/bash
  Binaries:
    Node: 12.22.12 - ~/.nvm/versions/node/v12.22.12/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v12.22.12/bin/yarn
    npm: 6.14.16 - ~/.nvm/versions/node/v12.22.12/bin/npm
  npmPackages:
    @vanilla-extract/css: 1.9.2 => 1.9.2
    @vanilla-extract/sprinkles: ^0.2.0 => 0.2.0
    @vanilla-extract/webpack-plugin: 2.2.0 => 2.2.0
    webpack: ^5.36.1 => 5.36.1
```

### Used Package Manager

yarn

### Logs

```shell
WARNING in ./src/App.tsx 71:25-40
export 'test' (imported as 'testStyles') was not found in './test.css' (module has no exports)
```

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

m7yue:
the same error i use in vite, how to resolve it!

[...]

Issue #940, opened 2022-12-08
errors when using external ui library implemented with vanilla-extract in next.js app


jneander:
### Describe the bug

In a Next.js app using the @vanilla-extract/next-plugin to support VE locally, everything works as expected. However, when trying to use view code from an external library that has styles implemented with vanilla-extract, there are render errors.

The minimal example for can be found here:
* [Next.js app](https://github.com/jneander/example-app-using-third-party-vanilla-extract-ui-library)
* [@jneander/x-throwaway-vanilla-extract-ui library](https://www.npmjs.com/package/@jneander/x-throwaway-vanilla-extract-ui) (This is used by the app linked above and is linked here for reference.)

The following behavior is observed while the Next.js app is running in dev mode:

A. The external code is not used on the page.
1. Refresh the page.
2. Observe that the page loads successfully.

B. The external code is added to the page.
1. Observe that the page injects the external code to the page successfully.
2. Refresh the page.
3. Observe that the page displays the error described below.

```
Server Error
Error: Styles were unable to be assigned to a file. This is generally caused by one of the following:

- You may have created styles outside of a '.css.ts' context
- You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started
```

### Reproduction

https://github.com/jneander/example-app-using-third-party-vanilla-extract-ui-library

### System Info

```shell
System:
  OS: macOS 13.0
  CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  Memory: 359.04 MB / 16.00 GB
  Shell: 3.2.57 - /bin/bash
Binaries:
  Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
  Yarn: 1.22.17 - /usr/local/bin/yarn
  npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
Browsers:
  Chrome: 108.0.5359.94
  Firefox: 97.0.2
  Safari: 16.1
npmPackages:
  @vanilla-extract/css: ^1.9.2 => 1.9.2 
  @vanilla-extract/next-plugin: ^2.1.1 => 2.1.1
```

### Used Package Manager

npm

### Logs

```shell
Server Error
Error: Styles were unable to be assigned to a file. This is generally caused by one of the following:

- You may have created styles outside of a '.css.ts' context
- You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started

[...]

Issue #850, opened 2022-09-28
NonErrorEmittedError when used in next.js and monorepo project


Kalesberg:
### Describe the bug

**Summary**
Loading Vanilla vars externally from monorepo package results in error when used in next.js project

**Environment**
next.js: 12.x
vanilla-extract/css: 1.7+
lerna: 5.4

**Reproduce**
Just clone the repo and run `yarn dev`

**Screenshot**
<img width="648" alt="Screen Shot 2022-09-28 at 04 51 31" src="https://user-images.githubusercontent.com/4900126/192736637-afd46a45-e74e-4a41-847c-d1bfb9515923.png">

**Highlighted snippets**
src/pages/styles.css.ts
```
import { style } from '@vanilla-extract/css';
import { vars } from '@kalesberg/uikit';

export const textSample = style({
  color: vars.color.secondary
});
```
And when using `textSample` class in next.js page ends up with the error posted above.

### Reproduction

https://github.com/Kalesberg/vanilla-issue-demo

### System Info

```shell
System:
    OS: macOS 12.1
    CPU: (8) arm64 Apple M1
    Memory: 1.07 GB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.13.0 - ~/.nvm/versions/node/v16.13.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.1.0 - ~/.nvm/versions/node/v16.13.0/bin/npm
  Browsers:
    Brave Browser: 104.1.42.88
    Chrome: 105.0.5195.125
    Safari: 15.2
  npmPackages:
    @vanilla-extract/css: ^1.9.1 => 1.9.1
    @vanilla-extract/next-plugin: ^2.1.0 => 2.1.0
```

### Used Package Manager

yarn

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

peterfoeng:
Hi @Kalesberg did you ever figure out what's happening?

roginfarrer:
I'm actually getting a very similar error, but with the Webpack plugin (which the next plugin uses)

```
[...]

Issue #679, opened 2022-05-09
"ReferenceError: $RefreshReg$ is not defined" when using React + Yarn Workspaces


Jpoliachik:
### Describe the bug

I encountered this problem in our Yarn Workspaces setup, where a shared `design-system` package uses Vanilla Extract to house token values, it creates a theme, and also contains shared React components. These are all exported through a barrel `index.ts` file. 

We have two React apps that import from this - a Create React App and a Next App. Both apps encounter this error during compilation when importing from `design-system` in development mode. The CRA app can 'bypass' it by disabling Fast Refresh altogether, but this isn't an option for Next. 

In my testing, I believe this error only occurs when a React component is exported alongside the Vanilla Extract `createTheme` in a separate workspace package. In testing I was able to `createTheme` in a standalone package and import and use it successfully in the Next app package. And I was able to export standalone React components with applied Vanilla Extract `style()` without issue. Only when using React components _and_ `createTheme` in the same package did I encounter this issue. 

```
error - ./pages/home.css.ts
NonErrorEmittedError: (Emitted value instead of an instance of Error) ReferenceError: $RefreshReg$ is not defined
```

Any ideas what might be causing this? 

### Link to reproduction

I created a simple repo to reproduce:
https://github.com/Jpoliachik/ve-workspace-react-issue

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

```node
  System:
    OS: macOS 12.2
    CPU: (8) arm64 Apple M1
    Memory: 658.64 MB / 16.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 3.2.0 - ~/.nvm/versions/node/v16.14.0/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  Browsers:
    Chrome: 101.0.4951.54
    Safari: 15.3
```

jose-elias-alvarez:
[...]

PR #970, opened 2023-01-19
Update addFileScope to handle commonjs source files

roginfarrer:
I think this should address the issues noted in #940 and #850 and #963

In my case, I had `.css.js` files in `node_modules` that looked something like this:

```js
'use strict';

Object.defineProperty(exports, "__esModule", {
  value: true
});

const _css = require('@vanilla-extract/css');

var myStyle = _css.style({});

exports.myStyle = myStyle;
```

When run through `addFileScope`, this would prepend an ESM style import to a CJS file, and would cause errors like those noted in the issues above.

I'm not 100% sure this is the fix for all scenarios (I think there may be more cases across the VE codebase where this may happen), but I wanted to push this up for discussion. It's not the most elegant way to ascertain whether a file is ESM or CJS, but it worked in my brief testing.


mihkeleidast:
I can confirm that this resolves my issues similar to #940 and #963, though I still get `NonErrorEmittedError` sometimes (for that the workaround in #679 works at the moment).

mihkeleidast:
@roginfarrer perhaps you can add a changeset as well, then there shouldn't be any more things blocking this...

mattcompiles (commented):
Thanks for the PR. @roginfarrer 

I'm ok with this change but feel a little uncomfortable owning the detection of whether a file is ESM or CJS. Could we instead use a tested library to handle this? I'd suggest `hasESMSyntax` from [mlly](https://www.npmjs.com/package/mlly?activeTab=readme).

roginfarrer:
@mattcompiles Good call with using a library 👍🏻. Updated!

mattcompiles (approved):
Thanks for the PR mate 🍻

PR #976, opened 2023-01-28
feat(recipes): export VariantDefinitions

m-shaka:
Hi. I REALLY love vanilla-extract!
I wanted to use `VariantDefinitions` not being exported.
I can calculate it from the type of `recipe` but it's useful if exported.

Here's my use case.

```ts
// Button.css.ts
const size = {
  small: {
    height: "24px",
  },
  medium: {
    height: "32px",
  },
} as const satisfies VariantDefinitions;

export type Size = keyof typeof size;

export const button = recipe({
  base: {/* */},
  variants: {
    size
  }
})

// Button.tsx
import * as styles from './Button.css'

export const Button: React.FC<{ size: styles.Size }> = ({ size }) =>
  <button className={styles.button({ size })}>/* */</button>
```

### DIFF
--- a/packages/recipes/src/index.ts
+++ b/packages/recipes/src/index.ts
@@ -10,7 +10,7 @@ import type {
   VariantSelection,
 } from './types';
 
-export type { RecipeVariants, RuntimeFn } from './types';
+export type { RecipeVariants, RuntimeFn, VariantDefinitions } from './types';
 
 function mapValues<Input extends Record<string, any>, OutputValue>(
   input: Input,


michaeltaranto:
Hey, would you be able to clarify the use case here? From looking at your example, would the [`RecipeVariants`] utility work for you?

[`RecipeVariants`]: https://vanilla-extract.style/documentation/packages/recipes/#recipevariants

m-shaka:
Sorry, I didn't know that.
`RecipeVariants` can extract the variant type and that was one of my points, you know.
However, `VariantDefinitions` can also make it easier to separate variant definitions because the type checker can help you by using the `satisfies` operator.

michaeltaranto:
As a general rule, we try to minimise the API footprint as it makes us less vulnerable to breaking changes in the future. In this case we would prefer not to export internal types unless there's a use case that cannot be achieved without it.

m-shaka:
I understood your point! Thank you for your explanation.

Issue #962, opened 2023-01-09
Jest with jest-transform still giving (0 , _sprinkles.s) is not a function error


Nvveen:
### Describe the bug

According to [Test Environment](https://vanilla-extract.style/documentation/test-environments/), adding the transform to the Jest options, should solve this issue. There was a change since #731  so I'm not sure what I'm missing. My setup is a bit big, since I use Nx (and in my real project a lot more), but I managed to reproduce it with a sort-of minimal example.

![image](https://user-images.githubusercontent.com/486978/211430780-8c028184-e478-4ead-a38b-ddf9d4308e88.png)

### Reproduction

https://github.com/Nvveen/nx-sprinkles-repro

### System Info

```shell
System:
    OS: Linux 5.10 Debian GNU/Linux 10 (buster) 10 (buster)
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
    Memory: 9.98 GB / 15.48 GB
    Container: Yes
    Shell: 5.7.1 - /usr/bin/zsh
  Binaries:
    Node: 16.19.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.3 - /usr/local/bin/npm
  npmPackages:
    @vanilla-extract/css: ^1.9.2 => 1.9.2 
    @vanilla-extract/jest-transform: ^1.0.1 => 1.0.1 
    @vanilla-extract/sprinkles: ^1.5.1 => 1.5.1 
    @vanilla-extract/vite-plugin: ^3.7.0 => 3.7.0 
    vite: ^4.0.1 => 4.0.4
```

### Used Package Manager

pnpm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

rothsandro:
The Nx Jest Resolver uses `identitiy-obj-proxy` to resolve CSS files, see [`resolver.ts`](https://github.com/nrwl/nx/blob/5970246b51c17428b8d9f04123b06f780ac824bc/packages/jest/plugins/resolver.ts#L43-L51).

asherccohen:
anyone got this to work?

jd-oconnor:
> anyone got this to work?

[...]

PR #982, opened 2023-01-30
Add CSS transformer solution to Jest docs

roginfarrer:
This solution published by @jd-oconnor is very elegant, and I think it's worth including in the documentation. See https://github.com/vanilla-extract-css/vanilla-extract/issues/962#issuecomment-1384742733

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
 strict-peer-dependencies=false
+registry=https://registry.npmjs.org/
--- a/.changeset/tasty-panthers-cheer.md
+++ b/.changeset/tasty-panthers-cheer.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/jest-transform': minor
+---
+
+If a regular CSS file is sent through to the transform, return a module that exports the filename.
--- a/packages/jest-transform/src/index.ts
+++ b/packages/jest-transform/src/index.ts
@@ -1,10 +1,24 @@
+import path from 'path';
 import type { Transformer } from '@jest/transform';
-import { transformSync, getPackageInfo } from '@vanilla-extract/integration';
+import {
+  transformSync,
+  getPackageInfo,
+  cssFileFilter,
+} from '@vanilla-extract/integration';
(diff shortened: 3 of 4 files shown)


mattcompiles:
Rather than documenting this, I wonder if we should just merge it directly into the transformer? Might save some peoples the hassle?

roginfarrer:
Cool! Updated to integrate it into the transform. 

One thing I thought I'd check is if we might want to do something like [identity-obj-proxy](https://github.com/keyz/identity-obj-proxy) instead of returning the filename. I think this is generally a more common practice for regular CSS files in Jest.

.npmrc
```diff
@@ -1 +1,2 @@
-strict-peer-dependencies=false
\ No newline at end of file
+strict-peer-dependencies=false
+registry=https://registry.npmjs.org/
```
mattcompiles, line 2: Isn't this the default value? Any context on why you added this?

mattcompiles (approved):
@roginfarrer [identity-obj-proxy](https://github.com/keyz/identity-obj-proxy) is normally used to mock CSS modules. While we could swap the behaviour, I think we should suggest what other frameworks say and suggest CSS modules use the `.module.css` extension. At the end of the day, this is just a nice to have to save some people some trouble not a first class feature. 

Thanks for the PR and quick response as always 🍻

.npmrc
roginfarrer: Sounds good to me 👍🏻 

It is the default value, but as someone who works with different registries, I frequently run into the situation where I switch projects and the previous (or global) registry setting applies to the new project. So it's pretty easy to accidentally add a different registry to the lock file when installing dependencies because you don't get any warning when that happens. This just enforces that default value regardless of the users' settings.

roginfarrer: Happy to drop it if you don't want it!

mattcompiles: Makes sense, no problems. Don't mind these kinds of changes, just want some context in history for why it was added if it ever became as issue.

PR #994, opened 2023-02-07
Add transform wrapper method, use it for `@keyframes` rules

askoufis:
Adding a `transformProperties` method that wraps `transformContent` and `pixelifyProperties`. Applying this to `@keyframes` rules as well to make it consistent with the other rules (missed this in #990).

### DIFF
--- a/packages/css/src/transformCss.ts
+++ b/packages/css/src/transformCss.ts
@@ -148,7 +148,7 @@ class Stylesheet {
     if (root.type === 'keyframes') {
       root.rule = Object.fromEntries(
         Object.entries(root.rule).map(([keyframe, rule]) => {
-          return [keyframe, this.transformContent(rule)];
+          return [keyframe, this.transformProperties(rule)];
         }),
       );
       this.keyframesRules.push(root);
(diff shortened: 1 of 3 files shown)


packages/css/src/transformCss.ts
```diff
@@ -220,6 +216,10 @@ class Stylesheet {
     });
   }
 
+  transformProperties(cssRule: CSSPropertiesWithVars) {
```
michaeltaranto, line 219: Should this be `CSSRule` instead?

askoufis: I thought so too, but type checking fails if we use `CSSRule` here. I made it `CSSPropertiesWithVars` as that's the input type for both `transformContent` and `pixelifyProperties`.

mattcompiles (approved):

packages/css/src/transformCss.ts
michaeltaranto: Oh its because we pass it `CSSRule['rule']` from the call sites. I see now.

packages/css/src/transformCss.ts
```diff
@@ -208,10 +206,8 @@ class Stylesheet {
   }
 
   addRule(cssRule: CSSRule) {
-    // Run `pixelifyProperties` before `transformVars` as we don't want to pixelify CSS Vars
-    const rule = this.transformVars(
-      this.transformContent(this.pixelifyProperties(cssRule.rule)),
-    );
+    // Run `normalizeProperties` before `transformVars` as we don't want to pixelify CSS Vars
```
michaeltaranto, line 209:
```suggestion
    // Run `transformProperties` before `transformVars` as we don't want to pixelify CSS Vars
```

packages/css/src/transformCss.ts
```diff
@@ -184,10 +184,8 @@ class Stylesheet {
   }
 
   addConditionalRule(cssRule: CSSRule, conditions: Array<string>) {
-    // Run `pixelifyProperties` before `transformVars` as we don't want to pixelify CSS Vars
-    const rule = this.transformVars(
-      this.transformContent(this.pixelifyProperties(cssRule.rule)),
-    );
+    // Run `normalizeProperties` before `transformVars` as we don't want to pixelify CSS Vars
```
michaeltaranto, line 187:
```suggestion
    // Run `transformProperties` before `transformVars` as we don't want to pixelify CSS Vars
```

michaeltaranto (approved):

PR #996, opened 2023-02-09
Allow shared variables between exports

### LINKED ISSUE #995: Object literal gets duplicated when sprinkles properties are exported
### Description

When the properties passed to `createSprinkles` are exported, the webpack plugin duplicates the object literal with the properties definition (I suspect the same happens with the other plugins).

For example, given the following code:

```ts
export const responsiveProperties = defineProperties({
 // properties, shorthands, conditions, etc
});

export const colorModeProperties = defineProperties({
 // properties, shorthands, conditions, etc
});

export const sprinkles = createSprinkles(
  responsiveProperties,
  colorModeProperties,
);
```

The bundled code will end up looking like:

```js
var responsiveProperties = { /* responsiveProperties definition*/ };
var colorModeProperties = { /* colorModeProperties definition*/ };
var sprinkles = (0,_vanilla_extract_sprinkles_createRuntimeSprinkles__WEBPACK_IMPORTED_MODULE_2__.createSprinkles)({ /* responsiveProperties definition*/ },{ /* colorModeProperties definition*/ });
```

The object literals for `responsiveProperties` and `colorModeProperties` end up duplicated as arguments for `createSprinkles`.

### Steps to reproduce the bug

Check the codesandbox link with the minimal reproduction. You can also reproduce by modifying the `webpack-react` example in the `vanilla-extract` repo to export the properties in the `sprinkles.css.ts` file. In both cases, take a look at `dist/main.js` file to see how the object literals end up duplicated.

### Reproduction

https://codesandbox.io/p/sandbox/hardcore-raman-105ni0?file=%2Fsrc%2Fsprinkles.css.ts&selection=%5B%7B%22endColumn%22%3A1%2C%22endLineNumber%22%3A49%2C%22startColumn%22%3A1%2C%22startLineNumber%22%3A4%7D%5D

### System Info

```shell
System:
    OS: macOS 13.2
    CPU: (8) arm64 Apple M1 Pro
    Memory: 94.94 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.13.1 - ~/.nvm/versions/node/v16.13.1/bin/node
    Yarn: 1.22.17 - ~/.dev/yarn/v1.22.17/bin/yarn
    npm: 8.1.2 - ~/.nvm/versions/node/v16.13.1/bin/npm
    Watchman: 2023.01.30.00 - /opt/homebrew/bin/watchman
  Browsers:
    Firefox: 107.0.1
    Safari: 16.3
  npmPackages:
    @vanilla-extract/babel-plugin: 1.1.7 => 1.1.7
    @vanilla-extract/webpack-plugin: 2.1.11 => 2.1.11
    webpack: 4.46.0 => 4.46.0
```

### Used Package Manager

yarn

### Logs

_No response_

### Validations

mattcompiles:
Fixes #995


packages/integration/src/processVanillaFile.test.ts
```diff
@@ -64,28 +65,104 @@ describe('stringifyExports', () => {
...
+
+    expect(serializeVanillaModule(['import "./styles.css"'], exports, null))
+      .toMatchInlineSnapshot(`
+      "import "./styles.css"
+      import { myFunction as _86bce } from 'my-package';
+      export var sprinkles = _86bce('arg1','arg2');"
+    `);
+  });
+
+  test('should re-use exports in handle serialized function args', () => {
```
ncardeli, line 137: Nice!

ncardeli (approved):
This is a very creative solution to the problem! Thank you for working on this!

Update: tried it on the `webpack-react` example and it's working as expected 👍

packages/integration/src/processVanillaFile.test.ts
```diff
@@ -64,28 +65,104 @@ describe('stringifyExports', () => {
...
+    };
+
+    const sprinkles = () => {};
+    sprinkles.__function_serializer__ = {
+      importPath: 'my-package',
+      importName: 'myFunction',
+      args: [complexExport],
+    };
+    const exports = {
+      default: complexExport,
```
markdalgleish, line 155: I think it'd be good to include non-default exports in this test too.

packages/integration/src/processVanillaFile.test.ts
```diff
@@ -64,28 +65,104 @@ describe('stringifyExports', () => {
...
+      export var otherStyles = 'some-style other-style';"
+    `);
+  });
+
+  test('should re-use exports if possible', () => {
+    const a = {
+      style: 'the-style',
+    };
+    const exports = {
+      a,
```
markdalgleish, line 104: I think it'd be good to include a re-used default export too.

markdalgleish (approved):
This is awesome. I've added a few minor comments, but otherwise lgtm!

PR #1003, opened 2023-02-18
Add docs for `addFunctionSerializer`

markdalgleish:
I'm excited to finally get this documented! Keen to get feedback on it though.

I'm not sure where this should go from an IA perspective, so I've added it to the end of "API" for now. I originally added it further down the nav under an "Advanced API" section, but then I felt like that was a bit too hidden, so I don't know 🤷‍♂️

### DIFF
--- a/site/docs/api/add-function-serializer.md
+++ b/site/docs/api/add-function-serializer.md
@@ -56,7 +56,7 @@ export function styledDiv(styles) {
 }
 ```
 
-This is a pretty simple implementation, but it has a couple of major problems that prevent it from working. Firstly, there's no way to convert this function to a string so it can be added to your runtime JavaScript bundle. Secondly, even if it could be serialized somehow, it relies on `@vanilla-extract/css` which can't be used in the browser.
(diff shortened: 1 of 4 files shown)


site/docs/api/add-function-serializer.md
```diff
@@ -0,0 +1,122 @@
...
+> ✋&nbsp;&nbsp;All arguments passed to the runtime function must be serializable!
+
+```ts
+// index.css.ts
+
+import { addFunctionSerializer } from '@vanilla-extract/css/functionSerializer';
+import { runtimeStyledDiv } from './runtime';
+
+export function styledDiv(styles) {
+  const className = style(styles);
```
graup, line 103: How about adding `import { style } from '@vanilla-extract/css';` ?

site/docs/api/add-function-serializer.md
```diff
@@ -0,0 +1,122 @@
...
+        className={[props.className, className]
+          .filter(Boolean)
+          .join(' ')}
+      />
+    );
+  };
+}
+```
+
+This is a pretty simple implementation, but it has a couple of major problems that prevent it from working. Firstly, there's no way to convert this function to a string so it can be added to your runtime JavaScript bundle. Secondly, even if it could be serialized somehow, it relies on `@vanilla-extract/css` which can't be used in the browser.
```
mattcompiles, line 59: As discussed, could we re-word this a bit as technically `@vanilla-extract/css` can be used in a browser?

mattcompiles (approved):

PR #1016, opened 2023-02-27
site: Move ids to headings

michaeltaranto:
Hoping this supports deep linking in the upcoming search PR.

### DIFF
--- a/site/src/mdx-components.tsx
+++ b/site/src/mdx-components.tsx
@@ -25,7 +25,8 @@ interface HeadingProps {
   children: ReactNode;
   component: ElementType;
   level: HeadingLevel;
-  id: string;
+  href?: string;
+  id?: string;
 }
 
 const Block = ({
@@ -106,7 +107,7 @@ const A = ({
   );
 };
 
-const Heading = ({ level, component, children, id }: HeadingProps) => {
+const Heading = ({ level, component, children, href }: HeadingProps) => {
   const headingElement = createElement(
     Box,
     {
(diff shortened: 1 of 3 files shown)


site/src/mdx-components.tsx
```diff
@@ -116,21 +117,17 @@ const Heading = ({ level, component, children, id }: HeadingProps) => {
...
 
-  return id ? (
-    <>
-      <Box
-        component="a"
-        display="block"
-        id={id}
-        style={{
-          visibility: 'hidden',
-          scrollMarginTop: 120,
```
askoufis, line 127: Are we ok with removing the scroll margin? It was kind of nice having heading anchors consistently scroll to the same spot.

michaeltaranto: Unless my Chrome is busted, I cannot see this having an effect with or without it. Are you able to replicate this as a regression?

askoufis: In both firefox and chrome, navigating between `h2`s and `h3`s in the aside makes the different scroll margins obvious. Ultimately it's because `h3`s have `xlarge` top padding but `h2`s have `xxlarge` top padding (and `h1s` have none).

michaeltaranto: Ah need to hoist that to the element with the id. Also aligning it to the header height as defined in the docs page style sheet.

askoufis: Hmm, your change didn't fix it 🤔 
Also, the anchor on the `h1` is gone now?

askoufis (approved):

PR #1018, opened 2023-02-28
Add esbuild as peer dependency of esbuild plugin

gabro:
The plugin imports the `Plugin` type from the `esbuild` module, so I *think* it is better to declare it as a dependency of some sort. 

The issue I've seen without adding `esbuild` as a dependency is that the `Plugin` import is a moving target since it will resolve to an apparently random version of `esbuild` depending on the user's setup.

For example, I've started getting some errors in my build when I updated `tsup`, which uses `esbuild` internally and which recently upgraded to 0.17.0. The issue ultimately boils down to Vanilla Extract's plugin arbitrarily resolving the `Plugin` type from version 0.14.0 which was transitively present in the dependency tree and which has an incompatible type wrt 0.17.0.

My hope is that using a peer dependency will at least surface potential incompatibilities errors in a more predictable manner.

---

As an aside, I've randomly sampled esbuild plugins from https://github.com/esbuild/community-plugins to see what they do, and I don't think I've found one that does the same as the others 😅. Some use `"esbuild": "*"` as a peer dep, some don't specify anything, some add esbuild as a dependency, some use `^`, some use `>=`, and so on.

I've ultimately opted for a peer dependency, since it feels more natural for a plugin (it's fair to assume the consumer of this will have `esbuild` as a dependency on their end)

### DIFF
--- a/packages/esbuild-plugin/package.json
+++ b/packages/esbuild-plugin/package.json
@@ -19,5 +19,8 @@
   },
   "devDependencies": {
     "esbuild": "^0.16.3"
+  },
+  "peerDependencies": {
+    "esbuild": ">=0.16.0"
   }
 }


.changeset/tall-hotels-speak.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/esbuild-plugin': minor
```
gabro, line 2: Not sure this should be minor or major, actually 🤔

MichaelDeBoey: This would be a breaking change, so it should be `major` imo

MichaelDeBoey: This would be a breaking change, so it should be `major` imo

```suggestion
'@vanilla-extract/esbuild-plugin': major
```

MichaelDeBoey (approved):
I think we should also do this for `@vanilla-extract/integration`
@markdalgleish WDYT?

If we do this in the integration package as well, this will help us in the Remix codebase to not have 2 different versions for the reasons mentioned above

AndrewLeedham (changes requested):
@gabro Seems like the esbuild types are 'baked in' to the VE .dts bundle. So, I think the rollup bundling process would also need to treat `esbuild` as external. See https://unpkg.com/browse/@vanilla-extract/esbuild-plugin@2.2.2/dist/vanilla-extract-esbuild-plugin.cjs.d.ts

MichaelDeBoey:
@AndrewLeedham and all other people interested in getting this merged: I created #1171 from latest `master` and made sure all other packages that have `@vanilla-extract/integration` in their `dependencies` aren't affected at all (they now have `esbuild` as their own dependency)

gabro:
@MichaelDeBoey thanks for picking this up! I haven't had any time to come back to it. Closing in favor of #1171

PR #1021, opened 2023-03-03
Migrate stylesheet snapshot tests to Playwright

markdalgleish:
The CSS snapshots taken in Jest are now handled during the Playwright test run. This is a nice tidy up, but it also was necessary to get our upcoming new compiler work to pass since it had issues running in Jest, and migrating to Vitest caused other issues.

The Jest snapshots were taken in production mode while the Playwright tests were only in development mode, so another nice uplift here is that we're running the screenshot tests across both prod and dev modes!

Also note that since the existing Playwright tests are no longer just for screenshots anymore, I've renamed the `screenshots` directory to `e2e`.


mattcompiles (approved):
This is just delightful. ❤️ 🍻

Issue #1013, opened 2023-02-24
Webpack build error on Next.js 13.2


re-taro:
### Describe the bug

Error with internal webpack configuration at build time when using `@vanilla-extract/next-plugin` in Next.js v13.2 delivered yesterday.

### Reproduction

https://github.com/re-taro-dev/card.re-taro.dev

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (12) x64 Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
    Memory: 3.98 GB / 15.79 GB
  Binaries:
    Node: 18.14.2 - ~\scoop\apps\nodejs-lts\current\node.EXE
    Yarn: 1.22.19 - ~\scoop\apps\yarn\current\bin\yarn.CMD
    npm: 9.5.0 - ~\scoop\apps\nodejs-lts\current\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1265.0), Chromium (110.0.1587.46)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vanilla-extract/css: 1.9.5 => 1.9.5
    @vanilla-extract/next-plugin: 2.1.1 => 2.1.1
    webpack: 5.75.0 => 5.75.0
```

### Used Package Manager

pnpm

### Logs

```shell
Failed to compile.

[...]

Issue #929, opened 2022-11-28
Unable to render when using with NextJS v13 & React Server Components


nick-kang:
### Describe the bug

App does not render when I import  a `.css.ts` file in a React Server Component. See [component](https://github.com/nick-kang/vanilla-extract-repro-001/blob/main/app/page.tsx) in repro repo.

Workaround is to convert the `.css.ts` to a default export and use [swc-plugin-vanilla-extract](https://github.com/kwonoj/swc-plugin-vanilla-extract) instead of `@vanilla-extract/next-plugin`

```
// page.css.ts before
export const container = style({
  background: 'blue'
})
```

```
// page.css.ts workaround
const container = style({
  background: 'blue'
})

export default {
  container
}
```

### Reproduction

https://github.com/nick-kang/vanilla-extract-repro-001

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
    CPU: (4) x64 Intel(R) Core(TM) i5-6500 CPU @ 3.20GHz
    Memory: 29.65 GB / 46.75 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 18.7.0 - /run/user/1000/fnm_multishells/553958_1669586582142/bin/node
    Yarn: 3.2.2 - /run/user/1000/fnm_multishells/553958_1669586582142/bin/yarn
    npm: 8.15.0 - /run/user/1000/fnm_multishells/553958_1669586582142/bin/npm
    Watchman: 2022.07.04.00 - /home/linuxbrew/.linuxbrew/bin/watchman
  Browsers:
    Chrome: 107.0.5304.121
    Firefox: 107.0
  npmPackages:
    @vanilla-extract/css: ^1.9.2 => 1.9.2 
    @vanilla-extract/next-plugin: ^2.1.1 => 2.1.1 
    @vanilla-extract/webpack-plugin: ^2.2.0 => 2.2.0
```

### Used Package Manager

pnpm

### Logs

```shell
TypeError: Cannot read properties of undefined (reading 'chunks')
    at getCssInlinedLinkTags (/node_modules/.pnpm/next@13.0.5_biqbaboplfbrettd7655fr4n2y/node_modules/next/dist/server/app-render.js:362:49)
[...]

PR #1022, opened 2023-03-03
next-plugin: prepare for app dir support

graup:
This by itself doesn't solve next 13 app dir compatibility, but it seems this will be one of the necessary steps.

See
https://github.com/vanilla-extract-css/vanilla-extract/issues/1013#issuecomment-1446302154
https://github.com/vanilla-extract-css/vanilla-extract/issues/929#issuecomment-1338310803

Leaving this here until we resolved the other open questions and ran some more tests.


michaeltaranto (approved):
If we confirm this doesn't break previous versions due to config validation then we should be good to go on this.

.changeset/few-snails-yell.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/next-plugin': minor
```
mattcompiles, line 2:
```suggestion
'@vanilla-extract/next-plugin': patch
```

mattcompiles:
Closing in favour of #1060

PR #1024, opened 2023-03-04
Don’t throw when failing to add debug objects to an empty file

huw:
When adding debug objects to an empty file (in `integration`), we currently bail whenever `result.code` is falsy (`!result.code`). However, a completely empty file will transform to falsy (since it will be the empty string `””`) even though it’s valid (I think; although this is probably a matter of opinion for the maintainers) and shouldn’t have any debug objects added to it.

This PR makes the simple change of checking to see if `result.code === null`, [which is what Babel will output if the transformation fails](https://github.com/babel/babel/blob/3b836c42dbb54dc2c0dffc2a17dfb3cd1490e3de/packages/babel-core/src/transformation/index.ts#L73). I decided to make this a patch change since while it does technically affect the API contract it’s more of a bugfix.

Give me a shout if I’ve gotten this wrong—I found the error while playing around with Remix’s integration and didn’t go to deep on why we might make the falsy check ^_^


mattcompiles (approved):

packages/integration/src/transform.ts
```diff
@@ -30,7 +30,7 @@ export const transformSync = ({
       configFile: false,
     });
 
-    if (!result || !result.code) {
+    if (!result || result.code === null) {
```
mattcompiles, line 33: Looks like it can be undefined as well. 

```suggestion
    if (!result || result.code == null) {
```

packages/integration/src/transform.ts
```diff
@@ -62,7 +62,7 @@ export const transform = async ({
...
```
mattcompiles, line 65: Looks like it can be undefined as well. 

```suggestion
    if (!result || result.code == null) {
```

huw:
Accepted both but worth noting that Babel [seems to explicitly protect against this in the source](https://github.com/babel/babel/blob/3b836c42dbb54dc2c0dffc2a17dfb3cd1490e3de/packages/babel-core/src/transformation/index.ts#L73). I’m probably missing something though—I only took a cursory look.

mattcompiles:
@huw you're probably right, it's just the types. No harm guarding against it as well.

huw:
Exactly. Thank you!

PR #1030, opened 2023-03-08
Add `createCompiler` to integration package

markdalgleish:
This PR adds a new compiler to the integration package, powered by [vite-node](https://www.npmjs.com/package/vite-node).

This allows us to cache results across builds which can dramatically improve build performance, particularly for esbuild consumers since it doesn't automatically manage caching of plugins.

I've also added a new esbuild plugin that makes use of the new compiler, but it's been marked as private for now. This allows us to run tests and fixtures against the new plugin without forcing us to release it immediately. In the meantime, since Remix uses `@vanilla-extract/integration` directly, Remix will be able to roll out the new Vanilla Extract compiler immediately without waiting for all official plugins to be migrated.

You can run a fixture against the new compiler like this:

```
pnpm build
pnpm start-fixture themed --type=esbuild-next
```


packages/integration/src/compiler.ts
```diff
@@ -0,0 +1,376 @@
...
+
+      const cssByFileScope = new Map<string, Array<Css>>();
+      const localClassNames = new Set<string>();
+      const composedClassLists: Array<Composition> = [];
+
+      const cssAdapter: Adapter = {
+        getIdentOption: () => identifiers,
+        onBeginFileScope: (fileScope) => {
+          // Before evaluating a file, reset the cache for it
+          cssByFileScope.set(fileScope.filePath, []);
```
mattcompiles, line 224: We should scope the cache to include packageName from the filescope

packages/integration/src/compiler.ts
```diff
@@ -0,0 +1,376 @@
...
+      cssDeps.add(relativePath);
+    }
+
+    if (moduleNode.file) {
+      watchFiles.add(moduleNode.file);
+    }
+
+    for (const importedModule of moduleNode.importedModules) {
+      queue.push(importedModule);
+    }
```
mattcompiles, line 37:
```suggestion
    queue.push(...moduleNode.importedModules);
```

mattcompiles (approved):
🎉 🚀 🍻

PR #1062, opened 2023-04-13
sprinkles: Add support for layers

markdalgleish:
This PR adds an optional `@layer` property to the options passed to `defineProperties`.

Example usage:

```ts
// sprinkles.css.ts
import { defineProperties } from '@vanilla-extract/sprinkles';
import { layer } from '@vanilla-extract/css';

export const sprinklesLayer = layer();

const properties = defineProperties({
  '@layer': sprinklesLayer,
  // etc.
});
```

### DIFF
--- a/.changeset/sprinkles-layers.md
+++ b/.changeset/sprinkles-layers.md
@@ -1,0 +1,20 @@
+---
+'@vanilla-extract/sprinkles': minor
+---
+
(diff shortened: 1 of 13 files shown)


askoufis (approved):

mattcompiles (approved):
Nice 🚀

PR #1081, opened 2023-05-02
Allow function selectors

aspirisen:
Accept function in selectors prop to allow circle css selectors references

More detailed info https://github.com/vanilla-extract-css/vanilla-extract/discussions/1077

### DIFF
--- a/.changeset/famous-coats-cough.md
+++ b/.changeset/famous-coats-cough.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Accept function in selectors props so you could resolve circle selectors references issue
--- a/packages/css/src/transformCss.test.ts
+++ b/packages/css/src/transformCss.test.ts
@@ -971,6 +971,48 @@ describe('transformCss', () => {
     `);
   });
 
+  it('should handle function selectors', () => {
+    expect(
+      transformCss({
+        composedClassLists: [],
+        localClassNames: ['testClass', 'parentClass'],
+        cssObjs: [
+          {
+            type: 'local',
+            selector: 'testClass',
+            rule: {
+              color: 'red',
+              selectors: () => ({
+                '&:nth-child(3)': {
+                  color: 'blue',
+                },
+                'parentClass > div > span ~ &.someGlobalClass:hover': {
+                  background: 'green',
+                },
+                'parentClass&': {
+                  background: 'black',
+                },
+              }),
+            },
+          },
+        ],
+      }).join('\n'),
+    ).toMatchInlineSnapshot(`
+      .testClass {
+        color: red;
+      }
+      .testClass:nth-child(3) {
+        color: blue;
+      }
+      .parentClass > div > span ~ .testClass.someGlobalClass:hover {
+        background: green;
+      }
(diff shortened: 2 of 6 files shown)


mattcompiles:
Hey @aspirisen 

Thanks for the PR. 

Unfortunately, because this relies on an implementation detail, the selector functions running after the identifiers for the whole file are created, we can't accept this PR. @graup suggested a reasonable workaround in the discussion and I'd also suggest avoiding selectors with circular relationships at all if possible.

graup:
Your solution happens to work only due to the way the VE compiler is run, but it's not what someone would expect from JS normally. In JS you cannot reference const/let variables before they are defined, no matter if you're in a function or not. We'd rather not rely on this "implementation quirk".

aspirisen:
> Unfortunately, because this relies on an implementation detail

I am not sure that functional selectors somehow hack VE internals, for example this is a valid js code
```
function test(){
    console.log(data)
}

const data = 5;

test()
```

It is natural behavior of JS, nothing special to VE

The same approach styled-components uses, for [example here](https://codesandbox.io/s/styled-components-base-forked-u0mwi6?file=/App.js)

```
export const Child = styled.div`
  ${() => css`
    ${Parent} & {
      color: red;
    }
  `}
`;

export const Parent = styled.div`
  &:has(${Child}) {
    padding: 30px;
  }
`;
```

Since VE doesn't allow child selectors you can't use the same approach as preprocessors allow, hence there can be situations where you have to have circle references
```
.child {}

.parent {
    & .child {
        color: red;
    }
    
    &:has(.child) {
        padding: 10px;
    }
}
```

Especially it will be useful if you are working on something complex like ui libraries where you can have complex selectors.

The approach [from discussion](https://github.com/vanilla-extract-css/vanilla-extract/discussions/1077#discussioncomment-5843959) works, but it makes code more complex and harder to follow connections for no reason

mattcompiles:
Hey @aspirisen 

I hear your points. My concerns were originally around the timing of when the CSS is generated. If you generated CSS each style at a time then this approach wouldn't work as the referenced selector wouldn't exist yet. While that's not how VE works today I didn't want to couple this feature to an internal implementation detail. 

I'm going to re-open the PR and raise this for discussion with the rest of the team to get their thoughts.

graup:
Hi @aspirisen, it just occurred to me that you can also do this using [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get):

```ts

export const child = style({
  background: 'blue',
  get selectors() {
    return {
      [`${parent} &`]: {
        color: 'red',
      },
    };
  },
});

export const parent = style({
  background: 'yellow',
  selectors: {
    [`&:has(${child})`]: {
      padding: '10px',
    },
  },
});
```

What do you think? At this point it's just a stylistic preference for arrow functions (which you cannot do with getters), isn't it?

aspirisen:
@graup looks interesting, for now I do not see disadvantages of this approach 👍
Maybe it is worth to mention it in docs? I can update this pr with mentioning getters approach in docs and add tests just to make sure it works.

graup:
That would be great!

Issue #1076, opened 2023-04-27
rollup watch mode doesn't work


jeetiss:
### Describe the bug

- setup `@vanilla-extract/rollup-plugin` 
- run watch mode `rollup -c -w`
- trigger rebuild
- plugin imports `'styles.css.ts.vanilla.css'` instead of generated files
- `Module not found: Can't resolve 'styles.css.ts.vanilla.css'`

### Reproduction

https://stackblitz.com/edit/node-r8p1k1?file=dist%2Findex.js

### System Info

```shell
System:
  OS: macOS 13.2.1
  CPU: (8) arm64 Apple M1 Pro
  Shell: 5.8.1 
Binaries:
  Node: 18.12.0
  Yarn: 1.22.17
  npm: 8.19.3
Browsers:
  Chrome: 112.0.5615.137
  Firefox: 111.0
  Safari: 16.3
  Safari Technology Preview: 16.4
```

### Used Package Manager

npm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

graup:
Hi! I tried your repro but it seems to work fine for me. (Run `yarn watch`, then edit the css file. Dist is updated properly.) Am I missing something?

graup:
Did some [more testing](https://stackblitz.com/edit/node-pikheo?file=style.css.js,Button.js,Button.style.css.js,index.js,rollup.config.js,dist%2Findex.js) with two .css.js files and it looks like whenever you save one of them, that one is properly imported, but the import of the other one breaks.

My initial guess is that when rollup rebuild is initiated, we clear all the cached asset import paths (generated according to assetNames), but then only the changed file is regenerated, making other assets point to file names instead of proper import paths.

https://user-images.githubusercontent.com/898549/236717798-f65b3be9-e53a-4098-bb8f-2a1dd1e46fc0.mov

jeetiss:
@graup thanks for the fast fix! Could you please ping me when it would be released

graup:
> @graup thanks for the fast fix! Could you please ping me when it would be released

This was released in @vanilla-extract/rollup-plugin@1.2.2 !

PR #693, opened 2022-05-22
rollup: clear emitted files cache for watch mode

graup:
Fixes #691

The issue was that `emittedFiles` still held references to the assets from the previous build. Those are accessed down below at `if (!emittedFiles.get(fileName))`. However, assets get cleared on a re-build, so we need to re-emit those files when we're in watch mode.

I wish I could write a test for this but I have no idea how to simulate rollup watch mode and file changes :/

Issue #691, opened 2022-05-17
Rollup plugin throws an error when in watch mode


roginfarrer:
### Describe the bug
`@vanilla-extract/rollup-plugin` throws the following error when in watch mode:

```
[!] (plugin vanilla-extract) Error: Plugin error - Unable to get file name for unknown file "3f979318".
Error: Plugin error - Unable to get file name for unknown file "3f979318".
    at error (/Users/rfarrer/projects/vanilla-extract-rollup-example/node_modules/rollup/dist/shared/rollup.js:198:30)
    at Object.FileEmitter.getFileName (/Users/rfarrer/projects/vanilla-extract-rollup-example/node_modules/rollup/dist/shared/rollup.js:15589:24)
    at /Users/rfarrer/projects/vanilla-extract-rollup-example/node_modules/@vanilla-extract/rollup-plugin/dist/vanilla-extract-rollup-plugin.cjs.dev.js:82:32
    at Array.reduce (<anonymous>)
    at Object.renderChunk (/Users/rfarrer/projects/vanilla-extract-rollup-example/node_modules/@vanilla-extract/rollup-plugin/dist/vanilla-extract-rollup-plugin.cjs.dev.js:80:31)
    at /Users/rfarrer/projects/vanilla-extract-rollup-example/node_modules/rollup/dist/shared/rollup.js:22834:37
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### Link to reproduction
1. Clone this repo https://github.com/graup/vanilla-extract-rollup-example
2. cd into packages/ui
3. Run `yarn rollup -c --watch`
4. Make a change to a `.css.ts` file and save

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

```node
System:
    OS: macOS 12.3
    CPU: (8) arm64 Apple M1
    Memory: 398.45 MB / 16.00 GB
    Shell: 3.4.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 18.0.0 - ~/Library/Caches/fnm_multishells/26791_1652796954290/bin/node
    Yarn: 1.22.18 - /opt/homebrew/bin/yarn
[...]

PR #1084, opened 2023-05-09
rollup-plugin: fix caching in watch mode

graup:
I have changed the asset emitting logic significantly.

Before, we emitted css assets in the `resolveId` stage and kept track of emitted assets in a plugin-local variable (`emittedFiles`), clearing them all on a rebuild. However, emitted css assets are external and not part of the module graph and thus not cached by rollup. Previously emitted assets are cleared on rebuilds and we didn't re-emit them (because on rebuild only the changed files get transformed/resolved).

I changed this to use the rollup caching mechanism by attaching the css output to the module's `meta` during `resolveId` and only emitting the actual css assets in the `moduleParsed` stage. Module info is cached between rebuilds, so this way we can access previous build outputs.

Fixes: #1076
(Also a better fix for #691. What I did in #693 only works in some cases.)

### DIFF
--- a/.changeset/healthy-turtles-help.md
+++ b/.changeset/healthy-turtles-help.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/rollup-plugin': patch
+---
+
+Fix emitting assets when in watch mode (#1076)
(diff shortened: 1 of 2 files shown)


mattcompiles (approved):
Makes sense mate. Also much cleaner to store the state in the Rollup data structures 👏

PR #1093, opened 2023-05-16
fix(vite-plugin): Fix HMR in Windows

kosmotema:
According [to Vite docs](https://vitejs.dev/guide/api-plugin.html#path-normalization), resolved ids will have `/` as directory separators.
But `watchFiles`, returned by the `compile` function, have platform-specific directory separators (`\` in Windows), so comparing file and module id (`file === validId`) will always be false in Windows and module itself will be added to watched files (which breaks HMR and leads to page reload).

So, by normalizing paths from `watchFiles` (using POSIX directory separators instead of platform-specific), module will be correctly ignored from adding to watched files and HMR will work.

### DIFF
--- a/.changeset/pink-students-grab.md
+++ b/.changeset/pink-students-grab.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Fix HMR in Windows by normalizing paths from `watchFiles`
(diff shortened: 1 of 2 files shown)


mattcompiles:
@kosmotema Thanks for the PR. Looks good, just checking if you've manually validated that this fixes your issue? I don't currently have a Window dev environment to validate it.

askoufis:
Reproduced this on Windows. Can confirm that the fix works. Wonder if maybe we should normalize the file paths inside `compile` instead? It would add some redundant work to all other integrations, but at least it would ensure that all integrations are working with the same file paths.

mattcompiles:
@askoufis I think we can merge this and get it fixed. Happy to discuss moving it internally separately. Maybe some plugins don't want POSIX on windows too? Unlikely though.

mattcompiles (approved):

Issue #944, opened 2022-12-10
selectors using a recipe with variants generates incorrect css


PHILLIPS71:
### Describe the bug

I'm in the process of converting an existing stitches based design system over to use vanilla-extract, but have ran into a issue on where I cannot reference child classes inside of a recipe variant. Instead it wants me to reference the parent inside the child after reading this error that I receive at build time.

> Style selectors must target the '&' character (along with any modifiers), e.g. `${parent} &` or `${parent} &:hover`.

> This is to ensure that each style block only affects the styling of a single class.

> If your selector is targeting another class, you should move it to the style definition for that class, e.g. given we have styles for 'parent' and 'child' elements, instead of adding a selector of `& ${child}`) to 'parent', you should add `${parent} &` to 'child').

So that's exactly what I've done in this case, the code below is the changes that I've made where that navbar recipe is the parent element.

```tsx
export const container = style({
  ...

  selectors: {
    [`${navbar({ kind: 'floating' })} &`]: {
      marginTop: vars.spaces[6],
      marginRight: vars.spaces[6],
      marginLeft: vars.spaces[6],
      borderRadius: vars.radii.md,
      borderBottom: 'none'
    },
  },
})
```

Although when the navbar has the kind floating variant applied the styles in the selector are not being set, after some digging I've realised that this is due to the styles are the getting generated seem to be incorrect. Below are the styles that have been generated:

```css
.Navbar_navbar__xhan6d0 .Navbar_navbar_kind_floating__xhan6d3 .Navbar_container__xhan6d4 {
    margin-top: var(--spaces-6__zxivqp3j);
    margin-right: var(--spaces-6__zxivqp3j);
    margin-left: var(--spaces-6__zxivqp3j);
    border-radius: var(--radii-md__zxivqp4d);
    border-bottom: none
}
```

Shouldn't the class names be defined as `.Navbar_navbar__xhan6d0.Navbar_navbar_kind_floating__xhan6d3 .Navbar_container__xhan6d4`, removing the space between the navbar variants classes as these are all class names that are appended into the navbar element?

### Reproduction

https://github.com/PHILLIPS71/vanilla-extract-issue

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (8) x64 Intel(R) Core(TM) i7-9700K CPU @ 3.60GHz
    Memory: 39.64 GB / 63.91 GB
  Binaries:
[...]

Issue #868, opened 2022-10-05
Using a recipe with a global style creates incorrect selector chain. 


caranicas:
### Describe the bug

When using a recipe with a global style the css that gets created expects the variant class to be child and not a chained selector. 

I have a codesandbox demo linked to to provide the full context, but  this is the long and short of it. 

creating a style like this 
```
globalStyle(`${exampleRecipe({ orientation: "vertical" })} > div`, {
  backgroundColor: "red"
});
```

and DOM like this
```
 <div class="${exampleRecipe({
    orientation: "vertical"
  })}">
      <div>Hello</div>
 </div>
```
results in  DOM like this
```
<div class="styles__14w0md60 styles_orientation_vertical__14w0md61">
      <div>Hello</div>
</div>
```
BUT css like this 
```
.styles__14w0md60 .styles_orientation_vertical__14w0md61 > div {
  background-color: red;
}
```
there is a space between `.styles__14w0md60` and ` .styles_orientation_vertical__14w0md61` 

which means my div is not red like I would expect. the two class selectors should be chained to properly target the dom node. 
````
.styles__14w0md60.styles_orientation_vertical__14w0md61 > div {
  background-color: red;
}
````

### Reproduction

https://codesandbox.io/s/vanilla-extract-recipe-bug-xpnu2j?file=/src/index.js

### System Info

```shell
System:
    OS: Windows 10 10.0.19044
    CPU: (24) x64 AMD Ryzen 9 3900X 12-Core Processor
    Memory: 4.73 GB / 15.95 GB
  Binaries:
    Node: 16.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 8.15.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 105.0.5195.127
    Edge: Spartan (44.19041.1266.0), Chromium (106.0.1370.34)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @vanilla-extract/css: ^1.9.0 => 1.9.0
    @vanilla-extract/css-utils: ^0.1.2 => 0.1.2
    @vanilla-extract/recipes: ^0.2.5 => 0.2.5
    @vanilla-extract/sprinkles: ^1.5.0 => 1.5.0
    @vanilla-extract/vite-plugin: ^3.5.0 => 3.5.0
    vite: ^3.0.7 => 3.1.0
```

### Used Package Manager

npm

### Validations

[...]

PR #1104, opened 2023-05-29
Expose recipe's class names

aspirisen:
Fixes: https://github.com/vanilla-extract-css/vanilla-extract/issues/944 https://github.com/vanilla-extract-css/vanilla-extract/issues/868

Hello, here I tried to expose class names from `recipe` to allow referencing from other styles.
Currently it adds base and variants class names to recipe runtime function:
```
import { recipe } from '@vanilla-extract/recipes';

const basic = recipe({
  base: {},
  variants: {
    rounded: {
      true: { borderRadius: 999 },
    },
  },
});

export const cssWithRecipe = style({
  color: 'red',
  selectors: {
    [`.${basic.base} &`]: {
      color: 'blue',
    },
    [`.${basic.rounded.true} &`]: {
      color: 'yellow',
    },
  },
});
```
Note: If there is no `base` inside `basic` then `basic.base` will be empty string. Probably we need to hide the `base` prop at least on type level if `base` is not defined.
We could also put the class names not into the function itself, but in special prop (like currently we have variants function), but then the selector could be potentially too long what is not so readable inside object brackets

Firstly I tried to make recipe runtime function to provide correct selector if the runtime function is called inside vanilla file like was expected in the original issue. I tried to add special env property `process.env.RUNNING_INSIDE_VANILA` and then use the env variable to decide if recipe's runtime function should create css selector or class name. However, there was a problem that bundlers could shim `process` and thus when the code reaches eval stage the global `process` object was not available.

We can also create another function inside recipe runtime function to create css selectors, something like `basic.select({ rounded: true })`

P.S. I will add docs and changesets when we will decide on final solution


aspirisen:
Personally I do not like this approach, I would prefer `[exampleRecipe({ orientation: "vertical" }) &]` to just work in vanilla files. But in this case we need to detect if runtime function is executing in VE eval or just in normal files. As I mentioned in the description we can't use `process`, but maybe we could use global variable.

Here https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/integration/src/processVanillaFile.ts#L89 we could add special variable:
```
  const adapterBoundSource = `
    var RUNNING_IN_VANILLA_EXTRACT = true;
    require('@vanilla-extract/css/adapter').setAdapter(__adapter__);
    ${source}
  `;
```
By that VE files could detect if they are evaluating in VE environment.

However, there will be problem with normal files runtime since there `RUNNING_IN_VANILLA_EXTRACT` will be not defined. To avoid this problem libraries and VE code could check if the variable is defined `typeof RUNNING_IN_VANILLA_EXTRACT === 'boolean'`. By that we could fix normal files runtime and this approach will work, but there is a disadvantage - code under `RUNNING_IN_VANILLA_EXTRACT` condition will not be removed in production build. 

I tried to use [DefinePlugin](https://webpack.js.org/plugins/define-plugin/) to set that `RUNNING_IN_VANILLA_EXTRACT = false`, but this plugins executes before loaders and thus in eval stage we have code without the `RUNNING_IN_VANILLA_EXTRACT` variable. Maybe there is a way where VE could eval the code and then in loader make variable replacement on its own.

graup:
I'm not sure `recipe.base` will work. That can be an array of classes too, in which case we're back to the original problem. Also like you said it can be empty, which might confuse people. Maybe we can throw a warning if someone uses the non-existing `base` as a string?

Btw, what output would you expect for ```[`${exampleRecipe({ orientation: "vertical" })} .child`]```? Concat all the recipe classes? (`.recipe-base1.recipe-base2.recipe-variant1.recipe-variant2 .child` ?)

aspirisen:
> Edit: we need to think about what our goal is. Is it to make multiple-classnames usable in sub selectors? Is it just to expose the underlying classnames (bases and variants) – with the caveat that those might still break in selectors?

@graup I think the main purpose is to allow recipes in selectors

> Btw, what output would you expect for [`${exampleRecipe({ orientation: "vertical" })} .child`]? Concat all the recipe classes? (.recipe-base1.recipe-base2.recipe-variant1.recipe-variant2 .child ?)

This is a good question, currently I just use `exampleRecipe({ orientation: "vertical" }).split(' ').join('')`. It works but has a lot of unnecessary classes from compound and base. I would say that recipes selectors should return just recipe own classes, without additional base and compound - just to reduce the size of output css file. I do not see where you will need i.e. special compound class name because they are always go together with variant and base class names. However, there is a problem that recipe without base `{}` has no "root" class at all. So if your recipe has only variants you can't select just recipe's root class.

My suggestion is the following:
1. Always generate recipe base class name, even if there is no `base` prop defined
2. `exampleRecipe()` or `exampleRecipe({})` return just own base class name
3. `exampleRecipe({ orientation: "vertical" })` returns own base class name plus the variant class name
4. Do not include compound and mixed base class names in selector mode

graup:
Hi @aspirisen, I had a chat about this with the other maintainers.

1. We need to ensure that for every recipe, unique base and variants classes are generated even if there are no styles in them. Both for the case when there's no base class and when there's composition being used. (This shouldn't be too difficult. There's a weirdness in the code now though [that defaults `base` to the empty string](https://github.com/vanilla-extract-css/vanilla-extract/blob/001be8338a869f41acf19091707a2e097fd80de3/packages/recipes/src/index.ts#L36), that might need a fix.)
2. The class names should be under a separate property, otherwise you get conflicts with the function prototype (e.g. you wouldn't be able to have a variant called "call"). Putting it somewhere like `classNames` also makes clear that those are not the composed styles but generated base class names.
3. Example API that I think could work:

```ts
// .css.ts
export const tableRecipe = recipe({
  base: [
    'base-one'
    'base-two'
  ],
  variants: {
    variant: {
      simple: [
        'simple-one',
        'simple-two'
      ]
    },
  },
});

globalStyle(`${tableRecipe.classNames.base} th`, {
  color: 'green',
});

globalStyle(`${tableRecipe.classNames.variants.variant.simple} th`, {
 color: 'red'
});

// .tsx
<div className={tableRecipe({})} />
<div className={tableRecipe({ variant: 'simple' })} />

// .css
.tableRecipe-base th {
  color: green;
}
.tableRecipe-variants-simple th {
  color: red;
}

// DOM
<div class="tableRecipe-base base-one base-two" />
<div class="tableRecipe-base base-one base-two tableRecipe-variants-simple simple-one simple-two" />
```

aspirisen:
Hey @graup looks good.
I will update this PR in coming days

packages/recipes/src/index.ts
```diff
@@ -63,7 +69,12 @@ export function recipe<Variants extends VariantGroups>(
 
   const config: PatternResult<Variants> = {
     defaultClassName,
+    rawDefaultClassName: defaultClassName.split(' ')[0],
```
aspirisen, line 72: When you have an array of "string" class names (i.e. `base: ['one', 'two']`) the `style()` function computes the final class names internally and concats all of them together. I make an assumption that the first one is the generated one

tests/compiler/fixtures/recipes/recipeClassNames.css.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+  },
+  variants: {
+    first: {
+      true: {
+        selectors: {
+          [`.${basic.classNames.variants.color.red} &`]: {
+            color: 'black',
+          },
+          [`.${basic.classNames.variants.spaceWithDefault.large}.${basic.classNames.variants.rounded.true} &`]:
```
aspirisen, line 45: I provide raw class name, without `.`. I feel that there can be confusion since both of the cases are called "class name". Also, maybe `.` separated selectors will be easier to read.

packages/recipes/src/index.ts
```diff
@@ -63,7 +69,12 @@ export function recipe<Variants extends VariantGroups>(
...
     variantClassNames,
+    // @ts-expect-error
+    rawVariantClassNames: mapValues(variantClassNames, (classNames) =>
```
mattcompiles, line 75: Could we do this work at runtime instead in `createRuntimeFn`? This will avoid extra data being sent to the runtime.

packages/recipes/src/createRuntimeFn.ts
```diff
@@ -64,5 +64,10 @@ export const createRuntimeFn = <Variants extends VariantGroups>(
 
   runtimeFn.variants = () => Object.keys(config.variantClassNames);
 
+  runtimeFn.classNames = {
```
mattcompiles, line 67: To prevent doing work at runtime for all users, we could instead make this a JS getter?

packages/recipes/src/types.ts
```diff
@@ -32,9 +38,16 @@ export type PatternOptions<Variants extends VariantGroups> = {
...
+  base: string;
+  variants: VariantsClassNames<Variants>;
+};
+
 export type RuntimeFn<Variants extends VariantGroups> = ((
   options?: VariantSelection<Variants>,
-) => string) & { variants: () => (keyof Variants)[] };
+) => string) & {
+  variants: () => (keyof Variants)[];
+} & { classNames: RecipeClassNames<Variants> };
```
mattcompiles, line 50: I think we could just add this to the previous intersection type.

mattcompiles (commented):
Hey @aspirisen 

Thanks for the work on this. We discussed this PR at our weekly catch-up today and we realised it'd be better to only do this work at runtime to avoid serialising more data into the runtime code for no reason. I've made some more specific comments but in general I think you can just remove `rawDefaultClassName` and `rawVariantClassNames` and instead compute them in a JS getter inside `createRuntimeFn`. 

Sorry about the back and forth feedback. Once this change is done we're happy to merge 😄

aspirisen:
Hey @mattcompiles 
I've made appropriate changes. I agree that it is better to move class names generation logic to runtime since mostly they will be used in vanilla files and there is no need to increase bundle size.

.changeset/weak-rabbits-design.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/recipes': minor
+---
+
+Make recipe's base class name to be always defined
```
mattcompiles, line 5:
```suggestion
```

.changeset/weak-rabbits-design.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/recipes': minor
+---
+
+Expose recipe's class names to allow their selection
```
mattcompiles, line 5:
```suggestion
Always create a base class name for a recipe
```

site/docs/packages/recipes.md
```diff
@@ -118,6 +118,59 @@ export const button = recipe({
...
+    [`.${button.classNames.base} &`]: {
+      padding: 10
+    },
+    [`.${button.classNames.variants.color.neutral} &`]: {
+      color: 'red'
+    }
+  }
+});
+```
+
```
mattcompiles, line 173:
```suggestion
> 🚧&nbsp;&nbsp;The class names do not have `.` at the beginning.

Exposed class names allow you to select special variant in cascade styles:

```ts compiled
// button.css.ts
import { recipe } from '@vanilla-extract/recipes';

export const button = recipe({
  base: { fontWeight: 'bold' },
  variants: {
    color: {
      neutral: { background: 'transparent' }
    }
  }
});

export const parent = style({
  background: 'blue',
  selectors: {
    [`.${button.classNames.base} &`]: {
      padding: 10
    },
    [`.${button.classNames.variants.color.neutral} &`]: {
      color: 'red'
    }
  }
});
```

```

mattcompiles: VE has a system for automatically handling this case so you don't need to add a `.` yourself.

mattcompiles (approved):
🍻 🚀 Thanks so much for your work on this. 👏 
Sorry it took awhile to get it merged.

PR #1117, opened 2023-06-09
feat: export SprinklesProperties type

gabro:
As discussed on Discord, exposing the `SprinklesProperties` type allows wrapping `createSprinkles` with custom utilities without resorting to awkward workarounds like:

```ts
type VarargParameters<T> = T extends (args: infer P) => any ? P : never;
type SprinklesProperties = VarargParameters<typeof createSprinkles>;
```

which it turns out are also very expensive for tsc to compute (I personally had compile times to skyrocket in a library I maintain).

### DIFF
--- a/.changeset/twenty-tables-report.md
+++ b/.changeset/twenty-tables-report.md
@@ -1,0 +1,5 @@
+---
+"@vanilla-extract/sprinkles": patch
+---
+
+Export `SprinklesProperties` type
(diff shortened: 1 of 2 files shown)


.changeset/twenty-tables-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vanilla-extract/sprinkles": patch
+---
+
+export `SprinklesProperties` type
```
askoufis, line 5:
```suggestion
Export `SprinklesProperties` type
```

gabro: Fixed 👍

askoufis (approved):

PR #1128, opened 2023-06-30
chore(deps): esbuild 0.17.19

belgattitude:
First a big thank you for vanilla-extract.

This PR updates esbuild to the latest esbuild 0.17.x (frm 0.17.6 -> [0.17.19](https://github.com/evanw/esbuild/releases/tag/v0.17.19)). It might bring fixes (and of course create issues) but my motivation is to reduce the number of installed esbuild versions 

```
├─ @esbuild-kit/core-utils@npm:3.1.0
│  └─ esbuild@npm:0.17.18 (via npm:~0.17.6)
├─ @vanilla-extract/integration@npm:6.2.1
│  └─ esbuild@npm:0.17.6 (via npm:0.17.6)
├─ vite@npm:4.3.9 [3ba6e]
│  └─ esbuild@npm:0.17.18 (via npm:^0.17.5)
├─ tsup@npm:7.1.0
│  └─ esbuild@npm:0.18.3 (via npm:^0.18.2)
(...)
```

Depending on how the update is considered, I'll also pr on remix-dev which locks esbuild to 0.17.6 as well

### DIFF
--- a/packages/esbuild-plugin-next/package.json
+++ b/packages/esbuild-plugin-next/package.json
@@ -19,6 +19,6 @@
     "@vanilla-extract/integration": "^6.0.2"
   },
   "devDependencies": {
-    "esbuild": "0.17.6"
+    "esbuild": "0.17.18"
   }
 }
--- a/packages/esbuild-plugin/package.json
+++ b/packages/esbuild-plugin/package.json
@@ -18,6 +18,6 @@
     "@vanilla-extract/integration": "^6.2.0"
   },
   "devDependencies": {
-    "esbuild": "0.17.6"
+    "esbuild": "0.17.18"
   }
 }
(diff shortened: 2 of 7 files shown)


askoufis:
Hey 👋, thanks for the PR.

We've been somewhat hesitant in the past to bump the esbuild version too often, primarily because it's a 0-versioned package, so theoretically any version change, even a patch, could break something. In addition to that, there have been unintentional breaking changes in esbuild in the past, so it has been advantageous to pin the version to a known "good" version.

While we obviously won't stay on this version forever, until there's a good enough reason to upgrade, e.g. bug fixes/performance/new features, we likely won't just bump the version because it's a few months old/to keep up with versions in other packages. Ultimately you can configure `resolutions`/`overrides` in your package manager of choice to alleviate this issue anyway.

belgattitude:
Fully understand the choices made. I'll keep an eye on it. Thx

Issue #1118, opened 2023-06-09
Nuxt3 + vanilla extract, issue with HMR when SSR is enabled


Hecatron:
### Describe the bug

I've recently had another look at using vanilla extract for use with Nuxt3.
It appears mostly fine although it looks as if HMR doesn't work if SSR is enabled for the site

I've noticed that vite / nuxt is detecting a change and reloading the page
but it's just not showing up even with a full manual page refresh.
Disabling SSR across the entire nuxt app seems to work, but prevents it from being used with themes such as docus.
I've tested this on both Windows 10 and a Rpi running raspian and the results are the same so I don't think it's an issue unique to windows pathing.

### Reproduction

https://github.com/Hecatron-Forks/vanilla-extract-example

### System Info

```shell
windows:

  System:
    OS: Windows 10 10.0.19045
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
    Memory: 2.27 GB / 15.68 GB
  Binaries:
    Node: 19.8.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.1.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (114.0.1823.41)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    @vanilla-extract/css: ^1.11.1 => 1.11.1
    @vanilla-extract/vite-plugin: ^3.8.2 => 3.8.2
```

Rpi / Raspian:
```
System:
    OS: Linux 6.1 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (4) arm64 Cortex-A72
    Memory: 7.20 GB / 7.63 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 20.0.0 - /usr/local/bin/node
    npm: 9.6.4 - /usr/local/bin/npm
  Browsers:
    Firefox: 102.12.0esr
  npmPackages:
    @vanilla-extract/css: ^1.11.1 => 1.11.1
    @vanilla-extract/vite-plugin: ^3.8.2 => 3.8.2
```

### Used Package Manager

pnpm

### Logs

```shell
I'm not seeing any errors in the console
```

### Validations

[...]

PR #1140, opened 2023-07-17
feat: Added example for Nuxt3

grbd:
I've added an example for Nuxt 3
If this goes through I might also be tempted to add some more for other frameworks as well, such as svelte etc.


examples/nuxt/nuxt.config.ts
```diff
@@ -0,0 +1,16 @@
...
+  telemetry: false,
+  vite: {
+    plugins: [
+      vanillaExtractPlugin({})
+    ],
+  },
+  // Note currently server side rendering seems to break hmr (hot module reloading) with Nuxt
+  // One workaround is to instead bundle the vue components into a seperate vite built library
+  // Or disable SSR while developing components
+  ssr: false,
```
graup, line 15: How about linking the issue https://github.com/vanilla-extract-css/vanilla-extract/issues/1118

examples/nuxt/components/HelloWorld.vue.css.ts
```diff
@@ -0,0 +1,5 @@
+import { style } from '@vanilla-extract/css'
+
+export const externalStyle = style({
+  color: 'red'
+})
```
graup, line 5: This is as barebones as it gets. How do you feel about showing at least a few more of the basic features, like themes, vars, variants and maybe dynamic variables?

askoufis:
@Hecatron Closing this for now. With many frameworks aligning towards Vite, I'm not sure how much value there is in having an example for every single Vite-based framework since they will all likely be configured the same way - with the Vanilla Extract Vite plugin.

PR #1142, opened 2023-07-19
Update getting-started.md

mikeytown19:
Not sure if this is intended, but when following allowing the setup i noticed these were named differently. 
Does it make sense to match the name from the above example?
image


mattcompiles (approved):
Thanks for the PR

PR #1150, opened 2023-07-25
:sparkles: inline/load css to style element

salamaashoush:
In some environments and use cases (eg Figma plugins) you can't load external resources with relative paths so you either have to host css separately on a remote server or inline it in the js/html so this PR adds a new option `useStyleLoader` to `@vanilla-extract/esbuild-plugin` that allows use to skip CSS loader and inject the CSS using style element in current document, by emitting a js file with the code to do that similar to (style-loader)

### DIFF
--- a/packages/esbuild-plugin-next/src/index.ts
+++ b/packages/esbuild-plugin-next/src/index.ts
@@ -11,12 +11,20 @@ import type { Plugin } from 'esbuild';
 
 const vanillaCssNamespace = 'vanilla-extract-css-ns';
 
+async function hashString(str: string) {
+  const encoder = new TextEncoder();
+  const data = encoder.encode(str);
+  const hash = await crypto.subtle.digest('SHA-256', data);
+  return hash;
+}
+
 interface VanillaExtractPluginOptions {
   outputCss?: boolean;
   runtime?: boolean;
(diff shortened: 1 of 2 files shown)


graup:
I'm not sure this should be an option of VE, it's a very specific use case. Isn't it possible to configure esbuild (maybe with a plugin) so that it transforms the css loader result into this kind of style loader?

salamaashoush:
> I'm not sure this should be an option of VE, it's a very specific use case. Isn't it possible to configure esbuild (maybe with a plugin) so that it transforms the css loader result into this kind of style loader?

I tried to get working it working as separate plugin but it didn't work

yaojiu19:
Can you release a version. thk

askoufis:
> Can you release a version. thk

@yaojiu19 I'm not sure if this PR is ready to test, or is even something we want to still pursue.

@salamaashoush Any updates?

salamaashoush:
I can rebase and update the MR if you want to test and merge it, currently i copied the plugin into my project repository and using it

yaojiu19:
> I can rebase and update the MR if you want to test and merge it, currently i copied the plugin into my project repository and using it

yeah, manually import css files is inconvenient in monorepo‘s project

PR #1157, opened 2023-08-07
Support extracting a CSS var name

adrhumphreys:
Ran into some situations where we were setting CSS var names from inside components (not ideal) but this meant that it was handy to have a way of getting the CSS variable name to be able to set it.

Example usage would be:
```ts
import { createVar, getVarName } from "@vanilla-extract/css";

const maxHeight = createVar();

export const root = style({
  vars: {
    [maxHeight]: "0px",
  },
  height: `calc(100vh - ${maxHeight})`,
});

// maxHeight could be exported and this could be called in a different file
export const maxHeightVarName = getVarName(maxHeight);
```

### DIFF
--- a/.changeset/cool-pumas-hope.md
+++ b/.changeset/cool-pumas-hope.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': minor
+---
+
+Support extracting a css var name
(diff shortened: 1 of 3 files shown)


mihkeleidast:
Curious as to how you would set the variable in the component with this? I found myself in a similar need a while back and stumbled onto [setElementVars](https://vanilla-extract.style/documentation/packages/dynamic/#setelementvars) from the dynamic package, which does both the extracting and also setting of the vars on the element. Perhaps that would suffice for you as well?

adrhumphreys:
@mihkeleidast yeah that 100% solves my issue. Do you think its worthwhile for me to make a PR to update the docs with a link from `createVar` to `setElementVars`? Or is this more of a one off, me just not reading the docs thorough enough

PR #1160, opened 2023-08-10
Allow custom identifier hashing

SombreroElGringo:
Hello, I am submitting a request to reopen Pull Request #657. This proposed enhancement aims to unlock my team, which is currently facing a classname clash issue, we are using multiple micro-frontends within a host application. Unfortunately, some of the generated hash values from these micro-frontends coincide with hash values from the host application. We are really hoping for this feature to be integrated, like that it will allows us to customize the generated classnames by applying prefixes specific to each micro-frontend and will guarantee the uniqueness of every hash.

Moreover, the implementation of this feature could help multiple persons with similar scenarios and other challenges, we can found in the following discussion:

- https://github.com/vanilla-extract-css/vanilla-extract/discussions/313
- https://github.com/vanilla-extract-css/vanilla-extract/discussions/273
- https://github.com/vanilla-extract-css/vanilla-extract/discussions/234

Also, I'd like to highlight a similar feature available in [Material UI](https://mui.com/system/styles/api/#creategenerateclassname-options-class-name-generator) allowing to customize the generated classnames.

Thank you for your consideration.

Best regards,

### DIFF
--- a/.changeset/curly-lies-flash.md
+++ b/.changeset/curly-lies-flash.md
@@ -1,0 +1,6 @@
+---
+'@vanilla-extract/css': minor
+'site': minor
+---
+
+Users can now provide a custom identifier hashing function
--- a/.changeset/curly-lies-flash.md
+++ b/.changeset/curly-lies-flash.md
@@ -1,6 +1,9 @@
 ---
 '@vanilla-extract/css': minor
-'site': minor
+'@vanilla-extract/dynamic': minor
+'@vanilla-extract/integration': minor
+'@vanilla-extract/recipes': minor
+'@vanilla-extract/sprinkles': minor
 ---
 
 Users can now provide a custom identifier hashing function
--- a/.changeset/curly-lies-flash.md
+++ b/.changeset/curly-lies-flash.md
@@ -1,9 +1,10 @@
 ---
 '@vanilla-extract/css': minor
-'@vanilla-extract/dynamic': minor
-'@vanilla-extract/integration': minor
-'@vanilla-extract/recipes': minor
-'@vanilla-extract/sprinkles': minor
+'@vanilla-extract/webpack-plugin': minor
+'@vanilla-extract/esbuild-plugin': minor
+'@vanilla-extract/rollup-plugin': minor
+'@vanilla-extract/vite-plugin': minor
+'@vanilla-extract/next-plugin': minor
 ---
 
 Users can now provide a custom identifier hashing function
--- a/packages/css/src/identifier.ts
+++ b/packages/css/src/identifier.ts
@@ -42,7 +42,7 @@ export function generateIdentifier(options?: GenerateIdentifierOptions): string;
 export function generateIdentifier(
   arg?: string | GenerateIdentifierOptions,
 ): string {
-  const indentOption = getIdentOption();
+  const identOption = getIdentOption();
   const { debugId, debugFileName = true } = {
     ...(typeof arg === 'string' ? { debugId: arg } : null),
     ...(typeof arg === 'object' ? arg : null),
@@ -58,7 +58,7 @@ export function generateIdentifier(
 
   let identifier = `${fileScopeHash}${refCount}`;
 
-  if (indentOption === 'debug') {
+  if (identOption === 'debug') {
     const devPrefix = getDevPrefix({ debugId, debugFileName });
 
     if (devPrefix) {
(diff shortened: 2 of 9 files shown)


packages/css/src/types.ts
```diff
@@ -120,7 +120,11 @@ export interface Composition {
   classList: string;
 }
 
-type IdentOption = 'short' | 'debug';
+type IdentOption =
+  | 'short'
+  | 'debug'
+  | ((scope: string, refCount: string, debugId: string | undefined) => string);
```
mattcompiles, line 126: Few thoughts about this:
- I think this should instead be an object for named params. 
- I think `scope` is too abstract, seeing as this is a hash of the file path. I think it'd be more useful to provide a general hash for the identifier which would be `scope + refCount`. Let's just call it `hash`.
- Maybe also add filename and packageName to be fed into a custom hash

packages/css/src/identifier.ts
```diff
@@ -38,6 +42,7 @@ export function generateIdentifier(options?: GenerateIdentifierOptions): string;
 export function generateIdentifier(
   arg?: string | GenerateIdentifierOptions,
 ): string {
+  const indentOption = getIdentOption();
```
mattcompiles, line 45:
```suggestion
  const identOption = getIdentOption();
```

.changeset/curly-lies-flash.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/css': minor
+'site': minor
```
mattcompiles, line 3: We should not version the site. Can we also add a minor for the affected plugins.

mattcompiles (commented):
Hi, Thanks for the PR. 

The microfrontend use-case convinces me this feature is necessary. I've left some comments to be addressed, but we can get this merged once that's done.

.changeset/curly-lies-flash.md
SombreroElGringo: Updated with the affected packages, let me know if I missed some

packages/css/src/identifier.ts
SombreroElGringo: Fixed

packages/css/src/types.ts
SombreroElGringo: Thank for the suggestion, I updated its type and added the extra params suggested, also for the filename I think it makes sense to just pass the file path and give the freedom to the developer to parse this file path or X to retrieve himself the filename or do X operations on it.

packages/css/src/types.ts
```diff
@@ -120,7 +120,15 @@ export interface Composition {
   classList: string;
 }
 
-type IdentOption = 'short' | 'debug';
+type CustomIdentFunction = (params: {
+  hash: string;
+  debugId?: string;
+  filePath?: string;
```
mattcompiles, line 126: `filePath` will always be available
```suggestion
  filePath: string;
```

site/docs/integrations/esbuild.md
```diff
@@ -100,6 +100,7 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
 
 - `short` identifiers are a 7+ character hash. e.g. `hnw5tz3`
 - `debug` identifiers contain human readable prefixes representing the owning filename and a potential rule level debug name. e.g. `myfile_mystyle_hnw5tz3`
+- A function accepting three parameters: A scope name, an index, and an optional rule level debug name, and returning the identifier.
```
mattcompiles, line 103: Can you update this and the other docs to reference the new API.

mattcompiles (commented):
Looking good, can you please update the type and the docs.

packages/css/src/types.ts
SombreroElGringo: Done

site/docs/integrations/esbuild.md
SombreroElGringo: Done

.changeset/curly-lies-flash.md
```diff
@@ -0,0 +1,9 @@
+---
+'@vanilla-extract/css': minor
+'@vanilla-extract/dynamic': minor
+'@vanilla-extract/integration': minor
+'@vanilla-extract/recipes': minor
+'@vanilla-extract/sprinkles': minor
```
mattcompiles, line 6:
```suggestion
'@vanilla-extract/css': minor
'@vanilla-extract/webpack-plugin': minor
'@vanilla-extract/esbuild-plugin': minor
'@vanilla-extract/rollup-plugin': minor
'@vanilla-extract/vite-plugin': minor
'@vanilla-extract/next-plugin': minor
```

PR #1161, opened 2023-08-11
Resolve RecipeVariants type to improve IDE hint

PrettyCoffee:
Hey, when testing out vanilla-extract, I found that the type hints are currently very verbose with the generic `RecipeVariants` type and not readable.

For example when having a button that has the variants `look`, `size`, `pill` and `square`, it currently looks like this in my IDE (VSCode):

image

By using a generic type like this:
```ts
type Resolve<T> = {
  [K in keyof T]: T[K]
} & {}
```
you can kind of "resolve" the type and only display the actually underlying type:

image

which has much better readability! :)

This is not changing the type itself, it is more like reassigning the attributes to a new type.

Loving the project so far, great work! ♥️


askoufis:
Thanks for the PR! Love the screenshots.

We've wanted to do this where it made sense in the repo, but never got around to it. It definitely improves the types significantly.

Your PR looks fine, but you'll need to add a changeset in order to get the change released. If you're unfamiliar with changesets, have a look at the links in [the changeset bot comment above](https://github.com/vanilla-extract-css/vanilla-extract/pull/1161#issuecomment-1674680287).

askoufis (approved):

PrettyCoffee:
@askoufis thanks a lot for adding the changeset! Unfortunately, I didn't have a lot of time last week. 😅

Issue #637, opened 2022-03-23
ReferenceError: __webpack_require__ is not defined


LekoArts:
### Describe the bug

Hi!

I'm not sure yet if this is a bug in webpack or in vanilla-extract, but I haven't seen this problem 5 days ago when I did test it out with https://github.com/gatsbyjs/gatsby/pull/35166. I've now re-run our example site https://github.com/gatsbyjs/gatsby/tree/master/examples/using-vanilla-extract and it fails with these kinds of errors:

```shell
ERROR in ./src/styles/404.css.ts
Module build failed (from ./node_modules/@vanilla-extract/webpack-plugin/loader/dist/vanilla-extract-webpack-plugin-loader.cjs.js):
ReferenceError: __webpack_require__ is not defined
    at eval (webpack-internal:///./src/styles/404.css.ts:1:1)
    at eval (<anonymous>)
    at Object../src/styles/404.css.ts (/Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/src/styles/404.css.ts:21:1)
    at __webpack_require__ (/Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/src/styles/404.css.ts:67:42)
    at /Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/src/styles/404.css.ts:123:37
    at /Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/src/styles/404.css.ts:126:12
    at Script.runInContext (node:vm:139:12)
    at Script.runInNewContext (node:vm:144:17)
    at Object.module.exports [as default] (/Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/node_modules/eval/eval.js:71:12)
    at Object.processVanillaFile
(/Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/node_modules/@vanilla-extract/integration/dist/vanilla-extract-integration.cjs.dev.js:108:50)
    at
/Users/lejoe/code/work/gatsby/examples/using-vanilla-extract/node_modules/@vanilla-extract/webpack-plugin/loader/dist/vanilla-extract-webpack-plugin-loader.cjs.dev.js:60:38
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at runNextTicks (node:internal/process/task_queues:65:3)
    at processImmediate (node:internal/timers:437:9)
```

[...]

PR #1167, opened 2023-08-22
Bump `eval` to `0.1.8` to fix missing Node globals issues

askoufis:
The `eval` package has caused issues [in the past](https://github.com/vanilla-extract-css/vanilla-extract/issues/637) that resulted in [pinning its version](https://github.com/vanilla-extract-css/vanilla-extract/pull/639/files#diff-62f6906083a84673a05063f208cfe697e0134acee3dded6930bab0d43f0789dfR19). At roughly the same time, `0.1.8` was released but we didn't update. `0.1.8` [fixes some issues](https://github.com/pierrec/node-eval/commits/master) we've run into with Node globals, and given its stability over the last ~1 year I think it's safe to update.

### DIFF
--- a/packages/integration/src/processVanillaFile.test.ts
+++ b/packages/integration/src/processVanillaFile.test.ts
@@ -1,6 +1,3 @@
-// @ts-expect-error
-import evalCode from 'eval';
-
 import { serializeVanillaModule } from './processVanillaFile';
 
 describe('serializeVanillaModule', () => {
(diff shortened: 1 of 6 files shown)


tadhglewis (approved):

packages/integration/src/processVanillaFile.test.ts
```diff
@@ -1,6 +1,3 @@
-// @ts-expect-error
-import evalCode from 'eval';
```
askoufis, line 2: I guess we don't run `tsc` over test files so this unused dep wasn't picked up.

packages/integration/src/processVanillaFile.ts
```diff
@@ -95,7 +94,7 @@ export async function processVanillaFile({
     filePath,
     { console, process, __adapter__: cssAdapter },
     true,
-  );
+  ) as Record<string, unknown>;
```
askoufis, line 97: `eval`'s result type is `unknown`, hence the cast.

mattcompiles (approved):

PR #1175, opened 2023-08-31
Allow `assignInlineVars` and `setElementVars` to accept potentially `null` or `undefined` values when a theme contract is not provided

### LINKED ISSUE #536: assignInlineVars doesnt accept undefined values
Looks like right now we can work around this by passing an empty string, like so:

```
 style={assignInlineVars({ [pointColor]: point.color ?? '' })}
```

However, a nice quality of life change would be  to just pass a possibly undefined value and let a `fallbackVar` handle it in my .css.ts

youngkyo0504:
Fixes #536

It works!

```ts
type DeepPartial<Obj> = { [Prop in keyof Obj]?: Obj[Prop] extends Primitive ? Obj[Prop] : DeepPartial<Obj[Prop]> }

export const vars = createThemeContract({
  color: null,

  hover: {
    containerPadding: null,
    radius: null,
  },
})

 style={assignInlineVars<DeepPartial<type of vars>>({ 
   hover: { radius: '15px' }
 })}
```

**EDIT**: The code snippet above is not a reflection of the feature implemented in this PR. See the linked issue or [the comment below](https://github.com/vanilla-extract-css/vanilla-extract/pull/1175#issuecomment-1786950320) for more details.


askoufis:
@youngkyo0504 This doesn't quite address the issue as you still can't actually assign `undefined` to a property in `assignInlineVars`. You'll see that it doesn't work if you try it in the test cases.

The other issue that needs to be addressed is what to actually do with the `undefined` value. Do we `unset` the style, or drop it from the style object entirely?

youngkyo0504:
@askoufis  Thanks for the reply! I think the issue have difference with my case. In my case DeepPartial does not work. I wrote test case wrong by mistake. It is correct test case in my case. 

```ts
const style = assignInlineVars<DeepPartial<S>>(vars,{
  color: 'red'
})
```

but It have somthing in common with that issue.  But  it doesn't work  in the test cases in issue. could I do additional work again? 

If `undefined` or `null`  comes as an argument , dropping it from the style object entirely is better than setting `unset`

youngkyo0504:
I add additional work in  9c89553a154db8894fcfc606cc90f110f8c515bd

askoufis:
@youngkyo0504 I've added some changes on top of your work.

There was no need to modify `MapLeafNodes` as that's only used when passing in a theme contract. It doesn't really make sense to allow `undefined` or `null` values if you're providing a theme contract as the entire point is to provide all values in the contract.

This feature really only applies to the first overload of `assignInlineVars`, so that's what I changed.

Additionally, I added the same functionality to `setElementVars` as it has a similar API. Also added some tests for it while I was there.

Finally I added some docs for this feature.

askoufis (approved):

Issue #1152, opened 2023-07-26
When developing with Next.js, css is not generated


Karibash:
### Describe the bug

When a page transition is made on the Next.js development server, the css of the destination page is not generated.
Reproduced with version 2.2.1 of @vanilla-extract/next-plugin, lowering the version to 2.1.3 makes it work correctly.

### Reproduction

https://stackblitz.com/edit/stackblitz-starters-dcvpht

### System Info

```shell
System:
    OS: macOS 13.4.1
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 1.16 GB / 32.00 GB
    Shell: 3.6.1 - /usr/local/bin/fish
  Binaries:
    Node: 18.16.0 - ~/.local/share/nvm/v18.16.0/bin/node
    npm: 9.5.1 - ~/.local/share/nvm/v18.16.0/bin/npm
  Browsers:
    Chrome: 114.0.5735.248
    Firefox: 81.0.2
    Safari: 16.5.2
```

### Used Package Manager

pnpm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

edenstrom:
I've been looking into this a bit. It only happens between navigation in the `pages` directory. Navigating in the `app` directory works perfectly. Also navigating from a route in `app` to a route in `pages` and vice versa.

One more thing is that HMR doesn't work either.

1. Go to home
2. Change index.css.js (CSS is updated)
3. Navigate to other
4. Change other.css.js (CSS is not updated)
5. Refresh
6. CSS is loaded and HMR works again

No solution in sight though. Trying to debug further.

Possible culprit? https://github.com/vanilla-extract-css/vanilla-extract/pull/1105

hazzo:
Same here, pls fix this :)

syfxlin:
[...]

PR #1180, opened 2023-09-11
fix(next-plugin): build not outputting css on windows

### LINKED ISSUE #1086: Next.js 13 - Builds not outputting css on Windows
### Describe the bug

- Clone the ["should be working repo"](https://github.com/SuttonJack/vanilla-extract-app-dir) in Windows

- Inspect with DevTools, and see footer class being generated without styles

image

Need help @markdalgleish @SuttonJack 

[Reference in discussion](https://github.com/vanilla-extract-css/vanilla-extract/issues/929#issuecomment-1539274116)

Edit:
Running the app inside container seems fine.

### Reproduction

https://github.com/SuttonJack/vanilla-extract-app-dir

### System Info

```shell
Windows 11
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

syfxlin:
Fix #1086 #929

Next.js removes the resource matching query (`!=!`) and the inline loader (`!virtual-file-loader.js?{}!`), so you need to pass the virtual CSS file directly to `noop.css`, and then extract the virtual CSS file from `noop.css` via a loader like this:

```typescript
import "noop.css?{ filename, source }"
```

This PR change is validated for Windows/Linux/macOS and can be used to output CSS normally in Next.js 13 & React Server Components.

References:
- [syfxlin/next-plugin-vanilla-extract](https://github.com/syfxlin/next-plugin-vanilla-extract)
- [syfxlin/vanilla-extract-demos](https://github.com/syfxlin/vanilla-extract-demos)


askoufis (approved):
Thanks for the contribution! Fingers crossed this addresses all existing next.js issues.

I checked out your demos repo on windows and linux and can confirm that everything is working as expected. Also tested your next plugin with some of the reproductions in the next.js issues and they work too.

hazzo:
Did not fixed this https://github.com/vanilla-extract-css/vanilla-extract/issues/1152. Had to rollback and apply the `patch-package` :S

askoufis:
> Did not fixed this #1152. Had to rollback and apply the patch with `patch-package` :S

The PR never stated that it would fix that issue. [I was hopeful in the above comment that it might fix other issues](https://github.com/vanilla-extract-css/vanilla-extract/pull/1180#pullrequestreview-1629912692), but it didn't. It seems that #1152 is caused by something else.

PR #1185, opened 2023-09-16
Bump `pnpm` to v8, `node` to v20

askoufis:
Doing a bit of spring cleaning :broom:. Up first is `pnpm` and `node`. Nothing of note for `node`, but the new `pnpm` major came with a new lockfile format.

**NOTE**: Not dropping support for any specific version of node in this PR. We don't even officially have a supported version anyway (AFAIK), but I think it's safe to drop testing on 14 given [it became EOL ~4 months ago](https://endoflife.date/nodejs). 16 just became EOL a few days ago, but happy to keep it around for now.

This might also require a tweak to the branch protection since it's expecting a `node 14` CI step to pass but it's now named `node 16`.


mrm007 (approved):

package.json
```diff
@@ -60,10 +52,18 @@
...
+    ]
+  },
+  "packageManager": "pnpm@8.10.2",
   "pnpm": {
     "patchedDependencies": {
       "html-render-webpack-plugin@3.0.1": "patches/html-render-webpack-plugin@3.0.1.patch"
     }
+  },
+  "volta": {
+    "node": "20.9.0"
```
askoufis, line 67: Can we keep the `pnpm` volta config, or does this cause an error if you don't enable have the volta envar set?

mrm007: We can, however it's prone to be forgotten (as was the case with this PR). Upgrading pnpm means you need to change `packageManager`, as well as the `volta.pnpm` keys.

`packageManager` is more of a de facto standard, which is why I added it.

mattcompiles (approved):

PR #1187, opened 2023-09-17
fix: fix typo

youngkyo0504:
normalise -> normalize

Does it need changeset?


askoufis:
Thanks for the PR, but I wouldn't consider this a typo. Either spelling is valid, one is American English (z), one is British English (s). I don't think this is worth changing, and we don't really have a strict guideline or anything about spelling, nor do I think we really need one.

PR #1193, opened 2023-10-02
fix(next-plugin): Next.js 13.5.x build not outputting css on windows

syfxlin:
CSS output is not working again in Next.js 13.5.x and Windows.
The `resolvedNextConfig.experimental.appDir` is no longer available in Next.js 13.5.x, so this pre-condition has been removed.

References: https://github.com/vercel/next.js/blob/76e06d55a8c42090dafda25ec31d9eeaa7ec647c/packages/next/src/build/index.ts#L418-L422


packages/next-plugin/src/index.ts
```diff
@@ -115,9 +115,7 @@ export const createVanillaExtractPlugin = (
         dir,
         resolvedNextConfig.experimental?.appDir ?? false,
```
askoufis, line 116: Should this also be removed?
```suggestion
```

syfxlin: I think this can be kept so that it is compatible with Next.js 13.4.x and lower versions.
References: https://github.com/vercel/next.js/blob/v13.4.19/packages/next/src/lib/find-pages-dir.ts#L50

plrs9816:
I think comments can be changed a bit
```ts
      // https://github.com/vercel/next.js/blob/5b52e7772dd6237474931ba4bbb856fbc33142ad/packages/next/src/build/index.ts#L431
      // https://github.com/vercel/next.js/blob/5b52e7772dd6237474931ba4bbb856fbc33142ad/packages/next/src/build/webpack-config.ts#L493
      // https://github.com/vercel/next.js/pull/43916
      // on Next.js 12, findPagesDirResult is a string. on Next.js 13, findPagesDirResult is an object
      const findPagesDirResult = findPagesDir(
        dir,
        resolvedNextConfig.experimental?.appDir ?? false,
      );
      // Skip nextConfig check since appDir is stable feature after Next.js 13.4
      const hasAppDir = !!(findPagesDirResult && findPagesDirResult.appDir);
```

askoufis (approved):
LGTM. Feel free to add the comments suggested in https://github.com/vanilla-extract-css/vanilla-extract/pull/1193#issuecomment-1764105247.

syfxlin:
Hi @askoufis, I added the comment suggested in add [#1193](https://github.com/vanilla-extract-css/vanilla-extract/pull/1193#issuecomment-1764105247), but I don't have access to merge pull request, can you help with that?

PR #1197, opened 2023-10-12
Use modern-ahocorasick

brandonpittman:
The existing `ahocorasick` package doesn't export anything which can cause issues with certain frameworks. The `modern-ahocorasick` package has the same API as the old one but actually exports something instead of adding to the global namespace.

### DIFF
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -109,7 +109,6 @@
   "dependencies": {
     "@emotion/hash": "^0.9.0",
     "@vanilla-extract/private": "^1.0.3",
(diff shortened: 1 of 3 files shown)


mattcompiles (approved):

mattcompiles:
@brandonpittman looks like the lockfile is invalid. Maybe a different pnpm version?

brandonpittman:
It did look like it updated the lockfile version.

brandonpittman:
Lowered my local PNPM version to match the repo's. Seems like it sorted out the lockfile changes.

Issue #1133, opened 2023-07-09
Mismatch between Server and Client Classes when Composing Styles with Vite


jahredhope:
### Describe the bug

Vanilla Extract adds an additional Identifier to classes to ensure it's safe for use in selectors. This gets stripped when unused.

In some cases this is only stripped from client code and remains in server code, resulting in a mismatch between server and client, also known as a hydration error.

> When composed styles are used in selectors, they are assigned an additional class if required so they can be uniquely identified. When selectors are processed internally, the composed classes are removed, only leaving behind the unique identifier classes. This allows you to treat them as if they were a single class within vanilla-extract selectors. - [v1.4.0](https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/css/CHANGELOG.md#140)

## Details

When creating a style composed of other styles, (either by passing an array to style or calling a Sprinkles function in a css file) Vanilla Extract will create an additional identifier class.
If the class is unused, the class will appear in the server render, but will be stripped for the client render, resulting in a mismatch between renders known as a hydration error.

This only occurs when:

- `emitCssInSsr` is not true
- A style is defined as a composition of other styles. (Or, a Sprinkles function called in a css.ts file)
- The resulting value must be exported, and NOT used in another selector

Though Sprinkles has been mentioned here, the demonstration below reveals this occurs without Sprinkles. It's just that your more likely to be in this scenario if you are using atomic styles.

## Attempted Workarounds / Solutions

As mentioned setting `emitCssInSsr` to true did solve my issue, and may be a more preferable configuration anyway.

- Perhaps it would make sense to default `emitCssInSsr` to `true`?

If not, I suspect the result of this code path:

https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/vite-plugin/src/index.ts#L163 calling `@vanilla-extract/integration`'s `processVanillaFile` is the key differentiator.

---

This may be related to https://github.com/vanilla-extract-css/vanilla-extract/issues/654 which has a similar error, but has quite a different scenario.

### Reproduction

https://github.com/jahredhope/ve-sprinkles-in-styles-bug

### System Info

```shell
System:
    OS: macOS 13.4
    CPU: (10) arm64 Apple M1 Pro
    Memory: 3.93 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - ~/.volta/tools/image/node/18.16.1/bin/node
[...]

PR #1203, opened 2023-10-23
Fix next pages router style loading, add next snapshot tests

renrizzolo:
I thought it would be a good idea to have some e2e tests for Next.js, seeing as there have been a few issues, particularly with the introduction of the App Router / keeping backwards compatibility with next 12. 
There was a lot of trial and error to get all this working, so if there's anything that looks odd I should have an explanation 😄 

### Next Plugin
- always output css
  - fixes a bug with composed selector global css class not being applied in SSR with pages router
- use `next-style-loader` for pages router in development (as per original comment, next-style-loader can mess up css order in development mode, but this is surely better than no css at all.)
  - fixes a bug with new page styles not being added to the document when client side routing #1152
- remove appDir config check
  
###  e2e / Fixtures
- added next (v13) app router (`@fixtures/next-app-router`) and next (v12) pages router (`@fixtures/next-pages-router`). 
These apps themselves import other fixtures, and serve them at `/[fixture]`. 
NOTE: I only added `recipes`, `sprinkles` and `features`. I can add them all, but I figured the fixtures themselves are well covered. 
- added next dev and prod snapshot testing. 
These need to be separate, and serial, because next can't really handle serving dev/prod at the same time. 
In order to be able to assert the screenshot from next-*/[fixture] matches the screenshot of [fixture], I had to move all screenshots to a single folder. I left the CSS in the existing test folders to at least keep them separate.
- add build step for the next apps to the `test:playwright` command. If this is annoying for development we could add a separate CI version of this command that is called only in the validate workflow.

### Other
- Updated playwright to ^1.38.1. This allows to do the custom `snapshotPathTemplate` for screenshots, which also removes the need for the `_autoSnapshotSuffix` workaround.
- bumped the port range due to chrome `net::ERR_UNSAFE_PORT` for 10xxx ports. I guess the newer playwright version has something to do with this popping up now?
- add `dedupe-peer-dependents` to `.npmrc`. This is the default in pnpm 8, and was needed so that multiple versions of next 12 wouldn't be installed and used by the next plugin and next 12 fixture app respectively.


askoufis (commented):
Really appreciate the effort on this. I've got just a few comments, we'll see how things go on CI 🤞

tsconfig.json
```diff
@@ -18,7 +18,7 @@
     "isolatedModules": true,
     "downlevelIteration": true
   },
-  "exclude": ["node_modules", "**/dist/**"],
+  "exclude": ["node_modules", "**/dist/**", "fixtures/next-app-router/**/*"],
```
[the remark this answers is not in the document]
renrizzolo: oops, I wrote some comments on my fork and forgot to add them here.
There was some type conflicts happening where react router in the site package was picking up react 18 types from the `next-app-router` package. 
Had another look and I think this can be resolved by not using react 18 types at all.

askoufis:
> next-style-loader can mess up css order in development mode, but this is surely better than no css at all

If we're going to add this functionality back in, it should probably be documented as a caveat/gotcha in [the next integration docs](https://vanilla-extract.style/documentation/integrations/next/).

packages/next-plugin/src/index.ts
```diff
@@ -115,15 +145,9 @@ export const createVanillaExtractPlugin = (
...
-        !!resolvedNextConfig.experimental?.appDir &&
-        !!(findPagesDirResult && findPagesDirResult.appDir);
+      const hasAppDir = !!(findPagesDirResult && findPagesDirResult.appDir);
 
-      const outputCss = hasAppDir
-        ? // Always output css since Next.js App Router needs to collect Server CSS from React Server Components
-          true
-        : // There is no appDir, do not output css on server build
-          !isServer;
+      const outputCss = true;
```
renrizzolo, line 150: for reference, if `outputCss` is false on the server (using pages router), the class created by the globalStyle below won't be applied to the ssr'd html

```ts
export const styleCompositionInSelector = style([
  style({ color: 'white' }),
  style({ backgroundColor: 'black' }),
]);

globalStyle(`body ${styleCompositionInSelector}`, {
  fontSize: '24px',
});
```

it *will* work if the selector looks like this

```ts
export const styleCompositionInSelector = style([
  {},
  style({ color: 'white' }),
  style({ backgroundColor: 'black' }),
]);
```

I think this is the same issue as #1133

fixtures/next-app-router/package.json
```diff
@@ -0,0 +1,38 @@
...
+  "dependencies": {
+    "@fixtures/features": "*",
+    "@fixtures/recipes": "*",
+    "@fixtures/sprinkles": "*",
+    "@vanilla-extract/css": "*",
+    "@vanilla-extract/recipes": "*",
+    "@vanilla-extract/sprinkles": "*",
+    "next": "npm:next@13.5.4",
+    "react": "npm:react@^18.2.0",
+    "react-dom": "npm:react-dom@^18.2.0"
```
renrizzolo, line 25: https://github.com/Thinkmill/manypkg#ignoring-this-rule

renrizzolo:
I'm not entirely sure how to reproduce the `next-style-loader` development ordering issue to comment on it. 
The only time i can see this happening is with hmr: change a style in a `css.ts` file, after saving it, it will be moved to then end of the style blocks in the document head.

askoufis:
> I'm not entirely sure how to reproduce the `next-style-loader` development ordering issue to comment on it. The only time i can see this happening is with hmr: change a style in a `css.ts` file, after saving it, it will be moved to then end of the style blocks in the document head.

Ah ok. I suppose that's almost what you'd want for HMR anyway. You'd want the most recent version of a style to be used, so if there's a conflict the newly inserted style should win.

site/docs/integrations/next.md
```diff
@@ -66,3 +66,13 @@ const withVanillaExtract = createVanillaExtractPlugin({
...
+
+## CSS load order issues
+
+There are some known issues with Next.js css load order and/or duplicate css.
+
+- When client-side navigating, new css is injected into the head, but unused css isn't removed. This means if route B has a component that applies some global css to a component also in route A, upon returning to route A, those styles from route B will still be applied.
+- When using App router, css can be duplicated which could cause unexpected overrides.
+
+[next.js #51030](https://github.com/vercel/next.js/issues/51030)
+[next.js #40080](https://github.com/vercel/next.js/issues/40080)
```
renrizzolo, line 78: @askoufis this is how I see it from my testing, but maybe I'm missing something.
There are quite a few issues in the next.js repo regarding css loading order etc., although it doesn't seem as bad with vanilla extract as it is with css modules for example. 

Here's a repro demonstrating the lingering css overrides. This seems to happen in every mode/router aside from pages router in production.
https://codesandbox.io/p/sandbox/next-51030-forked-phqdsr

To some degree it's kind of expected behavior? I guess we need to keep tabs on any future improvements on the next side.

huypham50:
Following up on this issue.

adrhumphreys:
I applied the fix in `@vanilla-extract/next-plugin` and it fixed the issue I was experiencing when navigating between pages. The CSS now is always applied and I don't need to refresh the page to load the CSS files.

It would be good awesome if we could look at getting that fix in as it would resolve https://github.com/vanilla-extract-css/vanilla-extract/issues/1152

askoufis:
Sorry for not following up on this. I'll try and get to this soon. Feel free to fix the conflicts if you get a chance @renrizzolo, otherwise I'll take a crack and try get this through.

askoufis:
@renrizzolo I'm liking where the PR is at, but it is quite big, so IMO it's worth de-risking these changes a bit by splitting things up into separate PRs.

There's probably 3 separate PRs here:
1. Updating the playwright version and tweaking the config, resulting in those snapshot file renames
2. Adding Next.js integration tests
3. Fixing Next.js pages router styles

Hoping it's not too hard to pick things apart. Happy to help out if you need.

huypham50:
@adrhumphreys All of my `vanilla-extract` deps are now latest and my css is still getting generated incorrectly locally

huypham50:
@adrhumphreys Just bumped all of my VE dependencies are now `latest` and the `css` is still getting generated incorrectly locally. Had to downgraded `@vanilla-extract/next-plugin` back to `2.1.3`. Looking forward to this fix!

huypham50:
@adrhumphreys Just bumped all of my VE dependencies to `latest` and the `css` is still getting generated incorrectly locally. Had to downgraded `@vanilla-extract/next-plugin` back to `2.1.3`. Looking forward to this fix!

askoufis:
@renrizzolo Should we close this in favour of https://github.com/vanilla-extract-css/vanilla-extract/pull/1365?

PR #1207, opened 2023-10-27
Add some missing simple pseudo selectors

CroModder:


CroModder:
PR created based on [this discussion](https://github.com/vanilla-extract-css/vanilla-extract/discussions/1201).

Details:
Didn't add the following pseudo elements because they don't have any browser compatibility data available:
- [::-moz-focus-inner](https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-focus-inner)
- [::cue-region](https://developer.mozilla.org/en-US/docs/Web/CSS/::cue-region)

Also, the following ones aren't added because they require attributes to be passed to them:
- [::highlight()](https://developer.mozilla.org/en-US/docs/Web/CSS/::highlight)
- [::part()](https://developer.mozilla.org/en-US/docs/Web/CSS/::part)
- [::slotted()](https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted)

askoufis (approved):

PR #1210, opened 2023-11-01
Update vite plugin

dagda1:
I've made a tentative PR to update the vite-plugin, which is now a few major versions below the current vite.

I'm keen to move to use Vite with remix, and this is a blocker.

Everything type checks and unit tests, but there are breaks in the playwright tests.  Some fail locally on the master branch also, and some snapshot tests where the color `white` is replaced with the hex value `#ffffff`.

I would love some input to get this over the line.

### DIFF
--- a/.changeset/empty-olives-agree.md
+++ b/.changeset/empty-olives-agree.md
@@ -1,0 +1,7 @@
+---
+'@vanilla-extract/integration': minor
+'@vanilla-extract/vite-plugin': minor
+'@vanilla-extract-private/test-helpers': minor
+---
+
+upgrade vite to 4.5.0
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "rollup-plugin-node-externals": "^5.0.0",
     "ts-node": "^10.0.0",
     "typescript": "^4.9.4",
-    "vitest": "^0.28.5"
+    "vitest": "^0.34.6"
   },
   "packageManager": "pnpm@7.9.5",
   "pnpm": {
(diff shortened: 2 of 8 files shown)


askoufis:
Thanks for the PR. I'm not sure if this change is really necessary. We intentionally keep our `vite` dev dependency low (v2) to maintain backwards compatibility with older vite versions (though we could probably drop v2/3 soon). The other versions are not pinned, so you should be able to use vite 4.5 with the vite plugin with no issue.

The peer dep range should include v4.5, so adding it to the range isn't really doing anything either, unless there was an accidental breaking change in 4.5 that we need to address in our integration.

askoufis:
Also, as far as remix integration goes, I'm not sure what the situation will be like with remix + vite - will remix continue to maintain its own integration with VE, or will the vite plugin be recommended instead?

dagda1:
@askoufis I don't know remix's plans.  I would imagine they move to vite but IDK.  Regarding upgrading to vite 4.5, I do get type errors with the return type of a vite plugin and vite 5 is in beta so I think you will need to do this switch at some stage.

askoufis:
> @askoufis I don't know remix's plans. I would imagine they move to vite but IDK. Regarding upgrading to vite 4.5, I do get type errors with the return type of a vite plugin and vite 5 is in beta so I think you will need to do this switch at some stage.

It has been confirmed that Remix are moving to Vite, I was more curious about the VE integration. I've confirmed that the intention moving forward is to just use the VE vite plugin with remix's vite plugin.

On the note of vite 4.5, if there is an error with the plugin, please raise an issue with a reproduction so we can better diagnose and fix it. As it stands, this PR doesn't really help towards fixing that issue.

askoufis:
@dagda1 I'm not seeing any type errors with regards to vite 4.5 + the VE vite plugin.

askoufis:
Superseded by #1240.

PR #1219, opened 2023-11-09
Pin `changesets/action` to `v1`

askoufis:
Not sure why but [`main` doesn't seem to work](https://github.com/vanilla-extract-css/vanilla-extract/actions/runs/6807432494/job/18510438611#step:6:13).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
         run: pnpm i
 
       - name: Create release PR or publish to npm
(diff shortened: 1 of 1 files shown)


mattcompiles (approved):

mrm007:
`changesets/action` pushes to a [`v1` tag on release](https://github.com/changesets/action/blob/main/scripts/release.js#L34-L40). The `main` branch doesn't have the compiled artifacts.

Issue #1228, opened 2023-11-13
debugId missing for custom identifiers


ufaubion-oracle:
### Describe the bug

I originally raised this issue on the Vanilla-Extract Discord server here. - https://discord.com/channels/885877446098964512/885877446891667529/1171464190200922152

**TL;DR** The `debugId` is `undefined` for custom identifiers.

For context, I ran into this issue when authoring a component library using Vanilla-Extract. To repo this issue, I created a very simple Next.js app to illustrate the issue. 

Here is the Vanilla-Extract styling for the app.

```ts
// src/app/page.css.ts
import { style } from '@vanilla-extract/css';

export const main = style({
  fontWeight: 'bold'
});
```

With the above styling I would expect the `debugId` passed into the custom identifier to be `'main'` however it is `undefined`.

```node
// next.config.js
const hash = require('@emotion/hash').default;
const [major] = (process.env.npm_package_version ?? '1.0.0').split('.');

const {
  createVanillaExtractPlugin
} = require('@vanilla-extract/next-plugin');
const withVanillaExtract = createVanillaExtractPlugin({
  identifiers: ({ debugId, hash: fileHash}) => `${debugId}__${hash(fileHash + major)}`
});

/** @type {import('next').NextConfig} */
const nextConfig = {};

module.exports = withVanillaExtract(nextConfig);
```

This results in the CSS class name `undefined__b6xo1n` instead of the expected `main__b6xo1n`.

### Reproduction

https://github.com/ufaubion-oracle/my-vanilla-app

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 632.83 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.17.1 - ~/.nvm/versions/node/v18.17.1/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 9.6.7 - ~/.nvm/versions/node/v18.17.1/bin/npm
    pnpm: 8.10.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Brave Browser: 119.1.60.114
    Chrome: 119.0.6045.123
    Edge: 116.0.1938.69
    Safari: 16.4
  npmPackages:
    @vanilla-extract/css: 1.14.0 => 1.14.0 
    @vanilla-extract/next-plugin: 2.3.2 => 2.3.2
```

[...]

PR #1229, opened 2023-11-13
Debug id for custom identifiers

ufaubion-oracle:
PR created based on https://github.com/vanilla-extract-css/vanilla-extract/issues/1228

Details:
I changed integrations transformer code to only skip "@vanilla-extract/babel-plugin-debug-ids" when the `identOption !== 'short'`.

### DIFF
--- a/packages/integration/src/transform.ts
+++ b/packages/integration/src/transform.ts
@@ -23,7 +23,7 @@ export const transformSync = ({
 }: TransformParams): string => {
   let code = source;
 
-  if (identOption === 'debug') {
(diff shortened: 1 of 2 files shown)


askoufis:
As discussed in the discord thread linked in the issue, I don't think that the best solution is just enabling it for everything except `short`. This would force the debug ID babel transform on consumers that want a custom identifier, but don't want debug IDs (e.g. in production).

I think ideally the custom identifier function would be separate to the debug identifier option. There could be a bit of nuance here if we want to avoid a breaking change in every integration.

PR #1231, opened 2023-11-15
Bump `postcss-load-config` to be able to load PostCSS configs defined as ES Modules

mrm007:
Discovered by @hi-ogawa here https://github.com/remix-run/remix/pull/8000#discussion_r1392440996

### DIFF
--- a/.changeset/tasty-ducks-try.md
+++ b/.changeset/tasty-ducks-try.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Bump `postcss-load-config` so to be able to load PostCSS configs defined as ES Modules
--- a/.changeset/tasty-ducks-try.md
+++ b/.changeset/tasty-ducks-try.md
@@ -2,4 +2,4 @@
 '@vanilla-extract/vite-plugin': patch
 ---
 
-Bump `postcss-load-config` so to be able to load PostCSS configs defined as ES Modules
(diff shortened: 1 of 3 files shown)


pnpm-lock.yaml
```diff
@@ -14671,6 +14671,24 @@ packages:
       ts-node: 10.9.1(@swc/core@1.2.112)(@types/node@16.11.10)(typescript@4.9.4)
       yaml: 1.10.2
 
+  /postcss-load-config@4.0.1(postcss@8.4.21)(ts-node@10.9.1):
```
mrm007, line 14674: We have v3 and v4 because we use an older version of `tailwind` on the site, which depends on v3.

```
site@0.0.1

devDependencies:
tailwindcss 2.2.19
└── postcss-load-config 3.1.0
```

.changeset/tasty-ducks-try.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Bump `postcss-load-config` so to be able to load PostCSS configs defined as ES Modules
```
askoufis, line 5:
```suggestion
Bump `postcss-load-config` to enable loading PostCSS configs defined as ES Modules
```

askoufis (approved):

.changeset/tasty-ducks-try.md
mrm007: 🧠💨

PR #1240, opened 2023-11-20
Add Vite 5 to peer dependencies

mrm007:
Supersedes #1210 since that PR made other breaking changes.

Also snuck in a fix for type errors observed when running locally, e.g. https://github.com/vanilla-extract-css/vanilla-extract/pull/1210#issuecomment-1790181830. It has no impact on the public API as the types are only used internally.


markdalgleish (approved):

markdalgleish:
It'd be good for this to go out at the same time as https://github.com/vanilla-extract-css/vanilla-extract/pull/1239 since Remix is already running Vite 5.

PR #1247, opened 2023-12-02
Fix style compilation for file paths containing webpack template strings

### LINKED ISSUE #1069: [@vanilla-extract/webpack-plugin] Error: Didn't get a result from child compiler
## Brief
```Error: Didn't get a result from child compiler``` will be thrown when ```@vanilla-extract/webpack-plugin``` is trying to parse a file under a folder which has a name with character "[]". 
This is a naming convention in some frameworks for dynamic routing, eg. [Next.js](https://beta.nextjs.org/docs/routing/defining-routes#example).
Thus for now, a vanilla-extract style file placed under a dynamic route folder like this won't work properly. 
## Reproduce Link
This is a webpack spa sample project which simulate the problem. 
https://github.com/KotoriK/vanilla-extract-webpack-resolve

### Reproduction

https://github.com/KotoriK/vanilla-extract-webpack-resolve

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
  Binaries:
    Node: 18.2.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 8.9.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1555.0), Chromium (112.0.1722.48)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vanilla-extract/css: ^1.11.0 => 1.11.0
    @vanilla-extract/webpack-plugin: ^2.2.0 => 2.2.0
    webpack: ^5.79.0 => 5.79.0
```

### Used Package Manager

pnpm

### Logs

```shell
ERROR in ./src/[id]/broken.css.js
Module build failed (from ./node_modules/.pnpm/@vanilla-extract+webpack-plugin@2.2.0_webpack@5.79.0/node_modules/@vanilla-extract/webpack-plugin/loader/dist/vanilla-extract-webpack-plugin-loader.cjs.js):
Error: Didn't get a result from child compiler
    at \home\vanilla-extract-webpack-resolve\node_modules\.pnpm\@vanilla-extract+webpack-plugin@2.2.0_webpack@5.79.0\node_modules\@vanilla-extract\webpack-plugin\dist\vanilla-extract-webpack-plugin.cjs.dev.js:167:25
    at finalCallback (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compiler.js:551:5)
    at \home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compiler.js:577:11
    at \home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compiler.js:1199:17
    at Hook.eval [as callAsync] (eval at create (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\tapable@2.2.1\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:6:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\tapable@2.2.1\node_modules\tapable\lib\Hook.js:18:14)
    at \home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compiler.js:1195:33
    at finalCallback (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compilation.js:2787:11)
    at \home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compilation.js:3092:11
    at Hook.eval [as callAsync] (eval at create (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\tapable@2.2.1\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:6:1)
    at Hook.CALL_ASYNC_DELEGATE [as _callAsync] (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\tapable@2.2.1\node_modules\tapable\lib\Hook.js:18:14)
    at \home\vanilla-extract-webpack-resolve\node_modules\.pnpm\webpack@5.79.0_webpack-cli@5.0.1\node_modules\webpack\lib\Compilation.js:3085:38
    at eval (eval at create (\home\vanilla-extract-webpack-resolve\node_modules\.pnpm\tapable@2.2.1\node_modules\tapable\lib\HookCodeFactory.js:33:10), <anonymous>:17:1)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at runNextTicks (node:internal/process/task_queues:64:3)
    at process.processImmediate (node:internal/timers:442:9)
 @ ./src/[id]/Broken.jsx 1:0-36 4:15-19
 @ ./src/index.js 3:0-35 7:48-54
```

### Validations

askoufis:
Fixes #1069.

You'd think this would be easily google-able, but apparently not. Eventually found out how to escape these template strings in [the webpack docs](https://webpack.js.org/configuration/output/#template-strings), but only after digging through [the source code of `TemplatedPathPlugin`](https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/TemplatedPathPlugin.js#L322) and working backwards from there.

~This feels like a rather naive fix to me. I'm not sure if changing this output filename could have any unintended side-effects. Also, I wanted to add a test case to one of the fixtures but wasn't sure which one. Would `low-level` be the best place for it?~
Using a more robust regex now, and added some unit tests and a webpack-specific fixture.


package.json
```diff
@@ -64,6 +64,7 @@
     }
   },
   "volta": {
-    "node": "20.9.0"
+    "node": "20.9.0",
+    "pnpm": "8.10.2"
```
askoufis, line 68: This was bugging me.

tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ESNEXT" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017','ES2018' or 'ESNEXT'. */,
     "module": "commonjs" /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */,
-    "lib": ["es2019", "es2017", "dom"],
+    "lib": ["es2021", "dom"],
```
askoufis, line 5: Required for `tsc` to recognize `replaceAll`. It was [added in node 15](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll#browser_compatibility), so IMO it's safe to use.

arturmuller:
This would be a very welcome fix!

packages/webpack-plugin/src/compiler.ts
```diff
@@ -64,9 +64,15 @@ function compileVanillaSource(
...
     );
     const compat = createCompat(isWebpack5);
-    // Child compiler will compile vanilla-extract files to be evaled during compilation
-    const outputOptions = { filename: loader.resourcePath };
 
+    // Escape webpack template strings in output files so they don't get replaced
+    const outputOptions = {
+      filename: loader.resourcePath
+        .replaceAll('[', '[\\')
+        .replaceAll(']', '\\]'),
```
mrm007, line 72: Related to the other comment:

```suggestion
      filename: loader.resourcePath.replaceAll(/(\[|\])/g, '\\$1'),
```

package.json
mrm007: This doesn't play nicely with folks who don't have pnpm support explicitly turned on (`VOLTA_FEATURE_PNPM=1`) https://github.com/volta-cli/volta/issues/737#issuecomment-1442296407

mrm007, line 68:
```suggestion
    "node": "20.9.0"
```

packages/webpack-plugin/src/compiler.ts
```diff
@@ -64,9 +64,15 @@ function compileVanillaSource(
       loader._compiler.webpack && loader._compiler.webpack.version,
     );
     const compat = createCompat(isWebpack5);
-    // Child compiler will compile vanilla-extract files to be evaled during compilation
-    const outputOptions = { filename: loader.resourcePath };
 
+    // Escape webpack template strings in output files so they don't get replaced
+    const outputOptions = {
+      filename: loader.resourcePath
+        .replaceAll('[', '[\\')
```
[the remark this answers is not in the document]
askoufis: Yeah will add a test in one of the fixtures. Not really sure which one though.

[\ is actually how it should be escaped, according to the docs. And it does work, I have tested it.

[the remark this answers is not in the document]
mrm007: That's very odd 🤯

packages/webpack-plugin/src/compiler.ts
```diff
@@ -64,9 +64,16 @@ function compileVanillaSource(
...
     const compat = createCompat(isWebpack5);
-    // Child compiler will compile vanilla-extract files to be evaled during compilation
-    const outputOptions = { filename: loader.resourcePath };
 
+    // Escape webpack template strings in output files so they don't get replaced
+    // Non-standard escape syntax, see the docs https://webpack.js.org/configuration/output/#template-strings
+    const outputOptions = {
+      filename: loader.resourcePath
+        .replaceAll('[', '[\\')
+        .replaceAll(']', '\\]'),
```
askoufis, line 73: Thinking about this some more, this would break a path that contains only a `[` or `]`, or some combination of non-matching square brackets. _OR_ if the path contains something similar to [a webpack template string](https://webpack.js.org/configuration/output/#template-strings) but isn't, e.g. `[myThing]`.

Maybe we could do something like `/\[(ext|id|...)\]/` for all the possible strings, which would be much more precise. Would be nice if there was a list of all template strings.

mrm007: I think we can check if the resourcePath matches `/\[.*\]/` and then do the replacements. If it hasn't been an issue so far, it likely won't be an issue in the future.

mrm007, line 73: What do you think about this? We probably want to test this in isolation, to make sure we're not missing any edge cases.

```suggestion
      filename: loader.resourcePath.replaceAll(
        /\[([^\[\]]*)\]/g,
        (_substring, match) => '[\\' + match + '\\]',
      ),
```

askoufis: Was about to suggest something similar. I think your one will match the inner pair of `[[]]`, but I doubt anyone would put that in a file path anyway.

mrm007: [You'd be surprised.](https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes#optional-catch-all-segments)

mrm007: I made it [this far](https://github.com/vanilla-extract-css/vanilla-extract/compare/webpack-square-brackets-fix...escape-webpack), but I'm not happy with how `[[blah]]` is escaped. Do you want to run some tests in Next.js?

askoufis: I'd have more confidence if we had integration tests. I've added some folders to one of the fixtures to test it. I can't quite get the escaping to work in all cases.

[foo] -> [\\foo\\] is correct
[[foo]] -> [[\\foo\\]] is correct
[[...foo]]] -> ??? No idea how to escape correctly.

askoufis: So I think we don't actually need to escape `[[...foo]]` at all, because it doesn't match [this regexp](https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/TemplatedPathPlugin.js#L19). So we only want to either escape single or double brackets, either with no contents, or some contents, but not `...`. 🥵

askoufis: This seems to work for all _four_ cases (we don't want to escape just a plain `[]` in a path). It's not quite perfect, as it won't handle `[[..foo]]` for example, but that's just incorrect syntax anyway.
```tsx
      filename: loader.resourcePath.replaceAll(
        /\[([^\[\]\.]+)\]/g,
        (_substring, match) => '[\\' + match + '\\]',
      )
```

askoufis:
I've added unit tests and a fixture for the escaping functionality.

packages/webpack-plugin/src/compiler.ts
```diff
@@ -55,6 +55,11 @@ function getRootCompilation(loader: LoaderContext) {
   return compilation;
 }
 
+const templateStringRegexp = /\[([^\[\]\.]+)\]/g;
+
+export const escapeWebpackTemplateString = (s: string) =>
+  s.replaceAll(templateStringRegexp, '[\\$1\\]');
```
askoufis, line 61: I went with the `$1` syntax as my assumption is that's less overhead compared to creating a replacement function.

tests/e2e/template-string-paths.playwright.ts
```diff
@@ -0,0 +1,43 @@
+import { expect } from '@playwright/test';
+import {
+  getStylesheet,
+  startFixture,
+  TestServer,
+} from '@vanilla-extract-private/test-helpers';
+
+import test from './fixture';
+import { webpack as testCases } from './testCases';
```
askoufis, line 9: Chose to restrict this fixture to only run fixtures using webpack.

mrm007 (approved):

PR #1250, opened 2023-12-07
fix(vite-plugin): prevent extra style module invalidations when using PostCSS

kosmotema:
Currently the CSS code that was processed with PostCSS (`cssSource`) is written to the `cssMap` cache:
https://github.com/vanilla-extract-css/vanilla-extract/blob/27552826716993d9923d2aa5f30513594ff3c1ee/packages/vite-plugin/src/index.ts#L241

However, the invalidation uses the source code that was not processed by PostCSS (`source`):
https://github.com/vanilla-extract-css/vanilla-extract/blob/27552826716993d9923d2aa5f30513594ff3c1ee/packages/vite-plugin/src/index.ts#L217

This results in unnecessary invalidation of those modules that may not have changed.

### DIFF
--- a/.changeset/great-geckos-marry.md
+++ b/.changeset/great-geckos-marry.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Prevent unnecessary module invalidations when using PostCSS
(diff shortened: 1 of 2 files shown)


askoufis (approved):
Thanks for the fix!

PR #1252, opened 2023-12-13
Upgrade ESBuild

Ethan-PageUp:
ESBuild version 0.17.6 was found to be causing issues with some of our components being rendered.

Upgrading to 0.19.9 to prevent anyone else from spending days debugging (and to stop us needing to delete the `node_modules` inside our vanilla-extract dependency 😆 )


BenBeattieHood:
`esbuild` added support for the TS `const` keyword in generic params [earlier this year](https://github.com/evanw/esbuild/issues/3021). This PR updates esbuild to a version that supports this, so that vanilla-extract can be run in repos using this newer syntax.

BenBeattieHood:
@Ethan-PageUp It might be more preferable to make esbuild a peerDependency in _only_ the integration and jest-integration plugins rather than upgrading the version in all packages here. This would allow consumers to set their own version of esbuild to use, rather than it being defined explicitly by vanilla-extract. See [here how esbuild-plugin does it](https://github.com/vanilla-extract-css/vanilla-extract/pull/1252/files#diff-f6b6a5a026d735011343e83065c5e06d8cdd17122630a5b4f256e2cf6fe47060R21-R28) for an example.

askoufis:
Thanks for the PR. We've been hesitant to upgrade the esbuild version too hastily as it has caused issues in the past. We'll take a look at upgrading the version if it seems appropriate. Support for `const` in generics would be nice.

packages/esbuild-plugin/package.json
```diff
@@ -18,14 +18,14 @@
     "@vanilla-extract/integration": "^6.2.0"
   },
   "peerDependencies": {
-    "esbuild": ">=0.17.6"
+    "esbuild": ">=0.19.9"
```
mrm007, line 21: This will cause a major version bump, which we don't really want. The plugin works with `>=0.17.6`, so we'll leave it as-is.

packages/integration/package.json
```diff
@@ -18,7 +18,7 @@
     "@babel/plugin-syntax-typescript": "^7.20.0",
     "@vanilla-extract/babel-plugin-debug-ids": "^1.0.4",
     "@vanilla-extract/css": "^1.14.0",
-    "esbuild": "0.17.6",
+    "esbuild": "npm:esbuild@~0.17.6 || ~0.18.0 || ~0.19.0",
```
mrm007, line 21: This syntax is not officially supported by npm, but I created a new project to [test the change](https://github.com/vanilla-extract-css/vanilla-extract/actions/runs/7649781260) and it worked.

<details><summary>Details</summary>

I tested with

```
  "dependencies": {
    "@vanilla-extract/css": "^1.14.0",
    "@vanilla-extract/esbuild-plugin": "^0.0.0-proxy-pr-1252-20240125041511",
    "esbuild": "^0.19.0"
  }
```

and

```
  "dependencies": {
    "@vanilla-extract/css": "^1.14.0",
    "@vanilla-extract/esbuild-plugin": "^0.0.0-proxy-pr-1252-20240125041511",
    "esbuild": "^0.17.6"
  }
```

</details>

mrm007: Update: it's [supported](https://github.com/npm/rfcs/blob/main/implemented/0001-package-aliases.md), but [not documented](https://github.com/npm/cli/issues/6839).

mrm007 (approved):

askoufis (approved):

mrm007:
Thanks @Ethan-PageUp! I made a few changes to reduce the impact and avoid breaking changes.

PR #1264, opened 2023-12-28
Use the new compiler in Vite plugin

### LINKED ISSUE #1211: Performance issues with Vite plugin
### Describe the bug

Hi,

I originally raised [an issue](https://github.com/mantinedev/mantine/issues/5184) in Mantine, but I don't think we can solve it from that end.

**TL;DR - it takes 30s to start the Vite dev server**
In the repo, it has 100 `.css.ts` files that imports from a common theme file. This theme file imports from `@mantine/core` and `@mantine/vanilla-extract`, but I don't think it's an issue with mantine.

I managed to reduce the start up time on my machine from 30s to 10s if I skip watching files in `node_modules`. I'm currently experimenting this with a pnpm patch as seen in this [commit](https://github.com/wenchonglee/ve-mantine-benchmark/commit/f73781ce980fbf9f3b8482ee0a6292e5394d8298#diff-cc160cdb1c61d54bfc5570f07e57484e9c36ca429561639124301328faf3f92a), but I'm not 100% sure it is sound. (10s is also not good enough for such a minimal repo)

I'm not very familiar with the internals of plugins and build tools, so I'd appreciate some insight into this and if it can/should be improved on Vanilla Extract's end.  

### Reproduction

https://github.com/wenchonglee/ve-mantine-benchmark

### System Info

```shell
System:
    OS: Windows 11 10.0.22621
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 22.55 GB / 31.95 GB
  Binaries:
    Node: 20.9.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.1.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.10.2 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (118.0.2088.76)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vanilla-extract/css: 1.13.0 => 1.13.0
    @vanilla-extract/dynamic: 2.0.3 => 2.0.3
    @vanilla-extract/integration: 6.2.2 => 6.2.2
    @vanilla-extract/recipes: 0.5.0 => 0.5.0
    @vanilla-extract/vite-plugin: 3.9.0 => 3.9.0
    vite: 4.5.0 => 4.5.0
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

mrm007:
Fixes #1211.

This PR rewrites the Vite plugin to use a new compiler powered by [vite-node](https://www.npmjs.com/package/vite-node), introduced by @markdalgleish in https://github.com/vanilla-extract-css/vanilla-extract/pull/1030.

**TL;DR: the new plugin is ~5x faster** (on average) than the current plugin. More details can be found in each changeset.

Pre-release versions (snapshots) of this plugin can be found here: [Github Actions](https://github.com/vanilla-extract-css/vanilla-extract/actions/workflows/snapshot.yml?query=branch%3Avite-plugin-new) (install them with `pnpm add @vanilla-extract/vite-plugin@vite-plugin-new`).

I've tested various snapshots in Remix using the [unstable Vite plugin](https://remix.run/docs/en/main/future/vite).

Using a fresh app created with [sku init](https://seek-oss.github.io/sku/#/./docs/getting-started?id=getting-started), with [Braid](https://seek-oss.github.io/braid-design-system/) and Remix, the results are below:

```jsonc
{
  "minimum compilation time (in ms)": {
    "current": 59,
    "new": 386
    // for small files with no dependencies, the new compiler adds some overhead
  },
  "maximum compilation time (in ms)": {
    "current": 6794,
    "new": 1422,
    "comparison": "20.93% of current"
  },
  "median": {
    "current": 2071,
    "new": 1160,
    "comparison": "55.96% of current"
  },
  "average": {
    "current": 2333,
    "new": 548,
    "comparison": "23.48% of current"
  }
}
```

I also tested in a [repo that had performance issues](https://github.com/vanilla-extract-css/vanilla-extract/issues/1211) with Mantine:

```json
{
  "minimum compilation time (in ms)": {
    "current": 6692,
    "new": 1395,
    "comparison": "20.84% of current"
  },
  "maximum compilation time (in ms)": {
    "current": 9478,
    "new": 1726,
    "comparison": "18.21% of current"
  },
  "median": {
    "current": 9117.5,
    "new": 1586,
    "comparison": "17.39% of current"
  },
  "average": {
    "current": 8731,
    "new": 1581,
    "comparison": "18.1% of current"
  }
}
```


mrm007:
All credit to @markdalgleish on this one. I just made a few small adjustments to already working code.

packages/vite-plugin/src/index.ts
```diff
@@ -69,100 +88,105 @@ export function vanillaExtractPlugin({
...
-      } else {
-        ssr = ssrParam?.ssr;
+        for (const file of watchFiles) {
+          // In start mode, we need to prevent the file from rewatching itself.
+          // If it's a `build --watch`, it needs to watch everything.
+          if (
+            config.command === 'build' ||
+            normalizePath(file) !== absoluteId
+          ) {
+            this.addWatchFile(file);
```
mrm007, line 173: Might be worth considering skipping `node_modules`, like the workaround linked in https://github.com/vanilla-extract-css/vanilla-extract/issues/1211:

https://github.com/wenchonglee/ve-mantine-benchmark/blob/main/patches/@vanilla-extract__vite-plugin@3.9.0.patch

packages/integration/src/compiler.ts
```diff
@@ -17,22 +17,27 @@ type Composition = Parameters<Adapter['registerComposition']>[0];
...
 
   for (const moduleNode of queue) {
+    if (moduleNode.id?.includes('@vanilla-extract/')) {
+      continue;
+    }
+
     const relativePath = moduleNode.id && relative(root, moduleNode.id);
 
-    if (relativePath) {
+    if (relativePath && cssFileFilter.test(relativePath)) {
```
mrm007, line 31: This reduces the amount of work for tree traversal significantly. It's especially noticeable when used with libraries like Mantine i.e. https://github.com/vanilla-extract-css/vanilla-extract/issues/1211

packages/integration/src/compiler.ts
```diff
@@ -135,6 +142,28 @@ const createViteServer = async ({
   };
 };
 
+class NormalizedMap<V> extends Map<string, V> {
```
mrm007, line 145: This is the most significant change in this file. Paths are now normalised so the cache can return a value whether you provide an absolute or relative path key.

packages/vite-plugin/src/index.ts
```diff
@@ -10,53 +9,99 @@ import {
...
+
+const fileIdToVirtualId = (id: string) => `${id}${virtualExtCss}`;
+const virtualIdToFileId = (virtualId: string) =>
+  virtualId.slice(0, -virtualExtCss.length);
 
 interface Options {
   identifiers?: IdentifierOption;
-  emitCssInSsr?: boolean;
   esbuildOptions?: CompileOptions['esbuildOptions'];
+  mode?: 'transform' | 'emitCss';
```
mrm007, line 26: The presence of `esbuildOptions` will switch to the esbuild-based compiler. And in `mode: 'transform'` we don't need the new compiler.

tests/e2e/features.playwright.ts-snapshots/features-vite-next--production.css
mrm007, line 1: This is exactly the same output as the current compiler:

https://github.com/vanilla-extract-css/vanilla-extract/blob/3a2429cad2e64606284cb62231db8b328f4e1257/tests/e2e/features.playwright.ts-snapshots/features-vite--production.css#L1-L7

tests/e2e/thirdparty.playwright.ts-snapshots/thirdparty-vite-next--production.css
mrm007, line 1: Works the same as the current compiler with complex file scopes i.e. from third-party code (Braid)

https://github.com/vanilla-extract-css/vanilla-extract/blob/973e526cacd568d8e0986090e81fecf584a28b31/tests/e2e/thirdparty.playwright.ts-snapshots/thirdparty-vite--production.css#L1-L12

packages/vite-plugin/src/index.ts
```diff
@@ -1,62 +1,98 @@
...
+
+    const { moduleGraph } = server;
+    const modules = Array.from(moduleGraph.getModulesByFile(absoluteId) || []);
+
+    for (const module of modules) {
+      moduleGraph.invalidateModule(module);
+
+      // Vite uses this timestamp to add `?t=` query string automatically for HMR.
+      module.lastHMRTimestamp = module.lastInvalidationTimestamp || Date.now();
+    }
```
askoufis, line 65: You could keep this as a `Set` and avoid the extra iteration from `Array.from`:

```suggestion
    const modules = moduleGraph.getModulesByFile(absoluteId);
    
    if (modules) { 
      for (const module of modules) {
        moduleGraph.invalidateModule(module);
  
        // Vite uses this timestamp to add `?t=` query string automatically for HMR.
        module.lastHMRTimestamp = module.lastInvalidationTimestamp || Date.now();
      }
    }
```

packages/vite-plugin/src/index.ts
```diff
@@ -1,62 +1,98 @@
...
+
+    const filesToWatch = [...files].filter(
+      (file) => !file.includes('node_modules'),
+    );
+
+    for (const file of filesToWatch) {
+      if (normalizePath(file) !== fromId) {
+        this.addWatchFile(file);
+      }
+    }
```
askoufis, line 86: Likely not performance-critical code, but you could just iterate over `files` once with a `for of` and put the `node_modules` check inside the loop, avoiding the creation of an array and the overhead from `filter`:

```suggestion
    for (const file of files) {
      if (!file.includes('node_modules') && normalizePath(file) !== fromId) {
        this.addWatchFile(file);
      }
    }
```

packages/integration/src/compiler.ts
```diff
@@ -17,22 +17,27 @@ type Composition = Parameters<Adapter['registerComposition']>[0];
 const globalAdapterIdentifier = '__vanilla_globalCssAdapter__';
 
 const scanModule = (entryModule: ModuleNode, root: string) => {
-  const queue = [entryModule];
+  const queue = new Set([entryModule]);
```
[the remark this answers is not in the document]
mrm007: I had the same thought as you. However, the [JavaScript `Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) keeps track of the order in which items were inserted:

> You can iterate through the elements of a set in insertion order. The insertion order corresponds to the order in which each element was inserted into the set by the `add()` method successfully (that is, there wasn't an identical element already in the set when `add()` was called).

packages/integration/src/compiler.ts
mrm007: No, because `queue` is still updated with every dependency and its dependencies. We just don't add it to `cssDeps`, which is returned from the function. The items in `cssDeps` are used to generate imports on line 351: https://github.com/vanilla-extract-css/vanilla-extract/blob/55a18052025aa9198e9e635c747301e734c0c8d7/packages/integration/src/compiler.ts#L350-L352

packages/vite-plugin/src/index.ts
mrm007: Good suggestion 👍 This code went through a few stages of refactoring.

mattcompiles (approved):

PR #1273, opened 2024-01-12
feat(integration): Allow `conditions` option to pass through to esbuild

lars-berger:
Modifying `conditions` is needed when you need to customize the import paths used by esbuild. [Relevant docs](https://esbuild.github.io/api/#how-conditions-work)

For example, I've got a UI library that uses SolidJS and it exports built JS and CSS. This is what the `"exports"` looks like in its `package.json`:
```json
// package.json
"exports": {
  ".": {
    "solid": "./dist/index.jsx",
    "import": {
      "types": "./dist/index.d.ts",
      "default": "./dist/index.js"
    }
  },
  "./styles": "./dist/index.css"
},
```
I'm then consuming this library along with Vanilla Extract in an Astro project. If I then do any imports of the UI library from a `.css.ts` file, the wrong import path is used. In the case above, the `"solid"` import condition needs to be used (due to how SSR needs a separate entrypoint with frameworks like Astro).

LMK if any changes are needed 👍


mrm007 (approved):

.changeset/beige-bobcats-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/integration': minor
+---
+
+Allow conditions option to be passed through to esbuild
```
askoufis, line 5:
```suggestion
Allow `conditions` option to be passed through to esbuild
```

askoufis (approved):

mrm007:
Thanks @lars-berger!

PR #1283, opened 2024-01-23
docs: add Astro integration page

JoshuaKGoldberg:
Adds an Astro page per discussion in https://github.com/vanilla-extract-css/vanilla-extract/discussions/678#discussioncomment-7496065 -> #1266.

@bejzik8 answered flawlessly so this shamelessly steals that explanation 😄. Hence the co-author attribute below. 

Co-authored-by: Mirko Basic <bejzik8@gmail.com>


site/docs/integrations/astro.md
```diff
@@ -0,0 +1,60 @@
...
+```
+
+```astro
+---
+// Button.astro
+import { button } from './button.css'
+---
+
+<button class={button}>Click Me!</button>
+```
```
JoshuaKGoldberg, line 56: This shows up as unstyled for me. 😬 
image

I took a peek at the `CompiledCode` and `SyntaxHighlighter` system but didn't want to make too many changes. I was scared. What would you all advise be done here?

FWIW the progress I made was trying to import and register the `markup` syntax highlighting language for Astro, but that didn't get all the way:

```ts
// @ts-ignore
import markup from 'react-syntax-highlighter/dist/esm/languages/prism/markup';
SyntaxHighlighter.registerLanguage('astro', markup);
```

mrm007: After a bit of digging, it looks like it will take a lot of work to have syntax highlighting for Astro files. Given it's only used on this page, I think we're ok with not having it.

site/docs/integrations/astro.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Astro
+parent: integrations
+---
+
+# Astro
+
+A plugin for integrating vanilla-extract with [Astro](https://astro.build).
```
askoufis, line 8: This makes it sound like there's a separate astro plugin, but that's not the case. Just reworded it a bit.

```suggestion
Vanilla Extract integrates with [Astro](https://astro.build) via the Vite plugin, rather than a separate Astro plugin.
```

site/docs/integrations/astro.md
```diff
@@ -0,0 +1,65 @@
...
+  borderRadius: '0.25rem',
+  color: 'white',
+  background: '#333'
+});
+```
+
+And now you can reference styles in your Astro component:
+
+```tsx
+//  Button.astro
```
mrm007, line 52: Bit of a hack, but it's not stupid if it works 😅 

Button.astro

askoufis (approved):

mrm007:
Thank you @JoshuaKGoldberg!

Issue #1280, opened 2024-01-22
Sandbox in README is broken


askoufis:
### Describe the bug

The sandbox linked in the README doesn't work. The sandbox is based off [the webpack example in the repo](https://github.com/vanilla-extract-css/vanilla-extract/tree/837774c13514c94f46796a44cfa9da8451d80974/examples/webpack-react). Probably just need to [update `webpack-dev-server` to v4 and tweak a setting](https://stackoverflow.com/a/43647767).

Potentially worth swapping the codesanbox to point to a vite example instead, just for the speed and ease of use.

### Reproduction

Navigate to https://codesandbox.io/s/github/vanilla-extract-css/vanilla-extract/tree/master/examples/webpack-react?file=/src/App.css.ts. Web page says `Invalid Host header`.

### System Info

```shell
N/A
```

### Used Package Manager

npm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

PR #1287, opened 2024-01-25
Update webpack and webpack-related deps

askoufis:
Originally just wanted to upgrade `webpack-dev-server` in the `webpack-react` example to address #1280, but that snowballed into upgrading some other deps too. Kept it scoped to just webpack-related stuff (mostly loaders) that were at least a major version behind.

Will followup afterwards with some more dep upgrades to the webpack example.

Followed [this migration guide](https://github.com/webpack/webpack-dev-server/blob/48f56d57775dc4cf78bfcf51c3761adf852b0310/migration-v4.md) for `webpack-dev-server` v3 -> v4.


packages/webpack-plugin/src/compiler.ts
```diff
@@ -81,7 +81,7 @@ function compileVanillaSource(
     );
     const ExternalsPlugin = compat.getExternalsPlugin(loader._compiler);
 
-    new NodeTemplatePlugin(outputOptions).apply(childCompiler);
+    new NodeTemplatePlugin().apply(childCompiler);
```
askoufis, line 84: `outputOptions` contained just a `filename` property. Looking at both the webpack 4 and 5 docs, this was never a valid property for this plugin. It was causing a type error, so I removed it.

test-helpers/src/startFixture/webpack.ts
```diff
@@ -96,25 +95,32 @@ export const startWebpackFixture = (
     });
     const compiler = webpack(config);
 
-    // @ts-expect-error - webpack version mismatch
-    const server = new WDS(compiler, {
-      hot,
-      stats: logLevel,
-      noInfo: true,
```
askoufis, line 103: These were removed in `webpack-dev-server` v4.

mrm007: Pushed [`86f7733` (#1287)](https://github.com/vanilla-extract-css/vanilla-extract/pull/1287/commits/86f773361151a15073e1dcc552580636b7d2462c) to restore the logging level. Feel free to drop the commit if you disagree.

mrm007 (approved):

test-helpers/src/startFixture/webpack.ts
askoufis: LGTM. I tried with something similar but kept getting type errors. Honestly not too concerned with the fixture log output, but thanks 👍.

PR #1290, opened 2024-01-26
Allow rollup@4.x.x as peer dependency

smholsen:
Spinning up this PR from https://github.com/vanilla-extract-css/vanilla-extract/discussions/1289.
@mrm007 I don't know exactly what to test/verify for this change though. 
Heres the breaking changes from 3.x.x to 4.x.x: https://github.com/rollup/rollup/releases/tag/v4.0.0

I've done a quick scan, and it seems theres a few references to `preserveModules` in a test, which I am unsure if might cause errors down the line if the peer dependency is on `4.x.x` due to the breaking change note:
```
inlineDynamicImports, manualChunks and preserveModules have been removed on input option level: Please use the corresponding output options of the same names (https://github.com/rollup/rollup/pull/5143)
```

Other than that it looks to me this should be safe. 
Anything else we need to verify for this?

### DIFF
--- a/.changeset/clean-wolves-attend.md
+++ b/.changeset/clean-wolves-attend.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/rollup-plugin': minor
+---
+
+Allow rollup@4.x.x as peer dependency
--- a/.changeset/clean-wolves-attend.md
+++ b/.changeset/clean-wolves-attend.md
@@ -2,4 +2,5 @@
 '@vanilla-extract/rollup-plugin': minor
 ---
 
-Allow rollup@4.x.x as peer dependency
+PR: https://github.com/vanilla-extract-css/vanilla-extract/pull/1290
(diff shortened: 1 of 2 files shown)


.changeset/clean-wolves-attend.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/rollup-plugin': minor
```
mrm007, line 2: Looks like we treat dependency bumps as `patch`, sorry about the confusion.

mrm007 (approved):
Thanks @smholsen!

mrm007:
I thought it would be a good idea to bump `devDependencies` for the Rollup plugin. Unfortunately, [this line](https://github.com/vanilla-extract-css/vanilla-extract/blob/a0fd6234bfe8294639d039bd25ec7135cc3445aa/packages/rollup-plugin/src/index.ts#L104) causes a TypeScript error and the sourcemaps test failed in unexpected ways, so I'll do that in a separate PR.

mrm007:
@graup do you see any issues with this PR?

smholsen:
Thanks for the quick follow up here guys! 🙌

PR #1308, opened 2024-02-03
Don't pass Remix Vite plugin to the vite-node compiler

mrm007:
Similar to #1300, we need to filter out Remix before passing plugins to the compiler. 
[Remix throws an error](https://github.com/remix-run/remix/blob/f0688dd49f41e4bcdf8bc689a8380c2e0ee5c760/packages/remix-dev/vite/plugin.ts#L966-L970) if it's loaded without a config file, like we do when we initialise the vite-node compiler.

This has been an issue with version 4.x of our Vite plugin since the very beginning. https://github.com/remix-run/remix/pull/7990 was merged long before we released [v4.0.0](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract/vite-plugin%404.0.0).

### DIFF
--- a/.changeset/vite-plugin-remix.md
+++ b/.changeset/vite-plugin-remix.md
@@ -4,4 +4,4 @@
 
 Don't pass Remix Vite plugin to the vite-node compiler
 
-Remix throws an error if it's loaded without a config file, like we do when we initialise the vite-node compiler.
+Remix throws an error if it's loaded without a config file, which is what we do when we initialise the vite-node compiler.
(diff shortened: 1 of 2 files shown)


askoufis (approved):

.changeset/vite-plugin-remix.md
```diff
@@ -0,0 +1,7 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Don't pass Remix Vite plugin to the vite-node compiler
+
+Remix throws an error if it's loaded without a config file, like we do when we initialise the vite-node compiler.
```
mrm007, line 7:
```suggestion
Remix throws an error if it's loaded without a config file, which is what we do when we initialise the vite-node compiler.
```

PR #1314, opened 2024-02-06
Load a new Vite config object for the vite-node compiler

### LINKED ISSUE #1313: HMR broken with @vanilla-extract/vite-plugin v4.0.1 and later
### Describe the bug

After upgrading from v3.9.5 to v4.0.3 HMR broke in our setup. Seem like v4.0.0 actually works alright and it broke in v4.0.1. Downgrading to v4.0.0 fixes the problem.

### Reproduction

https://github.com/jakst/vanilla-extract-hmr

1. Clone repo
2. run `pnpm install`
3. Start dev server
4. edit `src/styles.css.ts`
5. Observe HMR not functioning, and page fully reloading
6. Downgrade @vanilla-extract/vite-plugin to v4.0.0 (`pnpm add -E @vanilla-extract/vite-plugin@4.0.0`) 
7. Repeat steps 3 and 4
8. Observe HMR working properly

### System Info

```shell
System:
  OS: macOS 14.2
  CPU: (12) arm64 Apple M3 Pro
  Memory: 822.20 MB / 36.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.10.0 - ~/.proto/shims/node
  Yarn: 4.0.2 - ~/.proto/shims/yarn
  npm: 10.2.3 - ~/.proto/shims/npm
  pnpm: 8.15.1 - ~/.proto/shims/pnpm
  bun: 1.0.25 - ~/.proto/shims/bun
  Watchman: 2024.01.22.00 - /opt/homebrew/bin/watchman
Browsers:
  Brave Browser: 120.1.61.109
  Chrome: 121.0.6167.139
  Edge: 121.0.2277.98
  Safari: 17.2
npmPackages:
  @vanilla-extract/css: 1.14.1 => 1.14.1
  @vanilla-extract/recipes: 0.5.1 => 0.5.1
  @vanilla-extract/vite-plugin: 4.0.3 => 4.0.3
  vite: 5.0.12 => 5.0.12
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

mrm007:
Fixes #1313.

This is similar to the fix provided by @hi-ogawa in https://github.com/remix-run/remix/pull/7990.
I believe this is a better fix than #1300 ~~and #1308~~ (see note below). 

Now that we load a new copy of the Vite config, we don't have to worry about filtering plugins before passing them to the vite-node compiler, since they don't share any state with the app's Vite config. 

Edit: We still need to filter out Remix because the plugins are incompatible **with and without** a config file.

### DIFF
--- a/.changeset/integration-plugins.md
+++ b/.changeset/integration-plugins.md
@@ -1,0 +1,7 @@
+---
+'@vanilla-extract/integration': minor
+---
+
+A Vite config object can be passed to the vite-node compiler via the `viteConfig` option
+
+`viteResolve` and `vitePlugins` options are deprecated and will be removed in a future version.
--- a/.changeset/vite-plugin-plugins.md
+++ b/.changeset/vite-plugin-plugins.md
@@ -1,0 +1,7 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Resolve and pass a new copy of the Vite config to the vite-node compiler
+
+Previously, we were passing the same Vite config object to the vite-node compiler. This was causing compatibility issues with other plugins, such as Vitest and Remix.
--- a/packages/integration/src/compiler.ts
+++ b/packages/integration/src/compiler.ts
@@ -1,7 +1,8 @@
+import assert from 'assert';
 import { join, isAbsolute } from 'path';
 import type { Adapter } from '@vanilla-extract/css';
 import { transformCss } from '@vanilla-extract/css/transformCss';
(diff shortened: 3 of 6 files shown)


mrm007:
We still need to do something about Remix because [we start Vite without a config file](https://github.com/vanilla-extract-css/vanilla-extract/blob/1239750f810dffbdeb45b7558b4b7aec438623d5/packages/integration/src/compiler.ts#L58-L60).

askoufis (approved):

PR #1316, opened 2024-02-07
docs: fix filename in the code

suu3:
Hi there! 
While reading the document, I noticed that the file names in the code were different. so I just made a small change. 😃


askoufis (approved):
Thanks!

PR #1327, opened 2024-02-16
turn off file watchers

evertbouw:
I thought I was running into this problem https://github.com/laravel/vite-plugin/issues/152 but it was happening in prod builds, so none of the suggestions worked. I traced where the watchers were coming from and it was vanilla-extract. I patched in `watch: null` in my project and now CI/CD builds are working again.


mrm007:
Thanks @evertbouw!

Is your project public or do you have a way to reproduce this? I might write an integration test for the compiler if it's something that can be automated.

.changeset/moody-balloons-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/integration': patch
+---
+
+turn off file watchers
```
mrm007, line 5:
```suggestion
Turn off file watchers in the vite-node compiler
```

evertbouw:
@mrm007 Sorry it's proprietary. I'm guessing if you create more files than the system filewachter limit you should run into this, doesn't seems to happen in macos locally but it did on the linux CI/CD box.

askoufis:
I wonder if perhaps we're creating multiple compilers, instead of re-using the existing one. 
@evertbouw Could you try out this snapshot version of the vite plugin from #1397? `0.0.0-fix-watch-mode-20240429115131`.

evertbouw:
@askoufis that did not fix the problem.

frankwallis:
The issue can be seen by installing this tool on Linux: https://github.com/mikesart/inotify-info

Running `vite build` with the Vanilla Extract Vite Plugin configured will cause file watchers to be created:
image

Removing the Vanilla Extract Vite Plugin from the configuration means the file watchers do not get created.

askoufis:
> The issue can be seen by installing this tool on Linux: https://github.com/mikesart/inotify-info
> 
> Running `vite build` with the Vanilla Extract Vite Plugin configured will cause file watchers to be created: image
> 
> Removing the Vanilla Extract Vite Plugin from the configuration means the file watchers do not get created.

Thanks for sharing this tool, it definitely shows that `vite build` results in some files being watched. Indeed, setting `watch: null` does prevent this from happening, but it also breaks Vanilla Extract's HMR. Specifically, we use the file watcher to invalidate modules when they change: https://github.com/vanilla-extract-css/vanilla-extract/blob/221f07f83844a238604a00e4c2d3c75eb7f087ba/packages/compiler/src/compiler.ts#L194.

However, after looking at [a PR in React Router](https://github.com/remix-run/remix/pull/10510) where they turned off their file watcher, maybe we could use the existing vite dev server's file watcher to invalidate the vite node runner's module cache, and then disable the vite node server's watcher. I'll test it out.

PR #1337, opened 2024-03-03
Handle option field that doesn't exist in config

aspirisen:
if you pass object with fields that doesn't exist in the variants (i.e. just pass props as is) - runtime function will fall with exception.

### DIFF
--- a/packages/recipes/src/createRuntimeFn.ts
+++ b/packages/recipes/src/createRuntimeFn.ts
@@ -45,7 +45,7 @@ export const createRuntimeFn = <Variants extends VariantGroups>(
 
         const selectionClassName =
           // @ts-expect-error
-          config.variantClassNames[variantName][selection];
+          config.variantClassNames?.[variantName]?.[selection];
 
         if (selectionClassName) {
           className += ' ' + selectionClassName;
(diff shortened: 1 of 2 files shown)


askoufis:
While this change makes sense in theory, I'm keen to understand when this would occur/use cases for this. I suppose if you were taking in recipe options at runtime, and were not validating them, or perhaps ignoring a type mismatch for the recipe function?

aspirisen:
@askoufis the types are correct, but the runtime function thinks that all fields in the argument object are variants.
For example you have to use props destructure in your component to use only know recipe fields. It would be much easier to pass props argument as is to recipe runtime function

For example:
```
// VE file
export type TestStyles = RecipeVariants<typeof test>

export const test = recipe({
    variants: {
        $variant: {
            primary: [],
            'primary-light': [],
            outline: [],

            error: [],
        },

        $round: {
            true: [],
        },

        $size: {
            small: {},
            medium: [],
            large: [],
        },

        // more variants
    },
})

// Component file

export interface ComponentProps extends TestStyles {
    // some more props
}

function Component(props: PropsWithChildren<ComponentProps>) {
    // here you will have runtime error "Cannot read properties of undefined (reading '#<Object>')" because of children field
    // but there are no TS error (what is expected behavior)
    return <div className={test(props)}>{props.children}</div>
}
```

PR #1340, opened 2024-03-04
Fix link to function serialization API in the docs

weelink:
On the pages explaining recipes and sprinkles, the link to the function serialization API results in a 404, because it points to `../../api` instead of `../api`.


askoufis (approved):
Thanks for the PR @weelink!

PR #1341, opened 2024-03-04
Improve type hint of recipe's runtime fn

levrik:
As per https://www.totaltypescript.com/concepts/the-prettify-helper

**Before:**

image

**After:**

image


askoufis (approved):
Thanks for this @levrik!

PR #1345, opened 2024-03-05
Fix the absolute path issue in the windows

owjs3901:


askoufis:
Thanks for the PR @owjs3901. I didn't see it in time before raising my PR #1346 just a few minutes later, but I'm going to close this one in favour of mine as I don't think your code handles monorepo file paths.

owjs3901:
@askoufis 
I'm not good at English, so I left a comment late.
There is no problem with monorepo as well.
The problem only occurs in Windows.

owjs3901:
There is a better way! thank you(https://github.com/vanilla-extract-css/vanilla-extract/pull/1346)

PR #1349, opened 2024-03-05
fix(e2e): next snapshot tests

renrizzolo:
- add `next-pages-router` and `next-app-router` fixtures. This allows snapshot testing on the next 12 pages router and the next 13 app router.
- add build step for the next apps to the test:playwright command. If this is annoying for development we could add a separate CI version of this command that is called only in the validate workflow.
- add script to copy the next plugin dist to the next fixtures. Because we're running multiple versions of next in parallel, we need separate versions of the plugin too, due to how `mini-css-extract` plugin works/is imported.

#### How it works
The next apps import the html from existing fixtures, and serves it at `/[fixture]`. The route is tested against that fixture's own snapshot.
The next apps are served using the `http` module and next's built in [server](https://nextjs.org/docs/pages/building-your-application/configuring/custom-server) mode. *app router production test is run as a static export with Vite due to issues I ran into with next 13 `distDir` setting not sticking, so dev build would overwrite prod or vice versa.

NB: I only added the `recipes` fixture for now. We can add more in future, but most will require the `next-style-loader` fix from https://github.com/vanilla-extract-css/vanilla-extract/pull/1203 to pass.


fixtures/next-pages-router/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@fixtures/sprinkles": "*",
+    "@vanilla-extract/css": "*",
+    "@vanilla-extract/recipes": "*",
+    "@vanilla-extract/sprinkles": "*",
+    "next": "12.3.4",
+    "react": "npm:react@^17.0.2",
+    "react-dom": "npm:react-dom@^17.0.2"
+  },
+  "devDependencies": {
+    "@types/react": "npm:@types/react@^17",
```
askoufis, line 27: The rest of the repo is on react 18 as of https://github.com/vanilla-extract-css/vanilla-extract/pull/1326. Is there a reason we need react 17 specifically, or was this leftover from splitting off from your original PR? Does next v12 not support react 18?

askoufis (commented):
Thanks so much for splitting this out. Just some minor comments from me.

fixtures/next-pages-router/package.json
renrizzolo: yeah this must have been leftover from before everything was moved to react 18, will change

.changeset/small-carrots-hang.md
```diff
@@ -0,0 +1,9 @@
+---
+'@fixtures/next-app-router': patch
+'@fixtures/next-pages-router': patch
+'@fixtures/recipes': patch
+'@vanilla-extract-private/test-helpers': patch
+'@vanilla-extract-private/tests': patch
```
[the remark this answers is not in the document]
renrizzolo: Ah, I see, thanks will remove

askoufis (approved):

PR #1353, opened 2024-03-08
vite: Create compiler during `buildStart` instead of `configResolved`

himself65:
Related: https://github.com/dai-shi/waku/pull/559

### DIFF
--- a/.changeset/hungry-berries-deliver.md
+++ b/.changeset/hungry-berries-deliver.md
@@ -2,4 +2,4 @@
 "@vanilla-extract/vite-plugin": patch
 ---
 
-Fix a bug that could cause a `vite-node` compiler to be created but never destroyed
+Move `vite-node` compiler creation to a more appropriate plugin hook to ensure correct cleanup of resources
--- a/.changeset/hungry-berries-deliver.md
+++ b/.changeset/hungry-berries-deliver.md
@@ -1,0 +1,5 @@
+---
+"@vanilla-extract/vite-plugin": patch
+---
+
+Fix a bug that could cause a `vite-node` compiler to be created but never destroyed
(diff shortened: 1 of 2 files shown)


askoufis:
@himself65 Thanks for the PR. Could you provide some more context on what the issue is and how this change fixes it?

I can sort of see how `createCompiler` is better placed in the `buildStart()` hook, but I don't think the `!compiler` check is achieving anything here. Were you observing the `compiler` being recreated, leading to a bug?

himself65:
> @himself65 Thanks for the PR. Could you provide some more context on what the issue is and how this change fixes it?
> 
> I can sort of see how `createCompiler` is better placed in the `buildStart()` hook, but I don't think the `!compiler` check is achieving anything here. Were you observing the `compiler` being recreated, leading to a bug?

Yeah, the issue is when we call `resolveConfig` in node.js, `buildEnd` won't be called in this case, so that the node.js process cannot exit successfully

himself65:
> `!compiler` check

Oh, that is my mistake

askoufis:
> > @himself65 Thanks for the PR. Could you provide some more context on what the issue is and how this change fixes it?
> > I can sort of see how `createCompiler` is better placed in the `buildStart()` hook, but I don't think the `!compiler` check is achieving anything here. Were you observing the `compiler` being recreated, leading to a bug?
> 
> Yeah, the issue is when we call [`resolveConfig`](https://github.com/dai-shi/waku/blob/2852bd169991f95b468101177efc651136a65f87/packages/waku/src/lib/builder/build.ts#L600-L602) in node.js, `buildEnd` won't be called in this case, so that the node.js process cannot exit successfully

Oh I understand now. So `resolveConfig` calls the `configResolved` hooks of plugins, so a compiler is created but never destroyed. Makes sense to me. I'll add a changeset.

askoufis (approved):

askoufis:
@himself65 This has been released https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract/vite-plugin%404.0.5

PR #1365, opened 2024-03-19
fix: Next plugin - pages router style loading

renrizzolo:
#### Update next plugin
- always output css
- fixes a bug with composed selector global css class not being applied in SSR with pages router
- use `next-style-loader` for pages router in development (as per original comment, next-style-loader can mess up css order in development mode, but this is surely better than no css at all.)
- fixes a bug with new page styles not being added to the document when client side routing https://github.com/vanilla-extract-css/vanilla-extract/issues/1152
- remove appDir config check

#### Add more fixtures to next e2e tests
Didn't want to add them all, as it would increase the workflow time quite a bit.
I think the ones I've added are enough to validate the fixes.

 Not sure if we want to change all the fixtures to have a separate `html.ts` for consistency.

### DIFF
--- a/.changeset/honest-pugs-tap.md
+++ b/.changeset/honest-pugs-tap.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/next-plugin': minor
+---
+
+Pages router: use next-style-loader in dev, output css in ssr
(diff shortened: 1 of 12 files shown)


askoufis (approved):
Thanks for all your effort on this. I think https://github.com/vanilla-extract-css/vanilla-extract/pull/1357 will give us a bit more overhead to add more fixtures if we want to in the future.

Wouldn't worry about refactoring the other fixtures for now.

renrizzolo:
No worries!

PR #1368, opened 2024-03-21
Fix `walkObject` to work with module namespace objects

### LINKED ISSUE #1360: 4.x: Uncaught TypeError: e.constructor is not a function
### Describe the bug

I've been running `@vanilla-extract/vite-plugin@3.9.5` for quite some time in a Vue project, however it remains impossible to upgrade to v4 due to the following issue:

- if a add a file called `theme.css.ts` an error is thrown during `vite build`

```
> vite build

vite v5.1.6 building for production...
✓ 15 modules transformed.
x Build failed in 473ms
error during build:
walkObject@file:///home/projects/vitejs-vite-ogfgnu/node_modules/@vanilla-extract/private/dist/vanilla-extract-private.esm.js:27:19
createThemeContract@file:///home/projects/vitejs-vite-ogfgnu/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.esm.js:295:32
createGlobalTheme@file:///home/projects/vitejs-vite-ogfgnu/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.esm.js:314:38
@file:///home/projects/vitejs-vite-ogfgnu/src/styles/theme.css.ts:11:36
```

- if I rename this file to `theme.ts` the build succeeds, however the browser will then log a similar error once running `vite preview`

```
Uncaught TypeError: obj.constructor is not a function
    walkObject vanilla-extract-private.esm.js:27
    createThemeContract vanilla-extract-css.browser.esm.js:280
    createGlobalTheme vanilla-extract-css.browser.esm.js:299
    <anonymous> theme.ts:5
```

This can be reproduced in the attached stackblitz repo, by renaming the file in `src/styles/theme.css.ts` to `theme.ts` and updating the import in `src/App.vue`.

### Reproduction

https://stackblitz.com/edit/vitejs-vite-1q7kbr

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 55.58 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.nvm/versions/node/v20.11.0/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.0/bin/npm
    pnpm: 8.14.1 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 122.0.6261.129
    Edge: 122.0.2365.92
    Firefox Nightly: 125.0a1
    Safari: 17.3.1
```

### Used Package Manager

npm

### Logs

```shell
see above
```

### Validations

askoufis:
Fixes #1360.

`walkObject` doesn't work on module namespace objects when running inside `vite-node`. This is because [module namespace objects don't have a `constructor` function](https://262.ecma-international.org/6.0/#sec-module-namespace-objects).

This "bug" has gone unnoticed for a long time because `esbuild` was modifying the export in some way that made it work, but `vite-node` doesn't seem to do this.

I don't think we need to use `constructor` anymore since `walkObject` is constrained to only work on objects, at least as far as the type signature is concerned.

### DIFF
--- a/packages/private/src/walkObject.ts
+++ b/packages/private/src/walkObject.ts
@@ -11,7 +11,7 @@ export function walkObject<T extends Walkable, MapTo>(
   fn: (value: Primitive, path: Array<string>) => MapTo,
   path: Array<string> = [],
 ): MapLeafNodes<T, MapTo> {
-  const clone = obj.constructor();
+  const clone = {} as any;
 
   for (let key in obj) {
     const value = obj[key];
--- a/.changeset/chilly-hairs-serve.md
+++ b/.changeset/chilly-hairs-serve.md
@@ -2,14 +2,14 @@
 '@vanilla-extract/private': patch
 ---
 
-**walkObject**: Use an empty object instead of calling the input object's `constructor` when mapping over it
+**walkObject**: Use an empty object to initialize a clone instead of calling the input object's `constructor`
 
 This allows `walkObject` to be used on module namespace objects:
 
 ```ts
 import * as ns from './foo';
 
-// Would sometimes cause a runtime error
+// Runtime error in `vite-node`
 walkObject(ns, myMappingFunction);
 ```
 
--- a/.changeset/chilly-hairs-serve.md
+++ b/.changeset/chilly-hairs-serve.md
@@ -1,0 +1,19 @@
+---
+'@vanilla-extract/private': patch
+---
+
+**walkObject**: Use an empty object instead of calling the input object's `constructor` when mapping over it
+
+This allows `walkObject` to be used on module namespace objects:
+
+```ts
+import * as ns from './foo';
+
+// Would sometimes cause a runtime error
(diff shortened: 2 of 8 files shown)


packages/private/src/walkObject.ts
```diff
@@ -11,7 +11,7 @@ export function walkObject<T extends Walkable, MapTo>(
   fn: (value: Primitive, path: Array<string>) => MapTo,
   path: Array<string> = [],
 ): MapLeafNodes<T, MapTo> {
-  const clone = obj.constructor();
+  const clone = {} as any;
```
askoufis, line 14: The types are a bit gnarly here, `clone` was `any` before this change, so this just keeps the status quo.

michaeltaranto (approved):

gu-stav:
Thank you everyone 👏

PR #1381, opened 2024-04-12
fix(recipes): applying recipes throws for unknown options

christoph-fricke:
I recently started to move from [class-variance-authority](https://cva.style/docs/getting-started/variants) to vanilla-extract with recipes for the additional type safety. So far I am loving it apart from some nit-picks.

I am using both in React components where I find it quite handy to pass the props object to the recipe so it can apply all variants that are provided by the consumer. This works flawlessly with _class-variance-authority_. However, I noticed that the following code throws a runtime error "TypeError: Cannot read properties of undefined" in _vanilla-extract_ when non-variant properties are applied as well. I would love to avoid having to de-structure and construct a variant options object all the time.

```typescript
import { text, type TextVariants } from './text.css';

export interface TextProps extends TextVariants, ComponentPropsWithoutRef<"span"> {}

export function Text(props: TextProps) {
  return (
    <span {...props} className={text(props)}>
      {props.children}
    </span>
  );
}
```

This PR addresses the type error by ignoring unknown options when applying a recipe function. Therefore, recipes no longer throw a runtime error.

### DIFF
--- a/.changeset/calm-impalas-itch.md
+++ b/.changeset/calm-impalas-itch.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/recipes': patch
+---
+
+Applying recipes now ignores unknown options instead of throwing a runtime TypeError.
--- a/packages/recipes/src/createRuntimeFn.ts
+++ b/packages/recipes/src/createRuntimeFn.ts
@@ -45,7 +45,7 @@ export const createRuntimeFn = <Variants extends VariantGroups>(
 
         const selectionClassName =
           // @ts-expect-error
-          config.variantClassNames[variantName][selection];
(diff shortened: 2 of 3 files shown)


kasper573:
Don’t you need to destructure anyway to not pass on variant props to the underlying component, in this case the div?

christoph-fricke:
@kasper573 I noticed that some props appear in the DOM when just spread onto the component, but I haven't seen any warnings, yet. However, I am quite early in the switch, so I might notice them down the line.

Nonetheless, I believe that the `recipes` function should not throw a runtime error when unknown props are passed in. So I think this PR is still relevant.

PR #1386, opened 2024-04-20
Replace `outdent` dependency with `dedent`

askoufis:
Really the only reason to move is that `dedent` seems much more actively maintained than `outdent`.


mattcompiles (approved):

SimenB:
I think this change broke something in how dedenting works when passing a multiline string.

```ts
import { createGlobalTheme } from "@vanilla-extract/css";

const vars = createGlobalTheme(":root", {
  focusWithin: {
    highlightBoxShadow: `
      0px 0px 0px 4px #DEEFFF
    `,
  },
});
```

This works fine in `1.15.0`, but with `1.15.1` the backticks are turned into `"`, which is a syntax error

askoufis:
> I think this change broke something in how dedenting works when passing a multiline string in `@vanilla-extract/css`.
> 
> ```ts
> import { createGlobalTheme } from "@vanilla-extract/css";
> 
> const vars = createGlobalTheme(":root", {
>   focusWithin: {
>     highlightBoxShadow: `
>       0px 0px 0px 4px #DEEFFF
>     `,
>   },
> });
> ```
> 
> This works fine in `1.15.0`, but with `1.15.1` the backticks are turned into `"`, which is a syntax error.
> 
> It might not be the change in _this_ PR, but it who knows 🙂 It broke in the patch release this PR was included in at least

I can't seem to reproduce this. This is the CSS that I'm seeing generated. It looks a bit weird, but is otherwise completely functional:

image

If you find a minimal reproduction, please raise a separate issue.

SimenB:
Fair enough 😀 

https://github.com/vanilla-extract-css/vanilla-extract/issues/1405

PR #1425, opened 2024-06-14
css: Fix conditional rule merging bug

### LINKED ISSUE #1417: When using "globalStyle" with "layer" styles are missing from generated css output if the same selector is used multiple times
### Describe the bug

When using `globalStyle()` with `layer()`, if the same selector is used to define css rules multiple times with `globalStyle`, only rules from the last `globalStyle` call are found in generated css output, rules from other calls before the last one are missing.

```typescript
// Component.css.ts

import { layer, style, globalStyle } from '@vanilla-extract/css';

const testLayer = layer();

export const test = style({});

// please notice same "test" selector are used 2 times to define different css rules with "globalStyle"

globalStyle(test, {
  '@layer': {
    // rules defined here are missing from generated css output
    [testLayer]: { color: 'black' },
  },
});

globalStyle(test, {
  '@layer': {
    // rules defined here are correctly generated in css output
    // this is the last call of "globalStyle" for "test" selector
    [testLayer]: { fontSize: 16 },
  },
});
```

 ```typescript
// Component.tsx

import * as styles from './Component.css';

export default function test() {
  return <div css={styles.test}>test</div>;
}
```

```css
// generated css output
// it only has css properties defined from last rule
// "color: black" from the first "globalStyle(test)" call is missing

@layer Component_testLayer__1n2kpd00;
@layer Component_testLayer__1n2kpd00 {
  .Component_test__1n2kpd01 {
    font-size: 16px;
  }
}
```

I found the cause of this issue is from this line:
https://github.com/vanilla-extract-css/vanilla-extract/blob/e7d79f2770cf7a8b3bc64ecbd45be19a8f79a896/packages/css/src/conditionalRulesets.ts#L199 

When I update it as below code the issue is gone. 
```typescript
selectors[rule.selector] = { ...selectors[rule.selector], ...rule.rule };
``` 

Please help to fix this as my code structure is relying this pattern quite a lot and it's blocker for me ATM. Thanks team!

### Reproduction

https://not-available

### System Info

```shell
Not Available
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

askoufis:
Fixes #1417.

This fix doesn't preserve the separate declarations, but AFAIK this doesn't have any practical effect on the CSS, other than making it smaller. I also don't think it would be possible to preserve the separate declarations without refactoring how the merging works, so I'm happy with this solution for now.

### DIFF
--- a/packages/css/src/transformCss.test.ts
+++ b/packages/css/src/transformCss.test.ts
@@ -583,6 +583,7 @@ describe('transformCss', () => {
       @layer myLayer;
       @layer myLayer {
         .testClass {
+          color: red;
           font-size: 32px;
         }
       }
--- a/packages/css/src/transformCss.test.ts
+++ b/packages/css/src/transformCss.test.ts
@@ -549,6 +549,46 @@ describe('transformCss', () => {
     `);
   });
 
+  it.only('should merge declarations with the same selector when merging conditional rules', () => {
+    expect(
+      transformCss({
+        composedClassLists: [],
+        localClassNames: [],
+        cssObjs: [
+          {
+            type: 'local',
+            selector: '.testClass',
+            rule: {
+              '@layer': {
+                myLayer: {
+                  color: 'red',
+                },
+              },
+            },
+          },
+          {
+            type: 'local',
+            selector: '.testClass',
+            rule: {
+              '@layer': {
+                myLayer: {
+                  fontSize: '32px',
+                },
+              },
+            },
+          },
+        ],
+      }).join('\n'),
+    ).toMatchInlineSnapshot(`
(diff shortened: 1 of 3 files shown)


packages/css/src/transformCss.test.ts
```diff
@@ -549,6 +549,47 @@ describe('transformCss', () => {
     `);
   });
 
+  it.only('should merge declarations with the same selector when merging conditional rules', () => {
```
mattcompiles, line 552: it.only

askoufis: woops

mattcompiles (approved):

PR #1431, opened 2024-06-17
Fixed composedClassLists breaking order

### LINKED ISSUE #1430: composedClassLists order completely wrong on new Vite compiler
### Describe the bug

Branching off https://github.com/vanilla-extract-css/vanilla-extract/issues/1424 since this is a separate issue.

Please check out the repro repo:
https://github.com/otaviomad/ve-bug-repro/tree/other-bug (with current version compiler)
https://github.com/otaviomad/ve-bug-repro/tree/other-bug-non-vite (with previous version compiler)

If you run `pnpm run-many -t build && pnpm storybook` you can get them started. (The readme is wrong, I was working on another bug with the same repo base)

We have a base class, which extends another class:

image

We have a button, which extends said base class:

image

Then we have a stepper container class, which also extends said base class:

image

And finally, we have a child of the stepper container which has a composite selector for both of the previous components:

image

This is the generated CSS:

image

As you can see, the base class is sandwiched between the stepper container and the button class, breaking the css. 

I believe this issue occurs due to the new way that the class registration happens in the Vite compiler. After some investigation I have found the following:

This is how `composedClassLists` looked like in the old compiler, before being reversed by `transformCss`:

image

Notice how it goes from base -> button -> stepper, which is the correct order in the module graph.

This is how `composedClassLists` looks inside the new compiler, also before reverse ordering:

image

Notice how now stepper comes first, completely breaking the order. What's more, this is import order relevant. In our stepper, we're now importing "base" before "button", but if we change that order, the array order is also changed:

image

I've opened a PR with a possible fix.

Sorry for the images and lengthy issue, it's kinda deep ;)

### Reproduction

https://github.com/otaviomad/ve-bug-repro/tree/other-bug

### System Info

```shell
-
```

### Used Package Manager

pnpm

### Logs

_No response_

### Validations

otaviomad:
Closes #1430 

This PR achieves this by doing the following:

- Changing scanModule to be a recursive function that always keeps the dependencies last in the array order
- Building up the provided classlist based on the module id, so that it's always in order

The way that I did these things might not be to the Vanilla Extract's maintainers standard, so consider this a draft or a suggestion. I have no clue if this breaks something else as well, this is just me figuring out how this issue could be fixed.


otaviomad:
I see that I broke some tests by literally changing the orders. If they are not issues, I can update them to match the new order.

askoufis:
@otaviomad Sorry for taking so long to get to this PR, I kept getting stuck on reproducing the issue in a test. I've pushed the test to a branch - you can see the incorrect class name [here](https://github.com/vanilla-extract-css/vanilla-extract/compare/class-composition-bug-test?expand=1#diff-bfde3d87ff463cc28a87bb59fb75bb0258037b96a7264034d5ff8724c8f4585cR343). However, when I apply your fix to the compiler, it doesn't remove the unwanted `.base` class, but rather just moves it.

Hopefully the unit test can help you debug the issue. Thanks for looking into this, it's very much appreciated.

otaviomad:
@askoufis Impressive work!

I am on vacation at the moment and will only be able to look at it in two weeks. I'll pick it up as soon as I'm back :)

askoufis:
> @askoufis Impressive work!
> 
> I am on vacation at the moment and will only be able to look at it in two weeks. I'll pick it up as soon as I'm back :)

No worries, don't want to rush you. Enjoy your vacation!

otaviomad:
I believe I have fixed it. The problem arose from having some confusion on my `scanModule` re-implementation. 

The way that it works is that it _should_ return cssDeps as an array with the order being: 
nested modules -> shallow dep modules -> the actual scanned module

What was happening is that, to help push the nested modules to the beginning of the array, I would use a Set and for every occurrence of a module, I would delete it and add it again to the Set. However, this would rely on the nested modules actually being _ordered to the end_ of the array, _not the beginning_, which is what scanModule actually returned (which I did to keep the API similar to the previous version without having to reverse it in place on the compiler). So the fix was just to reverse the nested scanModule result to get the proper order that we wanted :)

otaviomad:
It seems playwright is failing, probably due to ordering stuffs. I dunno how I'd fix this one though. Any help?

renrizzolo:
> It seems playwright is failing, probably due to ordering stuffs. I dunno how I'd fix this one though. Any help?

assuming the new output is what you expect, you can locally change `updateSnapshots` in `playwright.config.ts` to `"all"`, then run `test:playwright`. 
Then verify/commit the changed snapshots and revert the playwright config change.

otaviomad:
It's done :) 
@askoufis may we get a formal review?

smitev:
Any progress on this one?
Seems like this will fix one of the issues we have, too.
I hope this will be merged in soon.

tests/e2e/thirdparty.playwright.ts-snapshots/thirdparty-esbuild-next--development.css
```diff
@@ -1,11 +1,11 @@
...
-  background-color: var(--depColor__vg1esd0);
-}
 .styles_depdepBlock__7q0fkw1 {
   --depdepColor__7q0fkw0: blue;
   background-color: var(--depdepColor__7q0fkw0);
 }
+.styles_depBlock__vg1esd1 {
+  --depColor__vg1esd0: green;
+  background-color: var(--depColor__vg1esd0);
+}
```
askoufis, line 8: I don't think this ordering is correct. Looking at the import order of styles in the `thirdparty` fixture (see [here](https://github.com/vanilla-extract-css/vanilla-extract/blob/9a0c6898e02ffe54c8455110ac0a7adecb334acc/fixtures/thirdparty/thirdparty-dep/index.mjs)), I would expect the `depBlock` style to be declared before the `depdepBlock` style.

otaviomad: I think I agree. I'll have a look at it to figure out why the order changed here.

otaviomad:
I _believe_ I fixed everything now. I changed the way it works so that there isn't a weird cache argument being mutated and no longer reversing it twice (which was the culprit, in the end). After hitting my head in the wall for a bit, I realized that "local" dependencies had to be kept in order as well as having the nested deps further up in the dep array. I changed the implementation to use an array, which will keep copies of the nested dependency ids unshifted to the beginning, only to be filtered out by creating a set later, which will make the array ids unique. 

Hopefully this is good enough now :)

otaviomad:
Turns out I was thinking about this wrong. It was never about unshifting or changing orders - the problem lied in the fact that scanModule ran sequentially - thus breaking the self-insertion of css modules as dependencies. Changing it to a recursive function was the fix from the get go, no reversing or breaking stuff needed to begin with. This allowed me to simplify some things - I think it's pretty good now and shouldn't break behaviour anywhere :)

askoufis:
Thanks for persevering on this @otaviomad. The implementation looks good to me. However, I wonder if the effect on CSS output is significant enough to warrant a major version for any consuming packages of the compiler (vite and esbuild-next). Thoughts?

otaviomad:
@askoufis I believe these are minor changes. No majors required.

askoufis:
> @askoufis I believe these are minor changes. No majors required.

I guess my concern was that someone could have written some extra styles to combat incorrect styles generated by this bug (which they don't necessarily know about). By fixing the CSS output, it _could_ result in those extra styles causing unwanted styling. I think it's likely a very rare edge case though, so I'm happy to merge this as a patch.

.changeset/late-icons-protect.md
```diff
@@ -0,0 +1,5 @@
+---
+"@vanilla-extract/integration": patch
+---
+
+Fixed composedClassLists breaking order
```
askoufis, line 5:
```suggestion
Fixes a bug in the `vite-node` compiler resulting in incorrectly ordered class compositions in certain cases
```

askoufis (approved):

PR #1436, opened 2024-06-24
Extend esbuild dependency range

SimenB:
- https://github.com/evanw/esbuild/blob/main/CHANGELOG.md#0200 (deno only)
- https://github.com/evanw/esbuild/blob/main/CHANGELOG.md#0210 (safe rather than sorry, so release as breaking)

Tests pass locally, so presumably this is safe 😀

### DIFF
--- a/.changeset/pink-socks-switch.md
+++ b/.changeset/pink-socks-switch.md
@@ -3,4 +3,4 @@
 '@vanilla-extract/integration': patch
 ---
 
-Update `esbuild` to v0.21
+Extend `esbuild` dependency range to include `~0.20.0` and `~0.21.0`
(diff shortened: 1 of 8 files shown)


.changeset/pink-socks-switch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/jest-transform': patch
+'@vanilla-extract/integration': patch
+---
+
+Update `esbuild` to v0.21
```
askoufis, line 6:
```suggestion
Extend `esbuild` dependency range to include `~0.20.0` and `~0.21.0`
```

askoufis (approved):
Thanks for the PR!

PR #1444, opened 2024-07-07
vite: allow passing overrides for vite config

XiNiHa:
Adds `viteConfigOverrides` option to the Vite plugin.
This option works as an escape hatch for users to deal with incompatibilities between their config and the VE Vite plugin.


askoufis:
I'd like to understand what problem this is actually solving for you before considering whether this feature should be added. The vite plugin already loads your existing vite config when creating the compiler - see [here](https://github.com/vanilla-extract-css/vanilla-extract/blob/28df959c1d407c0ae03b8dc2d5363de60513d4ff/packages/vite-plugin/src/index.ts#L146) - so anything you want to override could instead just be done in your own vite config, instead of by the plugin.

XiNiHa:
I want to actually "override" the config so have a separate config for running VE transforms, not just inheriting the primary config.

My usecase is kinda tricky. The Solid Vite plugin emits browser-only code if it's not an SSR build, which breaks VE evaluator to break. However since VE Vite plugin just takes my Vite config and runs transform with it, it's impossible to prevent Solid Vite plugin from emitting those code.

While this specific usecase can be fixed by filtering out the Solid Vite plugin (like what we're already doing with the Remix plugin) issues like this will always come up as the target environment between Vite and VE Vite plugin will be always different (unless it's a SSR build) Therefore having an escape hatch will be valuable enough to let users workaround their issues.

riccardoperra:
@XiNiHa did you found a solution to get solid work with vanilla extract?

XiNiHa:
> @XiNiHa did you found a solution to get solid work with vanilla extract?

I just `yarn patch`ed it 😅

https://github.com/contentstech-com/stackflow/blob/feat%2Fsolid-integration/.yarn%2Fpatches%2F%40vanilla-extract-vite-plugin-npm-4.0.12-ed14a9ba3b.patch

PR #1450, opened 2024-07-18
Add createViewTransition function

wuz:
View transitions are starting to get more support and it would be nice to use the same API style from `createContainer` to create locally scoped view transition names.

This PR adds a `createViewContainer` function with a `name` arg and an optional `debugId` arg.

Definitely open to suggestions on naming here - similar to CSS container queries, View Transitions have a `view-transition-name` css property and I thought it would track better with `createContainer -> container-name` to have `createViewTransition -> view-transition-name`.


wuz:
@askoufis not sure if you are the right person to ping here, but I'd love to contribute to this project. This PR was based on some work I've been doing and I'd love to get it merged. If there are other ways I can help out, I'm more than happy to do so as well! We're using vanilla-extract heavily and I'd love to help grow the project.

Thanks!

packages/css/src/viewTransition.ts
```diff
@@ -0,0 +1,7 @@
+import { generateIdentifier } from './identifier';
+
+// createViewTransition is used for locally scoping CSS view transitions
+export const createViewTransition = (name: string, debugId?: string) => {
```
askoufis, line 4: We generally rely on debug styles for local dev, so there's not really much benefit to adding a named suffix to the identifier. It will also stick around in production styles, which isn't ideal. IMO this should basically be identical to the `createContainer` function.

site/docs/api/create-view-transition.md
```diff
@@ -0,0 +1,39 @@
+---
+title: createViewTransition
+parent: api
+---
+
+# createViewTransition
+
+Creates a single scoped view transition name for use with [CSS View Transitions]. This avoids potential naming collisions with other containers.
```
askoufis, line 8:
```suggestion
Creates a single scoped view transition name for use with [CSS View Transitions].
This avoids potential naming collisions with other view transitions.
```

site/docs/api/create-view-transition.md
```diff
@@ -0,0 +1,39 @@
...
+
+> 🚧&nbsp;&nbsp;Ensure your target browsers [support view transitions].
+> Vanilla-extract supports the [container query syntax][css view transition] but does not polyfill the feature in unsupported browsers.
```
askoufis, line 11: Typo here. Also, we don't actually support the syntax just yet. Until [`csstype`](https://github.com/frenic/csstype) supports it, we'd have to implement the types ourselves.

```suggestion
> Vanilla-extract supports the [view transition syntax][css view transition] but does not polyfill the feature in unsupported browsers.
```

packages/babel-plugin-debug-ids/src/index.test.ts
```diff
@@ -230,6 +230,19 @@ describe('babel plugin', () => {
...
+  it('should handle createViewTransition assigned to const', () => {
+    const source = `
+      import { createViewTransition } from '@vanilla-extract/css';
+
+      const myViewTransition = createViewTransition("title");
+    `;
+
+    expect(transform(source)).toMatchInlineSnapshot(`
+      import { createViewTransition } from '@vanilla-extract/css';
+      const myViewTransition = createContainer("title", "myViewTransition");
```
askoufis, line 242: This will get picked up by tests.

```suggestion
      const myViewTransition = createViewTransition("title", "myViewTransition");
```

askoufis (commented):
Sorry for the delay @wuz. I'm happy to help get this through, though there's no API for actually writing a `@view-transition` rule just yet.

`@view-transition` rules are also a bit different to other existing rules in that they can be nested within media queries, but aren't related to any classnames. I might try make a POC for a `viewTransition` API, and maybe that can be released alongside this API eventually.

askoufis:
@wuz I also can't run CI on your fork, you might have some restrictive permissions enabled.

site/docs/api/create-view-transition.md
wuz: I believe csstype has support for the viewTransitionName property! Is that all it would need?

wuz:
@askoufis Sorry, was out of office all week last week! Looking at this now! Let me see what's going on with the CI stuff, not sure what the permissions would be.

packages/css/src/viewTransition.ts
wuz: My thinking here was that you might want to specifically name a view transition, but it makes sense that you would just use debugId for that. I'll change it!

site/docs/api/create-view-transition.md
wuz: Oh, I see what you are saying, there is support for the `@view-transition` property.

wuz:
Thinking about this a bit, the API could look something like this:

```ts
import { globalViewTransition, createViewTransition, createViewTransitionGroup } from '@vanilla-extract/css';

// assuming keyframes called fadeOut and fadeIn defined somewhere

globalViewTransition({
  navigation: 'auto',
});

const fade = createViewTransition({
  old: {
    animationName: fadeOut,
  },
  new: {
    animationName: fadeIn,
  },
});

createViewTransitionGroup(fade, {
  animationDuration: "0.25s",
  animationTimingFunction: "ease",
});
```

becomes:

```css
  @view-transition {
    navigation: auto;
  }

  ::view-transition-old(viewTransition_fade__jxjrfl) {
    animation-name: fadeOut;
  }

  ::view-transition-new(viewTransition_fade__jxjrfl) {
    animation-name: fadeIn;
  }

  ::view-transition-group(viewTransition_fade__jxjrfl) {
    animation-duration: 0.25s;
    animation-timing-function: ease;
  }
*/
```

askoufis:
I'm unsure if we'd expose APIs like `createViewTransition` or `createViewTransitionGroup` as they're quite simple wrappers over some `globalStyle` calls.

Don't know about whether `globalViewTransition` or just `viewTransition` is a better name for defining the `@` rule, but regardless this isn't really the right PR to discuss that.

packages/babel-plugin-debug-ids/src/index.ts
```diff
@@ -39,6 +39,13 @@ const debuggableFunctionConfig = {
   createContainer: {
     maxParams: 1,
   },
+  createViewTransition: {
+    maxParams: 2,
+    hasDebugId: ({ arguments: args }) => {
+      const previousArg = args[args.length - 1];
+      return t.isStringLiteral(previousArg) || t.isTemplateLiteral(previousArg);
+    },
+  },
```
askoufis, line 48: This should be basically identical to `createContainer`.

```suggestion
  createViewTransition: {
    maxParams: 1,
  },
```

packages/css/src/viewTransition.ts
```diff
@@ -0,0 +1,13 @@
...
+import { CSSViewTransition } from './types';
+
+// createViewTransition is used for locally scoping CSS view transitions
+// For now it is mostly just an alias of generateIdentifier
+export const createViewTransition = (debugId?: string) =>
+  generateIdentifier(debugId);
+
+export function viewTransition(rule: CSSViewTransition) {
+  appendCss({ type: 'view-transition', rule }, getFileScope());
+}
```
askoufis, line 13: There's a bit more to adding a `@view-transition` API than just adding this function. The actual generation of CSS needs to be implemented to. Happy to point you in the right direction if you want to implement it, but I was planning to do it in a PR separate to this one.

If you need to use VE to write a view transition rule, you can currently just use a global style as a hack:

```ts
globalStyle('@view-transition': {
  navigation: 'auto',
})
```

Otherwise, the `createViewTransition` stuff looks ready to release though.

wuz: I can remove it to get this out! Just was playing with getting it added, but makes sense to just get this merged.

site/docs/api/create-view-transition.md
```diff
@@ -0,0 +1,37 @@
...
+parent: api
+---
+
+# createViewTransition
+
+Creates a single scoped view transition name for use with [CSS View Transitions].
+This avoids potential naming collisions with other view transitions.
+
+> 🚧&nbsp;&nbsp;Ensure your target browsers [support view transitions].
+> Vanilla-extract supports the [view transition syntax][css view transition] but does not polyfill the feature in unsupported browsers.
```
askoufis, line 12:
```suggestion
> Vanilla-extract supports the [view transition syntax][css view transitions] but does not polyfill the feature in unsupported browsers.
```

askoufis (approved):
Thanks @wuz!

askoufis:
@wuz This has been released in https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fcss%401.16.0

wuz:
@askoufis Awesome! Thanks so much!

PR #1452, opened 2024-07-25
refactor(site): add styleVariants, change createVar to assignVars, change to iife, take unity

sukvvon:
* `Chevron.css.ts`: using styleVarinats is better than object which has many `style` in vanilla-extract so i change object to styleVariants in `direction`
* `ColorModeToggle.css.ts`: using `assignVars` is clear than creating many `createVar`s so i change it with `createThemeContract` like `SyntaxHighlighter.css.ts`
* `DocsPage.css.ts`, `HomePage.css.ts`: remove spread syntax in responsiveStyle for unity with other constants
* `GithubStars.tsx`: getCount use once and we does not use this in useEffect, using iife is more simple than creating getCount and use this once

### DIFF
--- a/site/src/Chevron/Chevron.css.ts
+++ b/site/src/Chevron/Chevron.css.ts
@@ -1,4 +1,4 @@
-import { style } from '@vanilla-extract/css';
+import { style, styleVariants } from '@vanilla-extract/css';
 
 export const root = style({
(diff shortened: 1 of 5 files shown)


askoufis (approved):
Thanks!

PR #1457, opened 2024-08-07
Upgrade `resolve.exports` to v2

emonadeo:
Using `"module"` as a conditional export (`package.json#exports`) is non-standard. According to the [node documentation](https://nodejs.org/api/packages.html#conditional-exports), the correct key here is `import`.

This also upgrades `resolve.exports` from `v1.1.0` to `v2.0.2`.

### DIFF
--- a/fixtures/next-app-router/CHANGELOG.md
+++ b/fixtures/next-app-router/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @fixtures/next-app-router
 
+## 0.0.5
+
+### Patch Changes
+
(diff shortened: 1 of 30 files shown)


askoufis:
Thanks for the PR @emonadeo. We've had the intention to update Vanilla Extract's dev and build tooling for a while now. We'll likely update the outdated `export` conditions alongside that change, rather than in isolation.

I would be happy to accept just the `resolve.exports` change however.

askoufis (approved):

PR #706, opened 2022-05-31
vite/rollup: Return empty sourcemaps

graup:
Fixes #699 

This suppresses the warnings in vite and rollup.

I'm not sure this is really what we want though. Logically this is wrong and the warning actually is correct.

PR #1463, opened 2024-08-18
Update rollup and rollup plugins

askoufis:
Pretty straightforward. Will comment on any changes that need clarification.

### DIFF
--- a/.changeset/twenty-bananas-relate.md
+++ b/.changeset/twenty-bananas-relate.md
@@ -1,0 +1,9 @@
+---
+'@vanilla-extract/css': patch
+'@vanilla-extract/integration': patch
+'@vanilla-extract/private': patch
+'@vanilla-extract/recipes': patch
+'@vanilla-extract/sprinkles': patch
+---
+
+Export types with `export { type T }` syntax
--- a/packages/rollup-plugin/src/index.ts
+++ b/packages/rollup-plugin/src/index.ts
@@ -101,7 +101,7 @@ export function vanillaExtractPlugin({
 
       return {
         code: output,
-        map: chunkInfo.map ?? null,
+        map: null,
       };
     },
   };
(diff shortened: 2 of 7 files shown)


.changeset/twenty-bananas-relate.md
```diff
@@ -0,0 +1,9 @@
+---
+'@vanilla-extract/css': patch
+'@vanilla-extract/integration': patch
+'@vanilla-extract/private': patch
+'@vanilla-extract/recipes': patch
+'@vanilla-extract/sprinkles': patch
+---
+
+Export types with `export { type T }` syntax
```
askoufis, line 9: [`rollup-plugin-dts@6`](https://github.com/Swatinem/rollup-plugin-dts/blob/388052e8cc20022357efb83d968ea9d84b41e22d/CHANGELOG.md#600) started emitting type exports this way. These are the only packages that export types.

packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap
```diff
@@ -507,44 +507,64 @@ export { altTheme, responsiveTheme, theme, vars };
...
+  ],
+  [
```
askoufis, line 525: [`rollup@3`](https://github.com/rollup/rollup/blob/af6a347551c80b1bf49e2076326ff3e03b5c0fef/CHANGELOG.md#300) started emitting both inline and external sourcemaps by default.

scripts/build-dts.ts
```diff
@@ -45,7 +45,7 @@ async function buildEntry(packageDir: string, entryPath: string) {
     const bundle = await rollup({
       input: dtsEntryPath,
       plugins: [
-        externals({
+        nodeExternals({
```
askoufis, line 48: This was just an export rename.

packages/rollup-plugin/src/index.ts
```diff
@@ -101,7 +101,7 @@ export function vanillaExtractPlugin({
 
       return {
         code: output,
-        map: chunkInfo.map ?? null,
+        map: null,
```
askoufis, line 104: The `map` property seems to have been removed, I think in rollup v4 (can't find patch notes for it though).

Based of the discussion in https://github.com/vanilla-extract-css/vanilla-extract/pull/706, I don't think this change affects anything as the `null` was merely used to suppress warnings caused by `chunkInfo.map` being `undefined`.

PR #1464, opened 2024-08-18
Update webpack-related deps

askoufis:
Update webpack-related deps such as loaders/plugins and the dev server.

### DIFF
--- a/test-helpers/src/startFixture/webpack.ts
+++ b/test-helpers/src/startFixture/webpack.ts
@@ -112,7 +112,7 @@ export const startWebpackFixture = (
             url: `http://localhost:${port}`,
             close: () =>
               new Promise<void>((resolveClose) =>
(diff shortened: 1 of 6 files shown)


test-helpers/src/startFixture/webpack.ts
```diff
@@ -112,7 +112,7 @@ export const startWebpackFixture = (
             url: `http://localhost:${port}`,
             close: () =>
               new Promise<void>((resolveClose) =>
-                server.close(() => {
+                server.stopCallback(() => {
```
askoufis, line 115: `close` was replaced with the async `stop` and callback variant `stopCallback`. https://github.com/webpack/webpack-dev-server/blob/c3b532c6360317319793dcda22c76fbfc05fbdcf/migration-v5.md

PR #1466, opened 2024-08-20
perf: Speed up dev prefix generation for long file paths

askoufis:
TL;DR: Replacing a greedy regex with a custom splitting function is ~12,000x faster in some cases. (Local dev) Cold start times decreased by over 50% in pnpm applications.

## The Problem

Across various internal applications, we noticed a sizeable regression in local dev start time when moving from yarn v1 to pnpm. Some analysis with `node --prof` revealed the culprit to be this regex that's used to generate dev prefixes for identifiers:

```js
/(?<dir>[^\/\\]*)?[\/\\]?(?<file>[^\/\\]*)\.css\.(ts|js|tsx|jsx|cjs|mjs)$/
```

The outputs of this regex are the `dir` and `file` strings, which represent the parent directory of the file, and the file name without the `.css.*` extension respectively.
With yarn v1, this regex was being executed quite often on filepaths such as:

```
node_modules/braid-design-system/dist/lib/components/Button/Button.css.mjs
```

Moving to pnpm means these node module paths become very long because they would resolve to their symlinked paths within the pnpm virtual store `.pnpm`. As a result, the regex was now being executed on file paths like this:

```
node_modules/.pnpm/braid-design-system@32.23.0_@types+react@18.3.3_babel-plugin-macros@3.1.0_react-dom@18.3.1_re_ctndskkf4y74v2qksfjwfz6ezy/node_modules/braid-design-system/dist/lib/components/ButtonDir/Button.css.mjs
```

## Benchmarks

To quantify the performance of this regex, I made a separate benchmark repo. https://github.com/askoufis/get-dir-file-benchmark.

Even without the benchmarks, it's relatively obvious that the cause of the poor performance is the greedy `*` tokens. Making them non-greedy helps a little bit, but we can do _much_ better.

The `customSplit*` functions in the benchmark repo are steps towards improving the performance of this function. The rough outline of the improvements are:
- Use the non-greedy `*` (`*?`)
- Don't use a regex, split on `/` instead
- Iterate through the path backwards, constructing `file` and `dir` as we iterate
- Iterate backwards, but just track useful indices and rely on `String.slice` to get the final strings
- No need to handle UTF-16 surrogates, implement a few escape hatches for various edge cases based on some assumptions we can make about the input

`customSplit4` is the final iteration of the function, and is more or less what's implemented in this PR, along with some unit tests. There are also existing unit tests for `addFileScope` that remain unchanged, so I'm fairly certain that there's no change in behaviour.

Since every identifier in a file will have its dev prefix derived from the same file path, I also implemented a simple cache using a `Map`. Would be happy to use a small memoization library, but IMO we don't need anything super sophisticated for this, though happy to be proven wrong.

**EDIT**: This has been [replaced with an LRU cache](https://github.com/vanilla-extract-css/vanilla-extract/pull/1466/commits/867c4dae4abd48e828db5ed5af90d84eea16b790).

If I've missed something, or there's still more perf on the table, let me know.

## Impact

Based on [the benchmarks](https://github.com/askoufis/get-dir-file-benchmark?tab=readme-ov-file#results), the new implementation can be **up to ~12,000x faster** on very long paths. It's only worse than the regex on very long paths where the file name is the majority of the path. I think this is an acceptable trade off as this is likely an uncommon situation.

Internally we have many applications that depend on [our design system](https://github.com/seek-oss/braid-design-system), as well as other internal component libraries, all of which utilize Vanilla Extract styles. Testing this change across a few of applications that use pnpm resulted in a **decrease of cold start times by more than 50%**. Applications that are still using yarn v1 were unaffected by this change.

### DIFF
--- a/packages/css/src/getFileAndDirFromPath.ts
+++ b/packages/css/src/getFileAndDirFromPath.ts
@@ -1,0 +1,69 @@
+const getLastSlashBeforeIndex = (path: string, index: number) => {
+  let pathIndex = index - 1;
+
+  while (pathIndex >= 0) {
+    if (path[pathIndex] === '/') {
+      return pathIndex;
+    }
+
+    pathIndex--;
+  }
+
+  return -1;
+};
+
+type FileAndDir = {
+  file: string;
+  dir?: string;
+};
+
+/**
+ * Assumptions:
+ * - The path is always normalized to use posix file separators (/) (see `addFileScope`)
+ * - The path is always relative to the project root, i.e. there will never be a leading slash (see `addFileScope`)
+ * - As long as `.css` is there, we have a valid `.css.*` file path, because otherwise there wouldn't
+ *   be a file scope to begin with
+ */
+const _getFileAndDirFromPath = (path: string): FileAndDir | undefined => {
+  let file: string;
+  let dir: string | undefined;
+
+  const lastIndexOfDotCss = path.lastIndexOf('.css');
+
+  if (lastIndexOfDotCss === -1) {
+    return;
+  }
+
+  const lastSlashIndex = getLastSlashBeforeIndex(path, lastIndexOfDotCss);
+  if (lastSlashIndex === -1) {
+    return { file: path.slice(0, lastIndexOfDotCss) };
+  }
+
+  let secondLastSlashIndex = getLastSlashBeforeIndex(path, lastSlashIndex - 1);
+
+  // If secondLastSlashIndex is -1, it means that the path looks like `directory/file.css.ts`,
+  // in which case dir will still be sliced starting at 0, which is what we want
+  dir = path.slice(secondLastSlashIndex + 1, lastSlashIndex);
+  file = path.slice(lastSlashIndex + 1, lastIndexOfDotCss);
+
+  return { dir, file };
+};
+
+const memoizedGetFileAndDirFromPath = () => {
+  const cache = new Map();
+
+  return (path: string) => {
+    const cachedResult = cache.get(path);
+
+    if (cachedResult) {
+      return cachedResult;
+    }
+
+    const result = _getFileAndDirFromPath(path);
+    cache.set(path, result);
+
+    return result;
+  };
+};
+
+export const getFileAndDirFromPath = memoizedGetFileAndDirFromPath();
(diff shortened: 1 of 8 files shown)


packages/css/src/getFileAndDirFromPath.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  // If secondLastSlashIndex is -1, it means that the path looks like `directory/file.css.ts`,
+  // in which case dir will still be sliced starting at 0, which is what we want
+  dir = path.slice(secondLastSlashIndex + 1, lastSlashIndex);
+  file = path.slice(lastSlashIndex + 1, lastIndexOfDotCss);
+
+  return { dir, file };
+};
+
+const memoizedGetFileAndDirFromPath = () => {
```
mattcompiles, line 52: Do you see a lot of hits on the cache? Just wondering if this is worth keeping as it'll leak memory slowly in a dev server right?

You could also consider using an LRU? 
https://www.npmjs.com/package/lru-cache

mattcompiles (approved):
Looks good, just one question about the cache but I'll leave as approved and let you make the call 👍

packages/css/src/getFileAndDirFromPath.ts
askoufis: Every identifier generated in a file except the first one will hit the cache, so there's a lot of hits, especially in something like a sprinkles file. I think it would take quite a large app to have a meaningful memory leak, but it's still probably a good idea to use an lru cache.

packages/css/src/identifier.ts
```diff
@@ -15,13 +16,11 @@ function getDevPrefix({
...
     const { filePath } = getFileScope();
 
-    const matches = filePath.match(
-      /(?<dir>[^\/\\]*)?[\/\\]?(?<file>[^\/\\]*)\.css\.(ts|js|tsx|jsx|cjs|mjs)$/,
-    );
+    const debugFileName = getDebugFileName(filePath);
 
-    if (matches && matches.groups) {
-      const { dir, file } = matches.groups;
-      parts.unshift(file && file !== 'index' ? file : dir);
```
askoufis, line 24: This conditional logic was moved inside `getDebugFileName` to take advantage of the cache.

PR #1468, opened 2024-08-29
fix(site/ColorModeToggle/ColorModeToggle.tsx): change ColorMode title colour to color

sukvvon:
* in british "colour" is right but vanilla-extract is using in international so i think naming "color" is good more than.
* we named component as "ColorModeToggle" so i think naming "color" is in contextually consistent.


askoufis (approved):
Thanks!

PR #1471, opened 2024-08-31
refactor(site): unify components with export const arrow function, simplify structure in system and typography

sukvvon:
* unify components with export const arrow function, add Props interface for each component, 
* simplify structure in system and typography


site/src/system/styles/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './reset.css';
```
askoufis, line 1: I'd honestly prefer not to have barrel files. They just cause indirection and don't serve much of a purpose in a relative small app like this docs site.

site/src/Blockquote/Blockquote.tsx
```diff
@@ -2,7 +2,11 @@ import { ReactNode } from 'react';
 import { Box } from '../system';
 import * as styles from './Blockquote.css';
 
-export default (props: { children: ReactNode }) => {
+export interface BlockquoteProps {
```
askoufis, line 5: This, any probably all the other `export`s you've added to prop types, are not necessary because those types aren't used outside the files they're defined in. They can be removed.

askoufis (commented):
Thanks for these changes. I have some minor comments.

askoufis:
@sukvvon Any plans to address the feedback on this PR? If not, I will close it.

sukvvon:
> @sukvvon Any plans to address the feedback on this PR? If not, I will close it.

@askoufis It might be better to close the PR for now. If a good idea comes up, I plan to create a new PR.

Issue #1474, opened 2024-09-12
Missing pseudo-class: -webkit-calendar-picker-indicator


corradopetrelli:
### Describe the bug

WebkitCalendarPickerIndicator is not available on vanilla-extract

### Reproduction

It is missing from simplePseudo file https://github.com/vanilla-extract-css/vanilla-extract/blob/1e48d966c2add14a92febe11fa0a1b66b5a45740/packages/css/src/simplePseudos.ts on codebase

### System Info

```shell
System:
    OS: macOS 14.6
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.78 GB / 16.00 GB
    Shell: 5.9 - /bin/zsh
Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.14.0/bin/npm
```

### Used Package Manager

npm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

askoufis:
@corradopetrelli This has been released in https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fcss%401.16.0.

PR #1475, opened 2024-09-12
fix(ISSUE-1474): add -webkit-calendar-picker-indicator pseudoclass

corradopetrelli:
Resolves: #1474

### DIFF
--- a/packages/css/src/simplePseudos.ts
+++ b/packages/css/src/simplePseudos.ts
@@ -59,7 +59,6 @@ const simplePseudoMap = {
   '::-webkit-search-results-button': true,
   '::-webkit-slider-runnable-track': true,
   '::-webkit-slider-thumb': true,
-  
   '::after': true,
   '::backdrop': true,
   '::before': true,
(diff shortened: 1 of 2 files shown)


packages/css/src/simplePseudos.ts
```diff
@@ -58,6 +59,7 @@ const simplePseudoMap = {
   '::-webkit-search-results-button': true,
   '::-webkit-slider-runnable-track': true,
   '::-webkit-slider-thumb': true,
+
```
corradopetrelli, line 62:
```suggestion
```

askoufis:
Thanks for the PR! For future reference, you can still use the pseudo-selector in your Vanilla Extract styles without the type being valid. It will still end up in the output CSS. There's many browser-specific psuedo-elements, some with very little documentation, so it's impractical to support all of them at a type level without constant upkeep of this file.

askoufis (approved):

corradopetrelli:
@askoufis thank you, that was exactly the workaround I used 😄

PR #1490, opened 2024-10-12
Fix serialization of exotic react compoonent

exah:
Right when we attempt passing `forwardRef` component to `addFunctionSerializer` (following official [guide](https://vanilla-extract.style/documentation/api/add-function-serializer/)) we get error:

```
Invalid exports.

```

This PR adds support for creating generic `styled` react components with custom ref using `addFunctionSerializer`.

### DIFF
--- a/.changeset/ninety-hotels-hear.md
+++ b/.changeset/ninety-hotels-hear.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/integration': major
+---
+
+Support serialisation of exotic react components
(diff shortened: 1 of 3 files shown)


bel0v (approved):

Danilqa (approved):

.changeset/ninety-hotels-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/integration': major
```
askoufis, line 2: Does this need to be a major version? It doesn't seem like a breaking change to me, so it can probably just be a `minor` instead.

```suggestion
'@vanilla-extract/integration': minor
```

PR #1500, opened 2024-11-07
feat: 🎸 expose esbuild log options

ArrayKnight:
To enable better control over what is reported during a build, the esbuild log options have been exposed to allow users to fine tune what they deem important to have logged

### DIFF
--- a/.changeset/soft-wombats-shop.md
+++ b/.changeset/soft-wombats-shop.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/integration': minor
+---
+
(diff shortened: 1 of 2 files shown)


ArrayKnight:
@askoufis I'd love to get this in so I can set:
```
logOverride: {
  'ignored-bare-import': 'silent'
}
```
I'm seeing a lot of logs due to "sideEffects": false in my package.json. My build is actually doing exactly what I want, and without producing side effects. But the amount of spam in my build logs is making parsing it for other content cumbersome

PR #1508, opened 2024-11-26
vite: add vite v6 to peer Dependencies

nilsingwersen:


LekoArts:
Since I've ran into https://github.com/withastro/astro/issues/12614 I think there will be more work necessary for vanilla-extract to be compatible with Vite 6 than just updating the peer dependencies.

SimenB:
This won't be enough as long as the plugin depends on `@vanilla-extract/integration`, which has these deps:

https://github.com/vanilla-extract-css/vanilla-extract/blob/dde9bf2d15120e546be372390fa562384a48f99b/packages/integration/package.json#L28-L29

mgrosperrin:
Hello,
I am looking to start a new project and would like to use vanilla-extract with Vite but I am blocked by the "old" dependency to Vite.

Any way to help validating that the PR resolve the issue?

askoufis:
@mgrosperrin As mentioned in https://github.com/vanilla-extract-css/vanilla-extract/issues/1509#issuecomment-2537484012, the `vitest` v3 release is currently the blocker for the peer dep upgrade. I should've posted in this PR when I made that comment, my bad.

From other user's testing, there doesn't seem to be any incompatibility with vite v6 and Vanilla Extract (except for Astro users), so feel free to ignore/suppress peer-dependency-related warnings if you're comfortable doing that.

jonathan-blueocean:
FYI, Vitest 3 was released yesterday

jonathan-blueocean:
I'm curious, why vitest isn't a devDependency, or why this package does not define devDependencies at all?

PR #1512, opened 2024-12-19
`createTheme`, `createGlobalTheme`: Add support for assigning themes to a layer

askoufis:
[Original discussion](https://github.com/vanilla-extract-css/vanilla-extract/discussions/1472).

This is currently possible in VE using existing APIs ([example](https://github.com/vanilla-extract-css/vanilla-extract/discussions/1472#discussioncomment-11000308)), however it's not exactly ergonomic, nor is it obvious that this is possible, so first-class support would be nice.

Sprinkles supports layers via an `@layer` property, so I implemented the same thing for `createTheme` and `createGlobalTheme`.

**EXAMPLE USAGE**:

```ts
// themes.css.ts
import { createTheme, createGlobalTheme, layer } from '@vanilla-extract/css';

export const themeLayer = layer();

export const [themeA, vars] = createTheme({
  '@layer': themeLayer,
  color: {
    brand: 'blue'
  },
  font: {
    body: 'arial'
  }
});

export const vars2 = createGlobalTheme(':root', {
  '@layer': themeLayer,
  space: {
    small: '10px',
    large: '20px',
  }
});
```

### DIFF
--- a/.changeset/nasty-wasps-tie.md
+++ b/.changeset/nasty-wasps-tie.md
@@ -1,0 +1,51 @@
+---
+'@vanilla-extract/css': minor
+---
+
+`createTheme`, `createGlobalTheme`: Add support for assigning themes to a layer
+
+Themes can now be assigned to a layer by name using the `@layer` key at the top-level of the theme definition.
+
+**EXAMPLE USAGE**:
+
+```ts
+// themes.css.ts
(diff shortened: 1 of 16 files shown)


mattcompiles (approved):
Nice one bud 🍻 🧁 🎄

PR #1517, opened 2024-12-20
chore(.github/workflows/validate.yml): change playwright runs-on 'macos-14' to 'macos-latest'

sukvvon:
* We are using `ubuntu-latest` and `windows-latest` in the `validate.yml`. Therefore, it would be better to use `macos-latest` for macOS as well.


askoufis (approved):
Thanks!

PR #1558, opened 2025-03-13
Fix runtime behavior to match the type definition

andjsrk:
Currently, the definition of `ClassNames` is recursive:
https://github.com/vanilla-extract-css/vanilla-extract/blob/221f07f83844a238604a00e4c2d3c75eb7f087ba/packages/css/src/types.ts#L166
But it is not handled correctly on:
https://github.com/vanilla-extract-css/vanilla-extract/blob/221f07f83844a238604a00e4c2d3c75eb7f087ba/packages/css/src/style.ts#L30-L36
This PR fixes this mismatch.

### DIFF
--- a/packages/css/src/style.ts
+++ b/packages/css/src/style.ts
@@ -27,7 +27,11 @@ function composedStyle(rules: Array<StyleRule | ClassNames>, debugId?: string) {
   const classList = [];
   const styleRules = [];
 
-  for (const rule of rules) {
+  // `flat(Infinity)` fails to compile because of TS2589
+  const flattenedRules = rules.flat(Infinity as 20) as Array<
+    StyleRule | string
+  >;
+  for (const rule of flattenedRules) {
     if (typeof rule === 'string') {
       classList.push(rule);
     } else {


andjsrk:
`dudupeAndJoinClassList` itself works fine, but how it is used in `composedStyle`, which is used for exported `style` function, is improper. Consider:
```ts
style([
  [classA, classB],
  {
    // some style properties
  },
])
```
I don't remember what this exactly emits, as I made this change about a year ago, but this would emit wrong result.

packages/css/src/style.ts
```diff
@@ -28,10 +28,10 @@ function composedStyle(rules: Array<StyleRule | ClassNames>, debugId?: string) {
   const styleRules = [];
 
   for (const rule of rules) {
-    if (typeof rule === 'string') {
-      classList.push(rule);
-    } else {
+    if (typeof rule !== 'string' && !Array.isArray(rule)) {
       styleRules.push(rule);
+    } else {
+      classList.push(rule);
```
askoufis: I'd prefer this logic to be negated. IMO it's a bit easier to understand that way.
```suggestion
    if (typeof rule === 'string' || Array.isArray(rule)) {
      classList.push(rule);
    } else {
      styleRules.push(rule);

```

.changeset/dry-rabbits-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Fix runtime behavior to match the type definition
```
askoufis:
````suggestion
`style`: Fixed a bug where nested arrays of classnames could cause missing or malformed CSS during [style composition](https://vanilla-extract.style/documentation/style-composition/) in certain situations.

For example, the following style composition would not generate CSS for the `backgroundColor: 'orange'` style, and would also generate malformed CSS:

```ts
const styleWithNestedComposition = style([
  [style1, style2],
  { backgroundColor: 'orange' },
  [style3],
]);
```
````

**Review by askoufis — COMMENTED** [review]
Ah I see now. Yes this is a bug. I've left a few comments.

I'll add a fixture test in a separate PR.

File: packages/css/src/style.ts [hunk 3]
```diff
@@ -28,10 +28,11 @@ function composedStyle(rules: Array<StyleRule | ClassNames>, debugId?: string) {
   const styleRules = [];
 
   for (const rule of rules) {
-    if (typeof rule === 'string') {
+    if (typeof rule === 'string' || Array.isArray(rule)) {
       classList.push(rule);
     } else {
       styleRules.push(rule);
+
```
askoufis:
```suggestion
```

askoufis (approved):

askoufis:
@andjsrk This fix was released in https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fcss%401.19.1.

PR #1559, opened 2025-03-13
Support passing arrays of styles to `globalStyle`

andjsrk:


askoufis:
I'm not too sure there's that much value in allowing arrays to be passed to `globalStyle`. It makes more sense for `style` because of style composition, but a thin abstraction around merging arrays of objects doesn't really provide much benefit other than slightly more convenience to the consumer.

If you've got an interesting use case then I'd be happy to hear it, but if not I'm leaning towards closing this PR.

andjsrk:
IIRC, I was writing utilities for building a style object, and encountered this inconsistency, and worked it around with manually  writing a rough wrapper for `deepmerge`(which is an implementation detail). Yeah, it is for convenience, but I hope `globalStyle` to be consistent with `style` in some way (if they still behave differently). I thought making `globalStyle` to accept multiple styles is the most reasonable way to keep them consistent so I submitted this PR, but I'm fine to close this PR if you think this change is not worth it or it is fine to make `style` to accept single style (which is a breaking change, at least at the time the PR submitted).

askoufis:
`globalStyle` and `style` provide different functionality, so their API differences are intentional, not accidental. Style composition is the main reason to support arrays in `style` - you can pass style objects or existing classnames from other `style` calls.

`globalStyle` is for defining any selector you want, and style composition doesn't really make sense here. I don't think first-class support for merging style objects is that valuable, and aligning the APIs doesn't seem worth it either, so I'm going to close this PR.

PR #1560, opened 2025-03-16
Remove unnecessary intersections

andjsrk:
Types that uses `WithQueries<StyleType>` are defined like `StyleType & WithQueries<StyleType>`, for example:
https://github.com/vanilla-extract-css/vanilla-extract/blob/221f07f83844a238604a00e4c2d3c75eb7f087ba/packages/css/src/types.ts#L73
while `WithQueries<StyleType>` is defined as:
https://github.com/vanilla-extract-css/vanilla-extract/blob/221f07f83844a238604a00e4c2d3c75eb7f087ba/packages/css/src/types.ts#L62
Additional `StyleType &` is unnecessary, as `StyleType & WithQueries<StyleType>` is `StyleType & StyleType & AllQueries<StyleType>`.

### DIFF
--- a/.changeset/warm-seahorses-taste.md
+++ b/.changeset/warm-seahorses-taste.md
@@ -2,4 +2,4 @@
 '@vanilla-extract/css': patch
 ---
 
-Remove unnecessary intersections
+Remove unnecessary intersections in a few types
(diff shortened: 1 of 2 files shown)


.changeset/warm-seahorses-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Remove unnecessary intersections
```
askoufis, line 5:
```suggestion
Remove unnecessary intersections in a few types
```

askoufis (approved):
Thanks for the fix!

Issue #1551, opened 2025-02-18
Does not refer to the build assetInlineLimit settings from vite-plugin version 5.


hoiheart:
### Describe the bug

Hello.

Posting the vite-plugin version of my project to 5, I found some different behavior than before.

### 1. It does not follow the build assetsInlineLimit settings set in vite.config.ts.

I set assetsInlineLimit to 0, and if you import the image from *.css.ts and use it as a background, it will be applied as base64 data image unlike before.

### 2. Doublequote does not auto wrap when using the imported image.

If use svg, the base64 code is not valid and must wrap the doublequote.

```tsx
import { style } from '@vanilla-extract/css';
import asset from './logo.svg';

export const logo = style({
  background: `url("${asset}") no-repeat 50% 50%`, ✅
  // If use svg, the base64 code is not valid and must wrap the doublequote.
  // background: `url(${asset}) no-repeat 50% 50%`, ❌
});
```

I created a link to the sample codesandbox.
Please check vite.config.ts and App.css.ts.

Thanks.

### Reproduction

https://codesandbox.io/p/devbox/vite-react-ts-forked-f5v63m

### System Info

```shell
System:
    OS: Linux 6.1 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (2) x64 AMD EPYC
    Memory: 2.05 GB / 4.14 GB
    Container: Yes
    Shell: Unknown
  Binaries:
    Node: 20.9.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 9.8.1 - /usr/local/bin/npm
    pnpm: 8.10.2 - /usr/local/share/npm-global/bin/pnpm
  npmPackages:
    @vanilla-extract/css: 1.17.0 => 1.17.0 
    @vanilla-extract/vite-plugin: 5.0.1 => 5.0.1 
    vite: ^2.2.3 => 2.3.3
```

### Used Package Manager

yarn

### Logs

```shell

```

### Validations

- [x] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

askoufis:
@hoiheart Thanks for raising this issue.

Your first issue will be fixed by #1665.

[...]

PR #1562, opened 2025-03-23
Merge build options from viteConfig in compiler

ngbrown:
I was trying to update Remix's use of @vanilla-extract/integration (which is still at v6) and came across how the new compiler package now inlines small files into the resulting CSS. These dataurl strings broke the Remix [integration tests](https://github.com/remix-run/remix/blob/main/integration/vanilla-extract-test.ts), so I figured it was better to match the prior behavior.

To match the prior behavior, I need to pass the option `viteConfig.build.assetsInlineLimit: 0` into the compiler, but the `build` options were being ignored because the compiler is setting one of them.

So this patch is the minimum to get the option passed. I see that there have been earlier efforts to do a deep merge on the options (at least for the Vite plugin), but this is more narrow and is for the compiler.

### DIFF
--- a/packages/compiler/src/compiler.ts
+++ b/packages/compiler/src/compiler.ts
@@ -117,6 +117,7 @@ const createViteServer = async ({
         // Can be removed once https://github.com/vitejs/vite/pull/19247 is released.
         exclude: [/node_modules/],
       },
+      ...viteConfig.build,
     },
     ssr: {
       noExternal: true,


askoufis:
With remix becoming react router, the upstream issue isn't really that relevant anymore, but this incidentally also fixes #1551. In #1665 I opted not to forward all the `build` properties as it felt a bit risky, so instead only `assetsInlineLimit` is propagated.

askoufis:
@ngbrown The fix in #1665 was released in [@vanilla-extract/compiler@0.3.4](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fcompiler%400.3.4).

Issue #1521, opened 2025-01-06
Support @starting-style


calebegg:
### Describe the bug

https://developer.mozilla.org/en-US/docs/Web/CSS/@starting-style

![image](https://github.com/user-attachments/assets/e2c112ab-223d-4a9d-b2dc-0dcbb4a58f6d)

### Reproduction

N/A

### System Info

```shell
N/A
```

### Used Package Manager

npm

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

askoufis:
@lfantone @calebegg This was released in [@vanilla-extract/css@1.18.0](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fcss%401.18.0).

PR #1566, opened 2025-04-05
Add support for `CSS anchor positioning` rules

lfantone:
This pull request adds complete support for CSS Anchor positioning. 
The update includes implementing all necessary types for rules as well as incorporating the new queries "@starting-style" and "@position-try", allowing for a better experience when using CSS anchor positioning rules

Closes https://github.com/vanilla-extract-css/vanilla-extract/issues/1521


mjanner:
Is this getting any attention?

mettson:
I would love to get this merged as soon as possible! Anyone who can review this PR?

askoufis:
@lfantone Thanks for the PR! I have a couple of requests for this PR:

- I would prefer the `@starting-style` feature be implemented in a separate PR to the `@position-try` feature, primarily to facilitate an easier review
- I appreciate that we don't have a contribution guide to make it easier to find where tests should be added, but we generally add new syntax to the `fixtures/features/src/features.css.ts` file in order to test it out, and update our snapshot tests accordingly. These are our E2E tests.
- The CSS generated by your implementation of `@position-try` doesn't align with the syntax defined in [the spec](https://www.w3.org/TR/css-anchor-position-1/#fallback-rule). Specifically, your implementation results in selectors being placed inside the `@position-try` rule, whereas the spec dictates that only declarations are valid inside `@position-try`.
- Additionally, while it's not mentioned in the spec that `@position-try` rules can be nested inside other at-rules (e.g. `@media`), it makes sense that this should be possible, and https://github.com/w3c/csswg-drafts/issues/11080 seems to suggest this. Honestly given that this spec is still in draft and doesn't even seem complete makes me hesitant to ship support for it in VE just yet. Happy to work through these issues in another PR though.

lfantone:
Thanks for the comment @askoufis, yeah you are right about splitting this into two different PR. I will convert this one to be the `@starting-style` implementation and then work on a another one 💪

lfantone:
I think I manage to get the e2e test with Playwright but all the _parcel cases_ fails with:

```bash
    {
      origin: '@parcel/transformer-css',
      message: 'Unknown at rule: @starting-style',
      name: 'SyntaxError',
      stack: undefined,
      codeFrames: undefined
    }
```

Which makes me wonder if parcel has support for this new css rule or if I'm missing something in some config.

askoufis:
> I think I manage to get the e2e test with Playwright but all the _parcel cases_ fails with:
> 
> ```shell
>     {
>       origin: '@parcel/transformer-css',
>       message: 'Unknown at rule: @starting-style',
>       name: 'SyntaxError',
>       stack: undefined,
>       codeFrames: undefined
>     }
> ```
> 
> Which makes me wonder if parcel has support for this new css rule or if I'm missing something in some config.

It should, though parcel uses `lightningcss` for its CSS bundling. Feel free to bump the `parcel` dep to the latest version (`2.15.0`) as I'd be very surprised if newer versions don't support `@starting-style`.

askoufis:
@lfantone It looks like there's still some anchor positioning code lying around. Additionally it looks like you've accidentally deleted the starting-styles changeset instead of the anchor positioning changeset.

.zed/settings.json
askoufis, line 1: Can this directory be removed please?

lfantone:
Added two more PRs (#1589 & #1589) that implements the initial discussion about splitting the functionalities to make it easier to review.

fixtures/features/src/features.css.ts
```diff
@@ -60,3 +60,11 @@ export const styleVariantsCompositionInSelector = styleVariants({
 globalStyle(`body ${styleVariantsCompositionInSelector.variant}`, {
   fontSize: '24px',
 });
+
+// Style with starting-style
```
askoufis:
```suggestion
```

tests/e2e/themed.playwright.ts-snapshots/themed-parcel--production.css
askoufis, line 1: Looks like the newer version of parcel is re-ordering and deduplicating CSS, resulting in these changes. They're not incorrect changes though, just calling them out.

.changeset/tender-beans-cheat.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/css': minor
+---
+
+Add support for `@starting-style` rules
```
askoufis:
````suggestion
`style`: Add support for `@starting-style` rules

**EXAMPLE USAGE**:

```ts
import { style } from '@vanilla-extact/css';

export const styleWithStartingStyle = style({
  backgroundColor: 'black',
  '@starting-style': {
    backgroundColor: 'white',
  },
});
```
````

File: packages/css/src/types.ts [hunk 3]
```diff
@@ -52,12 +52,16 @@ export type MediaQueries<StyleType> = Query<'@media', StyleType>;
 export type FeatureQueries<StyleType> = Query<'@supports', StyleType>;
 export type ContainerQueries<StyleType> = Query<'@container', StyleType>;
 export type Layers<StyleType> = Query<'@layer', StyleType>;
+export type StartingStyleQueries<StyleType> = {
```
askoufis: Thoughts on the `StartingStyleQueries` name? While the `Queries` suffix aligns with most of the other types, it doesn't really make sense since `@starting-style` isn't a query like `@media` or `@container`. It's more similar to `@layer` in that it _declares_ something. Maybe it should just be called `StartingStyle`?

packages/css/src/transformCss.ts
```diff
@@ -589,6 +615,42 @@ class Stylesheet {
...
 
+  transformStartingStyle(
+    root: CSSStyleBlock | CSSSelectorBlock,
+    rules: WithQueries<StyleWithSelectors>['@starting-style'],
+    parentConditions: Array<string> = [],
+  ) {
+    if (rules) {
+      // Check if there are any nested at-rule keys inside this block.
+      // The presence of any key starting with '@' indicates nested queries,
+      // which are not allowed for @starting-style.
```
askoufis: Just flagging that this is actually valid CSS. `@media` inside `@starting-style` is valid ([example](https://codepen.io/askoufis/pen/qEELoQO)), but it relies on your browser supporting CSS nesting.

Actually, looking at the CSS output of your implementation, it generates nested CSS (`@starting-style` inside a selector). However, `@starting-style` is baseline, _and_ it's newer than CSS nesting, so all browsers that support `@starting-style` support CSS nesting, so I think this is fine. However, I think for the sake of simplicity, disallowing _further_ nested CSS within `@starting-style` is the right move for now. We can always remove this restriction in the future if necessary.

packages/css/src/transformCss.ts
```diff
@@ -408,6 +409,11 @@ class Stylesheet {
         selectorRule['@container'],
         conditions,
       );
+      this.transformStartingStyle(
+        root,
+        selectorRule!['@starting-style'],
+        conditions,
+      );
```
askoufis, line 417: This transformation inside `selectors` needs a unit test.

packages/css/src/transformCss.ts
```diff
@@ -550,6 +571,11 @@ class Stylesheet {
         this.transformLayer(root, supportsRule!['@layer'], conditions);
         this.transformMedia(root, supportsRule!['@media'], conditions);
         this.transformContainer(root, supportsRule!['@container'], conditions);
+        this.transformStartingStyle(
+          root,
+          supportsRule!['@starting-style'],
+          conditions,
+        );
```
askoufis, line 579: This transformation inside `@supports` needs a unit test.

packages/css/src/transformCss.ts
```diff
@@ -589,6 +615,42 @@ class Stylesheet {
...
+          rule: omit(rules, specialKeys),
+        },
+        conditions,
+      );
+
+      // Process any simple pseudos or selectors associated with this style.
+      if (root.type === 'local') {
+        this.transformSimplePseudos(root, rules, conditions);
+        this.transformSelectors(root, rules, conditions);
+      }
```
askoufis, line 652: This logic should have a unit test too.

askoufis (commented):
We're pretty close on this. A few tweaks and some unit tests and it should be ready to merge.

packages/css/src/types.ts
lfantone: Sounds like a good suggestion!

vceder:
Hi @askoufis, I'm helping luciano with this PR and I've added the tests you asked about and taken a look through the rest, see that we've gotten out of sync with main so will rebase put please take a look at your convince 🙂

askoufis:
@vceder I'm happy with where this PR is currently, though I can't approve a CI run. I think @lfantone might have to enable that in their fork.

.changeset/tender-beans-cheat.md
```diff
@@ -0,0 +1,15 @@
+---
+'@vanilla-extract/css': minor
+---
+
+`style`: Add support for `@starting-style` rules
+**EXAMPLE USAGE**:
+```ts
```
askoufis, line 9:
````suggestion
`style`: Add support for `@starting-style` rules

**EXAMPLE USAGE**:

```ts
````

**Comment by larrifax** [comment]
@lfantone, @vceder what's the status of this? Looks like the PR needs a rebase & then a CI run..

Looking forward to this feature 👍

**Comment by askoufis** [comment]
@lfantone I'm keen to get this PR ready for release if you've got some time. I think you need to change some permissions on your fork in order for me to run CI. Assuming CI passes I'd be happy to merge.

**Comment by lfantone** [comment]
There we go! I enable Actions in the fork and sync it with upstream.

🙏 Sorry for the long waiting

**Comment by askoufis** [comment]
@lfantone Could I have permission to push to `master` of your fork? I've fixed a bug and would like to push a commit but can't.

Alternatively you can apply this patch with `git apply 0003-Fix-incorrect-starting-style-serialization.patch`.

[0003-Fix-incorrect-starting-style-serialization.patch](https://github.com/user-attachments/files/24147839/0003-Fix-incorrect-starting-style-serialization.patch)

**Comment by lfantone** [comment]
@askoufis Yes! Added you as a collaborator in the repo and also apply the given patch

File: packages/css/src/transformCss.test.ts [hunk 9]
```diff
@@ -2500,6 +2500,412 @@ describe('transformCss', () => {
...
+              }
+            }
+          }
+        }
+      }
+    `);
+  });
+
+  // BUG: This is a long-standing bug with all at-rules, not just starting-style. This should result
+  // in a runtime error as well as a type error.
```
askoufis, line 2840: This is an unrelated bug I found while fixing a bug in this PR. I'll raise a separate issue for it as I don't want it to block this PR.

askoufis:
@lfantone Thanks for all your help on this.

askoufis (approved):

PR #1573, opened 2025-04-29
fix(`rollup-plugin`): Fix relative paths on Windows env

jussirantala:
All relative paths were resolved to current dir `./` even though they should be for example `../` on Windows 11 and it was caused because `chunkInfo.fileName` had backward slashes `\` instead of forward slashes `/`. This caused broken builds.

Looks like `normalize` or `relative` doesn't work with backward slashes so the bug was fixed with replacing Windows separator with posix separator. On posix system it just replaces posix separator with posix sep again so Mac or Linux is not affected (I tested on Mac).


askoufis:
@jussirantala Could you try and recreate the issue in a test case or raise an issue with a minimal reproduction? I can't manage to get `chunkInfo.fileName` to contain a relative path as rollup prevents emitting files to relative paths. Is there another way to get relative filenames?

jussirantala:
> @jussirantala Could you try and recreate the issue in a test case or raise an issue with a minimal reproduction? I can't manage to get `chunkInfo.fileName` to contain a relative path as rollup prevents emitting files to relative paths. Is there another way to get relative filenames?

@askoufis I'll try to find time for that but what do you mean with rollup preventing emitting files to relative paths? You just need to have something being imported from a upper directory. For example you could import some classes in your styling file from `../internal/baseStyles` but this bug changes the path to `./internal/baseStyles` causing a broken build.

jussirantala:
@askoufis 

So you were using Windows when you tested, right?

askoufis:
@jussirantala I'm still not able to reproduce what you're talking about. [This](https://github.com/vanilla-extract-css/vanilla-extract/compare/rollup-windows-path-test?expand=1#diff-71eed780184ba3d1ce66f1cce1ba6872fa8d58d1237f7ed6e462bdccc24e3327R373) was my attempt to understand you situation. I made a nested entrypoint in a package that imports a file from a parent directory. The relative path looks correct to me.

This is why it would be very helpful if you could create a minimal reproduction of the issue in a repo so I can test it out. Alternative, if you fork the repo and create a failing test case then that would also be useful so I can validate your solution.

jussirantala:
> @jussirantala I'm still not able to reproduce what you're talking about. [This](https://github.com/vanilla-extract-css/vanilla-extract/compare/rollup-windows-path-test?expand=1#diff-71eed780184ba3d1ce66f1cce1ba6872fa8d58d1237f7ed6e462bdccc24e3327R373) was my attempt to understand you situation. I made a nested entrypoint in a package that imports a file from a parent directory. The relative path looks correct to me.
> 
> This is why it would be very helpful if you could create a minimal reproduction of the issue in a repo so I can test it out. Alternative, if you fork the repo and create a failing test case then that would also be useful so I can validate your solution.

@askoufis 
It doesn't look correct to me: https://github.com/vanilla-extract-css/vanilla-extract/compare/rollup-windows-path-test?expand=1#diff-71eed780184ba3d1ce66f1cce1ba6872fa8d58d1237f7ed6e462bdccc24e3327R876

It's pointing to current directory instead of the upper directory.

askoufis:
> It's pointing to current directory instead of the upper directory.

@jussirantala
It's pointing to the current directory because that's where the file is emitted to. If I change the rollup input so it emits that file into a nested directory, then the imports reference the parent directory. See [here](https://github.com/vanilla-extract-css/vanilla-extract/compare/vanilla-extract-css:e2c439f...vanilla-extract-css:c09b99e#diff-71eed780184ba3d1ce66f1cce1ba6872fa8d58d1237f7ed6e462bdccc24e3327R875-R879). The `./..` in the import paths is a bit strange but it's a simple way to enforce a relative path in all cases, and has worked fine for years in the plugin.

Once again, I'd greatly appreciate if you could make a minimal reproduction of your specific issue, instead of me trying to blindly reproduce it.

jussirantala:
Thanks for testing it. I will try to find out time for the minimal repro.

PR #1583, opened 2025-05-08
`vite`: Fix incorrect bundling of imported assets when `base` is set in vite config

askoufis:
During CSS generation, the VE compiler currently honors the configured [`base` path](https://vite.dev/guide/build#public-base-path) (because we forward the entire vite config to the compiler server), applying adjustments to imported asset URLs. When Vite attempts to bundle these adjusted URLs during a `vite build`, it tries to resolve and bundle these asset URLs with respect to the configured `base` path, but it fails to do so because the paths have already been adjusted.

I believe the correct behavior here is to not let the compiler adjust these paths at all, i.e. to ignore the `base` config. Doing so allows build process to correctly resolve and bundle these assets.

cc0 test image found [here](https://free-images.com/display/squirrel_tail_bushy_tail.html).


tests/compiler/compiler.test.ts
```diff
@@ -521,4 +528,27 @@ describe('compiler', () => {
...
+      export var imageStyle1 = 'image_imageStyle1__1lseqzh0';
+      export var imageStyle2 = 'image_imageStyle2__1lseqzh1';
+    `);
+
+    expect(css).toMatchInlineSnapshot(`
+      .image_imageStyle1__1lseqzh0 {
+        background-image: url('/fixtures/vite-config/test.jpg');
+      }
+      .image_imageStyle2__1lseqzh1 {
+        background-image: url('/fixtures/vite-config/test.jpg');
```
askoufis, line 550: Without the change to the compiler to address this issue, the first `background-image` URL would be prepended with `/assets`, which would lead to bundling issues.

Issue #654, opened 2022-04-13
Missing class name when using style to compose sprinkles and other styles


ericmatthys:
### Describe the bug

In specific situations, it appears that all expected class names are not included. An example of the problem is seen with this error message, where the client is missing a class name. I have reproduced this in both a Next.js app and a Storybook app. After a fast refresh, like changing the color from green to red, it will start working again, but the initial build is always broken in both Next.js and Storybook.

`Warning: Prop className did not match. Server: "Example__109b2pw2 sprinkles_width_100%__doo1ph0 Example__109b2pw1" Client: "sprinkles_width_100%__doo1ph0 Example__109b2pw1"`

The example is not as reduced as I'd like yet, but there are a few requirements for the bug to happen afaik.

```
import { style } from "@vanilla-extract/css";
import { sprinkles } from "./sprinkles.css";

export const container = style([
  sprinkles({
    width: "100%"
  }),
  style([{ fontSize: "2em" }])
]);

export const label = style({
  selectors: {
    [`${container}:hover &`]: {
      backgroundColor: "green"
    }
  }
});
```

1. `style` must be used with a `sprinkles` call. Basic styles cannot be included directly here, it can only be composing `sprinkles` and other `style` calls. The example is simplified, so I know including the second `style` doesn't make much sense in this context.
2. The return of that `style` is then used in another selector style. This is what makes or breaks the example. If that unique class name is not included in the DOM, then the label won't turn green on hover.
3. It seems like there needs to be another import that uses `sprinkles` as well, hence the `Divider.css.js` in the example.
4. The responsive properties in `sprinkles` seem to matter as well.

### Link to reproduction

https://codesandbox.io/s/affectionate-turing-kwrpqd

When working, hovering over Hello World should show the green background behind World. When broken, there is no change on hover.

### System Info

Output of `npx envinfo --system --npmPackages @vanilla-extract/css,@vanilla-extract/webpack-plugin,@vanilla-extract/esbuild-plugin,@vanilla-extract/vite-plugin,@vanilla-extract/sprinkles,webpack,esbuild,vite --binaries --browsers`:

My reproduction with Storybook outside of Code Sandbox:

```node
  System:
    OS: macOS 12.1
    CPU: (10) x64 Apple M1 Max
    Memory: 831.85 MB / 64.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 14.17.3 - ~/.volta/tools/image/node/14.17.3/bin/node
[...]

PR #1585, opened 2025-05-10
fix: processVanillaFile fileScopes race

sashank-gogula-glean:
## Problem
`processVanillaFile` (1) awaits while serializing css for each fileScope. For a file with multiple fileScopes (which happens when one `.css.ts` file imports another), it will yield to the event loop before calling `transformCss` on the next fileScope.

If the integration is handling multiple files concurrently, it could yield to another `processVanillaFile` (2).

Since all `processVanillaFile` calls share one [adapaterStack](https://github.com/vanilla-extract-css/vanilla-extract/blob/9241939b9c8f44b78bd6cc6285e5f6c41b6dfd62/packages/css/src/adapter.ts#L13), (2) could push its adapter onto the stack and then yield back to (1) while (2) is awaiting serialization.

If this happens, [currentAdapter](https://github.com/vanilla-extract-css/vanilla-extract/blob/9241939b9c8f44b78bd6cc6285e5f6c41b6dfd62/packages/css/src/adapter.ts#L15) will point to the adapter of (2) for the `transformCss` of the next fileScope in (1).

This causes the `markCompositionUsed` called by `transformCss` to update the wrong adapter. `unusedCompositions` of (1) would then include an identifier that was actually used. The identifier is then stripped and any selectors that referenced it will not work. 

The new test in `processVanillaFile.test.ts` fails without this change.

## Solution
Add the scoped adapter to the stack before calling functions that use it (`evalCode` and `transformCss`) and remove the adapter before any awaits. This way `currentAdapter` will always point to the relevant adapter.

## Context
Pretty sure this is what's going on here: https://github.com/vanilla-extract-css/vanilla-extract/issues/654

### DIFF
--- a/packages/integration/src/processVanillaFile.ts
+++ b/packages/integration/src/processVanillaFile.ts
@@ -1,5 +1,6 @@
 import type { FileScope, Adapter } from '@vanilla-extract/css';
 import { transformCss } from '@vanilla-extract/css/transformCss';
+import { removeAdapter, setAdapter } from '@vanilla-extract/css/adapter';
 import evalCode from 'eval';
 import { stringify } from 'javascript-stringify';
 import dedent from 'dedent';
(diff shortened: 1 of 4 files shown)


sashank-gogula-glean:
@askoufis can this be reviewed?

sashank-gogula-glean:
@askoufis bump

askoufis:
@sashank-gogula-glean Could I have permission to push to your fork?

sashank-gogula-glean:
@askoufis sent an invite

packages/integration/src/processVanillaFile.ts
```diff
@@ -158,20 +166,6 @@ export async function processVanillaFile({
     cssImports.push(virtualCssFilePath);
   }
 
-  // We run this code inside eval as jest seems to create a difrerent instance of the adapter file
-  // for requires executed within the eval and all CSS can be lost.
```
askoufis, line 162: Hopefully this is no longer an issue. I can't find any issues in the jest repo about it, but will monitor post-release in case it resurfaces any bugs.

askoufis:
@sashank-gogula-glean Thanks for the PR!

askoufis:
@sashank-gogula-glean This fix was released in https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fintegration%408.0.9.

PR #1589, opened 2025-05-18
Add support for `@position-try` rules types

lfantone:


phaux:
why is this API so different from `keyframes` or `fontFace`?

askoufis:
> why is this API so different from `keyframes` or `fontFace`?

I was going to do a proper review and bring up the same issue. This should really be a proper API like `keyframes` and `fontFace` since it's generating a [dashed-ident](https://developer.mozilla.org/en-US/docs/Web/CSS/dashed-ident) that needs to be scoped, just like classnames, css vars, etc.

Issue #1488, opened 2024-10-10
@vanilla-extract/vite-plugin loads css files multiple times in Storybook when using theme


SamBernaerdtOntoforce:
### Describe the bug

**Minimal Problem:**
When using themes, vite loads a new `(filename).css.ts.vanilla.css?t=1728553686586` file for every other file that uses itin the development build of storybook.

<img width="1339" alt="image" src="https://github.com/user-attachments/assets/ddcbaeb7-c1a5-4e35-88ea-3b509cc220cd">

**Expected:**
Each .css.ts file should only be loaded once (until it changes and causes a hot reload)

**Problem:**
This might not be an issue on the small reproducible project, but on a larger project, where multiple theme files are composited, each of the theme files is loaded for each component using a theme value. This causes so many file loads that it freezes storybook.

<img width="1084" alt="image" src="https://github.com/user-attachments/assets/a39c84e8-180a-4ec7-8210-229ee3e276fd">

### Reproduction

https://github.com/SamBernaerdtOntoforce/vanilla-extract-theme-and-storybook

### System Info

```shell
System:
    OS: macOS 14.7
    CPU: (12) arm64 Apple M2 Pro
    Memory: 279.81 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.7.3 - ~/.nvm/versions/node/v21.7.3/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v21.7.3/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v21.7.3/bin/npm
    pnpm: 8.8.0 - ~/.nvm/versions/node/v21.7.3/bin/pnpm
  Browsers:
    Chrome: 129.0.6668.100
    Edge: 129.0.2792.79
    Safari: 18.0
  npmPackages:
    @vanilla-extract/css: ^1.16.0 => 1.16.0
    @vanilla-extract/vite-plugin: ^4.0.16 => 4.0.16
    vite: ^5.4.8 => 5.4.8
```

### Used Package Manager

yarn

### Validations

- [X] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [X] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

askoufis:
[...]

PR #1591, opened 2025-05-20
Implement `enableDevCache` option in Vite plugin

hartz89:
Addresses issue #1488 

This PR implements an `enableDevCache` option in `@vanilla-extract/vite-plugin` which helps alleviate dev server performance issues in large projects caused by redundant `.vanilla.css` file loading.

I made this an option rather than a default because there is an obvious trade-off here: adding these filesystem and hashing operations incurs cost, so the default / `false` setting is likely faster in small projects.  But in large projects, this seems worthwhile.

It cut my cold start load times by about 25%. I think the performance improvement would've been bigger if I hadn't already added sprinkles and recipes runtime imports to `optimizeDeps.include`- they were causing some HMR churn on startup so it made the duplicate imports more costly.

### DIFF
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -1,5 +1,5 @@
 import path from 'path';
-
+import fs from 'fs/promises';
 import type {
   Plugin,
   ResolvedConfig,
@@ -16,6 +16,7 @@ import {
   getPackageInfo,
   transform,
   normalizePath,
+  hash
 } from '@vanilla-extract/integration';
 
 const PLUGIN_NAME = 'vite-plugin-vanilla-extract';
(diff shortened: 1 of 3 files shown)


packages/vite-plugin/src/index.ts
```diff
@@ -217,9 +266,10 @@ export function vanillaExtractPlugin({
         }
 
         for (const file of watchFiles) {
+          const normalizedFilePath = normalizePath(file);
```
hartz89, line 269: I wasn't sure if I should put the invalidation check before or after the "add watch file" block, or how these blocks might affect one another. Let me know if you have thoughts.

packages/vite-plugin/src/index.ts
```diff
@@ -91,14 +100,54 @@ export function vanillaExtractPlugin({
...
+  // Helper to determine if a module should be invalidated based on dev cache
+  async function shouldInvalidateModule(moduleId: string): Promise<boolean> {
+    if (
+      !enableDevCache ||
+      !moduleId
+    ) {
+      return true;
+    }
+
+    const hash = await getFileHash(moduleId);
```
hartz89, line 127: Is the hashing step just a waste of CPU time? Should we just use the file contents as the values of the map?

PR #1596, opened 2025-05-27
fix: add Primitive to Serializable

cqh963852:
# Fix addFunctionSerializer won't accept `string[]` type

After check `__recipe__` usage, I'm sure it can accept string[] type here.

https://github.com/vanilla-extract-css/vanilla-extract/blob/2ac5a6b824f0518eea356b3d882f4a90cc890eab/packages/integration/src/processVanillaFile.ts#L261

https://github.com/vanilla-extract-css/vanilla-extract/blob/2ac5a6b824f0518eea356b3d882f4a90cc890eab/packages/integration/src/processVanillaFile.ts#L196


packages/integration/src/processVanillaFile.test.ts
```diff
@@ -138,6 +139,28 @@ describe('serializeVanillaModule', () => {
...
 
+  test('should handle deprecated __recipe__ function serialization', () => {
+    const sprinkles = () => {};
+    // Once support for `__recipe__` is removed, this test can be removed,
+    // and all other tests that use `__function_serializer__` should be updated
+    // to use `addFunctionSerializer`
+    addFunctionSerializer(sprinkles, {
+      importPath: 'my-package',
+      importName: 'myFunction',
+      args: ['arg1', 'arg2'],
```
askoufis, line 150: There's an existing test that uses an array of primitives, but it sets `__function_serializer__` directly, so function was never actually tested.

This change actually makes [the example in the docs](https://vanilla-extract.style/documentation/api/add-function-serializer/) type check.

askoufis (approved):
Thanks for the fix!

PR #1597, opened 2025-05-30
Improve ESM package entrypoints

drwpow:
## Changes

The packages’ entrypoints could be improved. Specifically:

- `module` is not a valid entrypoint (and has never been—it’s simply been “allowed” but never been officially a standard). It’s now-replaced with `exports`
- The official Node.js-supported ESM entrypoint is `"import"`, not `"module"` ([docs](https://github.com/vanilla-extract-css/vanilla-extract/discussions/1588)) (however, just to not break anything, `"module"` was left alone)
- `"./*": "./*"` was added to preserve backwards-compatible behavior. Whenever `"exports"` is specified, all imports are forbidden by default, and this explicitly re-allows manual importing.

All this will result in **free, backwards-compatible ESM support** for all packages. This shouldn’t cause any breaking changes for anyone, only improve bundler compatibility today and for the future.

_Note: I was originally starting on https://github.com/vanilla-extract-css/vanilla-extract/discussions/1588, but got distracted when I noticed the package ESM entrypoints needed updating_

## Feedback

- Tests should pass, obviously
- This shouldn‘t result in any breaking changes for everyone, however, please request changes if anything feels “off”!


packages/css/adapter/package.json
```diff
@@ -10,9 +10,11 @@
     "./package.json": "./package.json",
     ".": {
       "browser": {
+        "import": "./dist/vanilla-extract-css-adapter.browser.esm.js",
         "module": "./dist/vanilla-extract-css-adapter.browser.esm.js",
```
drwpow, line 14: Where package entrypoints were already specified, they were left alone, `"import"` was merely added to ensure ESM gets picked up

packages/compiler/package.json
```diff
@@ -5,6 +5,14 @@
   "main": "dist/vanilla-extract-compiler.cjs.js",
   "module": "dist/vanilla-extract-compiler.esm.js",
   "types": "dist/vanilla-extract-compiler.cjs.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/vanilla-extract-compiler.cjs.d.ts",
+      "import": "./dist/vanilla-extract-compiler.cjs.js",
+      "default": "./dist/vanilla-extract-compiler.esm.js"
+    },
+    "./*": "./*"
```
drwpow: For net-new `exports`, it prevents importing anything other than the default. `"./*": "./*"` allows everything to be imported, which preserves backwards-compatible behavior for everyone (i.e. this only guides the default ESM import to the correct location, without disrupting anything else)

packages/css/adapter/package.json
drwpow: IMO `"module"` should just be removed—it’s usually ignored—but I’d wait for a breaking version in the future to remove it. There’s no harm to keeping it

askoufis:
> Dunno if you’ve checked out [unbuild](https://github.com/unjs/unbuild).

I have. I do enjoy the package.json exports-first approach to managing entrypoints. I'll consider it when I get around to migrating off `preconstruct`.

> Yeah I might have missed something here in regards to autogenerating—if there’s deeper work here then feel free to close!

I should've clarified. I think they were auto-generated a while ago, but have had some manual tweaks since then, but are mostly stock (I think). Happy with this PR doing some more manual tweaks, but ideally we'd move to a tool that's more hands-off.

> The only thing I wasn’t sure about was the fact that some libraries seemed to have a `.dev.cjs.js` output. That would be unavailable for import without the `./*` glob (unless we manually added an entry for that, too).

I'm not that concerned about the `.dev.cjs.js` entrypoints. They'll go away with other tooling anyway. Vanilla Extract doesn't contain any dev/prod environment-specific code as far as I know.

> I like to add `./package.json` as an export

That would be great too. I've gotten into the habit of doing this in other libraries too, so it would be much appreciated.

drwpow:
Updated the PR to only expose `./package.json`s. Most of the time it’s overkill, again, just errs on the side of caution to not break existing setups.

Funny enough, right after talking about unbuild, I tried it in a pnpm monorepo with a bunch of packages and it fell over 🙃. Ran into some sort of memory leak that I didn’t have time to debug. Thought about tsdown, ended up using plain Rolldown, but I might switch eventually! So yeah I’m definitely on the Rolldown train now 😄

askoufis (approved):
Really appreciate the effort. Thank you so much! Looking forward to your rollup plugin PR.

drwpow:
Ah! So I ran across this, which I had never seen before, so now I understand why there were intentional omissions of `import`: the goal was to not have ESM compatibility: https://github.com/evanw/esbuild/issues/3497#issuecomment-1817573265

2023 isn’t that long ago, but even in 2 years, ESM support has improved even further, and more tools have made the leap. Node’s even shipping better CJS <> ESM interop.

So all that to say now I understand the previous pattern a bit better. I still think this PR is a good decision—just use ESM. And it also gives me more confidence that removing `module` would actually be safe. But at least now we know it was catering to esbuild specifically.

askoufis:
@drwpow That issue is a good find. Lots of useful info there.

On a related note, I'm gonna have to roll this change back. #1609 has been raised, and I've noticed similar issues in an internal webpack project. I think Node is interpreting the `.js` ESM files as CJS, despite it being resolved via the `import` condition. Running `publint` on some packages seems to confirm this, e.g.:

```
pkg.exports["."].import is ./dist/vanilla-extract-vite-plugin.esm.js and is written in ESM, but is interpreted as CJS. Consider using the .mjs extension, e.g. ./dist/vanilla-extract-vite-plugin.esm.mjs
```

In hindsight, [preconstruct intentionally doesn't generate a Node ESM-compatible bundle](https://preconstruct.tools/configuration/#importconditiondefaultexport). It also doesn't provide a way to configure `.mjs` extensions (that I'm aware of). I think I'll end up moving to `tsdown` or something soon enough, and fix these entrypoints for good.

drwpow:
Oh shoot. That’s fine. Yeah even unbuild ONLY lets you ship `.mjs` and `.cjs` for that very reason. Sorry for the difficulty there! Making the jump to ESM is hard 🙃

askoufis:
> Oh shoot. That’s fine. Yeah even unbuild ONLY lets you ship `.mjs` and `.cjs` for that very reason. Sorry for the difficulty there! Making the jump to ESM is hard 🙃

All good. I appreciate the effort, and it's made the issue front-of-mind for me, so I'm more determined to get it right next time.

MattyBalaam:
Appreciate the work everyone is doing trying to support esm better. It’s very difficult.

PR #1600, opened 2025-06-03
chore(.nvmrc): change node version from '22.15.0' to 'lts/jod'

sukvvon:
* relate #1577 

The reason for previously pinning to `22.15.0` seems to be to avoid unexpected changes that could occur when using `lts/*`, as it would automatically upgrade to Node 24 once it becomes the active LTS.

By switching to `lts/jod`, we can:

* Automatically stay up to date with the latest patch versions within the Node 22 LTS line
* Avoid major version upgrades, which is a drawback of using `lts/*`
* Reduce the maintenance burden of manually updating patch versions

Overall, I believe this strikes a good balance between stability and ease of maintenance.


askoufis (approved):
Thanks!

PR #1602, opened 2025-06-07
fix(webpack-plugin): fix 'require is not defined' error in Node 24

Pnlvfx:
The webpack plugin was failing when used with Storybook in Node.js 24 environments due to ESM module resolution. Node.js attempts to execute the code as ESM where 'require' is not defined (reference: storybookjs/storybook#31434).

This fix addresses the compatibility issue by properly handling module imports in ESM contexts.

The solution ensures consistent behavior across both CommonJS and ESM module systems.

### DIFF
--- a/packages/webpack-plugin/src/plugin.ts
+++ b/packages/webpack-plugin/src/plugin.ts
@@ -7,6 +7,10 @@ import type { Compiler, RuleSetRule } from 'webpack';
 import { ChildCompiler } from './compiler';
 import createCompat, { type WebpackCompat } from './compat';
 
+import { createRequire } from 'node:module';
+
+const require = createRequire(import.meta.url);
+
 const pluginName = 'VanillaExtractPlugin';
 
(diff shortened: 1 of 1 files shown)


askoufis:
@Pnlvfx Thanks for the PR. I think this is the right fix, though there's a few more usages of `require.resolve` within the webpack and next.js plugins that need addressing. Preconstruct replaces `import.meta.url` in the CJS bundle with its own shim, so it still works.

I'll push a few more changes to your branch and get this PR merged and released as some of our internal tools are blocked by this bug.

askoufis (approved):

Pnlvfx:
> @Pnlvfx Thanks for the PR. I think this is the right fix, though there's a few more usages of `require.resolve` within the webpack and next.js plugins that need addressing. Preconstruct replaces `import.meta.url` in the CJS bundle with its own shim, so it still works.
> 
> I'll push a few more changes to your branch and get this PR merged and released as some of our internal tools are blocked by this bug.

Great, thank you. i'm happy to have contributed to this ❤️

PR #1604, opened 2025-06-09
Add "extract" option to Rollup plugin

drwpow:
## Changes

Satisfies #1588. Adds an `extract` option to `@vanilla-extract/rollup-plugin` that allows people to bundle one `.css` file rather than leave imports intact. This is ideal for library builds that want to ship `.css` directly, and not rely on downstream consumers to post-process the generated files.

```ts
export default {
  plugins: [
    vanillaExtractPlugin({
      extract: {
        name: "bundle.css",
      },
    }),
  ],
};
```

### Alternatives

- Some similar plugins have an `inject` mode where the JS injects `<style>` tags ([rollup-plugin-styles](https://www.npmjs.com/package/rollup-plugin-styles)). I didn’t include that behavior here, but that could be added as a followup 
  - As a tiny opinion, I’m against it because I feel there are no advantages and only downsides vs loading `.css` directly, and is usually a workaround from some other limitation rather than an ideal approach. But if there’s a demand I don’t feel strongly enough to block folks from using it.
- We could have bundled 1 CSS file per entry, rather than 1 CSS file for the whole build for a more “treeshakeable” approach, but that usually introduces more problems than it solves
  - The possible advantage is that you can have multiple smaller CSS files, however, [CSS does not behave like JS modules do](https://esbuild.github.io/content-types/#css-import-order). With multiple entry files, you have the problem of shared chunks, which means you’d either have duplicate CSS loading in every bundle, or you generate additional common chunks that require loading them manually. The latter scenario is worse because there’s almost a guarantee CSS gets loaded out-of-order, which leads to styling in prod behaving differently than dev. Overall, it’s unwieldy at best, broken at worst, vs loading 1 CSS file
  - 1 CSS file caches well, with the right CDN and headers. Also a single network always beats multiple.
  - When you step back and think about the ideal way to treeshake CSS, probably the best approach is simply **doing what this plugin does by default: leaving the `.css` imports inline** and letting consumers take responsibility.
- Rather than strip `import` statements at the end, we could have simply not generated them in the first place. But that means we’d have to assume responsibility for the mountain of work Rollup does for free: module graph scanning, treeshaking, parsing, and the list goes on.
  - Further, requiring the `extract` option to fork code paths in the first step from the existing behavior, introduces risk as now there are almost 2 entirely-independent logic paths running alongside one another. It increases the possibility that users would get completely-different outputs based on a single setting, which isn’t ideal.
  - Lastly, an alternative would be to **make a deeper change and forego virtual IDs altogether** but that’s riskier for no apparent benefit.

## Reviewing

- Tests added
- Existing tests updated
- See comments for high points


fixtures/react-library-example/src/checkbox/checkbox.css.ts
```diff
@@ -0,0 +1,22 @@
+import { style } from '@vanilla-extract/css';
+import { vars } from '../styles/vars.css.js';
+
+export const label = style({
+  display: 'block',
+  ...vars.typography.body.medium,
+
+  '::before': {
```
drwpow, line 8: I didn’t even look at these styles at all; they’re nonsense. But they’re just here to generate lines of code to test in the final bundle, and pull from the theme

fixtures/react-library-example/tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "baseUrl": "src",
+    "module": "NodeNext",
```
drwpow, line 4: `NodeNext` is a minor thing to help the test more accurate:

```diff
- import * as styles from "./checkbox.css";
+ import * as styles from "./checkbox.css.js";
```

Requiring file extensions ensures we’re not making simple mistakes when matching modules. For example, it’d be easy to simply scan for `/\.css$/` in imports, but that wouldn’t exist in other setups. People could alias it, rename it, etc. etc. So all that to say, by ensuring the imports don’t match the original filename, we’re leveraging Rollup properly to resolve and target the modules we want to bundle

packages/rollup-plugin/src/index.ts
```diff
@@ -9,31 +9,74 @@ import {
...
-interface Options {
+export interface Options {
+  /**
+   * Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc) can be configured by selecting from the following options:
+   * - "short": 7+ character hash. e.g. hnw5tz3
+   * - "debug": human readable prefixes representing the owning filename and a potential rule level debug name. e.g. myfile_mystyle_hnw5tz3
+   * - custom function: takes an object parameter with `hash`, `filePath`, `debugId`, and `packageName`, and returns a customized identifier.
+   * @default "short"
+   * @example ({ hash }) => `prefix_${hash}`
+   */
```
drwpow, line 24: These JSDoc additions are all optional; can modify these if needed

packages/rollup-plugin/src/index.ts
```diff
@@ -9,31 +9,74 @@ import {
...
+   * Extract .css bundle to a specified filename
+   * @default false
+   */
+  extract?:
+    | {
+        /**
+         * Name of emitted .css file.
+         * @default "bundle.css"
+         */
+        name?: string;
```
drwpow, line 46: I thought about allowing things like `"[name]"` or `"[hash]"`, but for reasons outlined in the PR notes, we’re not generating 1 per entry, so `[name]` wouldn’t be respected. I didn’t know if anyone would need `[hash]`. We could add it if needed.

packages/rollup-plugin/src/index.ts
```diff
@@ -9,31 +9,74 @@ import {
...
 }: Options = {}): Plugin {
   const isProduction = process.env.NODE_ENV === 'production';
 
+  let extractedCssIds = new Set<string>(); // only for `extract`
+
   return {
     name: 'vanilla-extract',
+
+    buildStart() {
+      extractedCssIds = new Set(); // refresh every build
```
drwpow, line 70: This may be a little overly-defensive. I just wasn’t sure if there are scenarios where the root function scope sticks around between builds when in watch mode, etc.

fixtures/react-library-example/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@fixtures/react-library-example",
```
drwpow, line 2: Added this fixture because for bundling I really wanted all the following conditions:

- both global and scoped CSS in the same package
- some deterministic ordering present (i.e. some CSS is expected to load before other) 
- Vanilla Extract styles used in runtime, which didn’t have to be React, I was just too lazy to use another framework or do some pseudo-HTML template thing

packages/rollup-plugin/test/rollup-plugin.test.ts
```diff
@@ -33,46 +55,163 @@ async function buildAndMatchSnapshot(outputOptions: OutputOptions) {
...
+        (file) => file.type === 'chunk' && file.fileName.endsWith('.js'),
+      ) as OutputChunk[];
+      for (const jsFile of jsFiles) {
+        expect(jsFile.code).not.toMatch(/^import .*\.css['"]/m);
+      }
+
+      // assert bundle CSS reflects order from @fixtures/react-library-example/index.ts
+      const map = JSON.parse(String((sourcemapAsset as OutputAsset).source));
+      expect(map.sources).toEqual([
+        'src/styles/reset.css.ts.vanilla.css',
```
drwpow, line 98: This is probably the most important assertion out of the whole test, and gives great assurance it’s working as intended:

1. The fact that it’s read off the `.css.map` file means it went into the actual `.css` bundle, and isn’t just “floating around” in Rollup somewhere
2. The completeness of this list ensures all CSS made it into the bundle
3. The order ensures it was bundled in module order, more-or-less (surprisingly, the order is not automatic—Rollup’s naive behavior is loading `utility.css` before all the React code)

packages/rollup-plugin/src/index.ts
```diff
@@ -72,10 +116,11 @@ export function vanillaExtractPlugin({
         },
       };
     },
+
     // Emit .css assets
     moduleParsed(moduleInfo) {
       moduleInfo.importedIdResolutions.forEach((resolution) => {
-        if (resolution.meta.css) {
+        if (resolution.meta.css && !extract) {
```
drwpow, line 123: This is the only code point where previous behavior was modified. But since this is a new option, it’s backwards-compatible with all existing uses

packages/rollup-plugin/src/index.ts
```diff
@@ -104,5 +151,49 @@ export function vanillaExtractPlugin({
         map: null,
       };
     },
+
+    // Generate bundle (if extracting)
+    async buildEnd() {
+      if (!extract) {
+        return;
```
drwpow, line 158: In the 2 new build stages, we’re only running code if `extract` is specified, completing the loop on ensuring backwards-compatible behavior.

packages/rollup-plugin/test/rollup-plugin.test.ts
```diff
@@ -33,46 +55,163 @@ async function buildAndMatchSnapshot(outputOptions: OutputOptions) {
...
+      ).toMatchSnapshot();
     });
   });
 
-  it('should build with preserveModules', async () => {
-    // Preserve JS modules
-    await buildAndMatchSnapshot({
-      format: 'esm',
-      preserveModules: true,
+  describe('Rollup settings', () => {
```
drwpow, line 125: These tests were not modified at all! Choose **Hide Whitespace** to see a better view

packages/rollup-plugin/src/lib.ts
```diff
@@ -0,0 +1,121 @@
...
+      break;
+    }
+    // importedIds preserves the import order within each module
+    chain.push([lastImporterId, mod.importedIds.indexOf(currentId)]);
+  }
+  return chain.reverse();
+}
+
+/** Compare import chains to determine a flat ordering for modules */
+export function sortModules(modules: Record<string, ImportChain>): string[] {
```
drwpow, line 82: Note: you’d really think that this would just come “for free” in Rollup, but [it doesn’t](https://github.com/rollup/rollup/issues/3888). I know this seems like a big pile of indirection, and overkill, but it really is necessary for Rollup to bundle in the right order.

For example, take the fixture entry:

```ts
// 1. Style reset
import './styles/reset.css.js';

// 2. Design library
export { default as Button } from './button/button.js';
export { default as Checkbox } from './checkbox/checkbox.js';
export { default as Radio } from './radio/radio.js';

// 3. Utility CSS should be last
import './styles/utility.css.js';
```

If left to its own devices, Rollup bundles in this order, more-or-less:

1. reset.css.ts
3. utility.css.ts ← :x: That’s a problem!
4. vars.css.ts
5. button.css.ts
3. checkbox.css.ts
3. radio.css.ts

This happens because it’s crawling from the entry, and the plugin stages fire basically as soon as it **discovers** a module. So in all the stages, you’re going in the order in which it was first seen, not necessarily where it sits in the final graph.

All the information about ordering is deep in Rollup’s graph, though! So all this function is doing is leveraging the information Rollup already has, to get the order we want. We’re not duplicating any work Rollup is doing; we’re only iterating over the module graph to sort the way we want.

pnpm-lock.yaml
```diff
@@ -21564,6 +21607,34 @@ snapshots:
       - '@babel/core'
       - babel-plugin-macros
 
+  next@12.3.4(@babel/core@7.23.9)(react-dom@19.1.0(react@19.1.0))(react@19.1.0):
```
askoufis: Noticed some unexpected additions to the lockfile, such as `next`, react 19, etc. I feel like I've had this happen to me when running the next.js tests locally. Regardless, they should be reverted.

packages/rollup-plugin/src/lib.ts
```diff
@@ -0,0 +1,121 @@
...
+  let mod: ModuleInfo = getModuleInfo(id)!;
+  /** [id, order] */
+  const chain: ImportChain = [[id, -1]];
+  // resolve upwards to root entry
+  while (!mod.isEntry) {
+    const { id: currentId, importers } = mod;
+    const lastImporterId = importers.at(-1)!;
+    if (!lastImporterId) {
+      break;
+    }
```
askoufis: The `!` shouldn't be necessary here since you `break` on a falsy value on the next line.
```suggestion
    const lastImporterId = importers.at(-1);
    if (!lastImporterId) {
      break;
    }
```

packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap
```diff
@@ -858,3 +858,899 @@ render();
...
+[
+  [
+    "app.css",
+    "html, body {
+  font-size: 100%;
+  height: 100%;
+  line-height: 1;
+  margin: 0;
+}
+.vars__15dpm870 {
```
askoufis, line 872: It looks like the the vars CSS is being bundled 3 times, presumably because it's referenced in all 3 component's `.css.ts` files.

I suppose it's the safest way to ensure CSS ordering is respected, but I wonder if there's some deduplication that could be done here? Would tracking resolution IDs that have already been added to the CSS bundle be enough?

The alternative is to rely on CSS minification to dedupe the output. VE doesn't claim to generate minimal and/or optimized CSS, so this is also a viable option.

packages/rollup-plugin/src/index.ts
askoufis: I'd expect `[hash]` to be useful for caching, or can you get away with no hash by using the correct headers?

packages/rollup-plugin/src/lib.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/** [id, order] tuple meant for ordering imports */
+export type ImportChain = [string, number][];
+
+/** Trace a file back through its importers, building an ordered list */
+export function buildImportChain(
+  id: string,
+  { getModuleInfo, warn }: Pick<PluginContext, 'getModuleInfo' | 'warn'>,
+): ImportChain {
+  let mod: ModuleInfo = getModuleInfo(id)!;
```
askoufis: Would prefer to not use `!` here and on line 68 if possible. I find `!` usually obfuscates type-issues and a lack of proper error/falsiness handling.

packages/rollup-plugin/src/lib.ts
```diff
@@ -0,0 +1,121 @@
...
+        });
+      }
+    }
+  }
+
+  return { bundle: cssBundle, extractedCssIds };
+}
+
+/** [id, order] tuple meant for ordering imports */
+export type ImportChain = [string, number][];
```
askoufis: You could use named tuple items here which would make the `[id, order]` comments redundant.
```suggestion
/** [id, order] tuple meant for ordering imports */
export type ImportChain = [id: string, order: number][];
```

packages/rollup-plugin/src/index.ts
drwpow: Yeah good call. I forget if we get it “for free” but will test to make sure `[hash]` is respected at the very least (and add a test if possible).

packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap
drwpow: Good catch! Let me dig into “why” this happened first to see if it’s a simple bug or if it’s a correctness / cascade issue.

drwpow: Ha just a dumb typo. When we deeply-crawl all imports, we keep track of all the files we’ve already added to the bundle. I just forgot to add the check to skip it if it’s already been added :P

packages/rollup-plugin/src/index.ts
drwpow: Fixed! This now respects the `assetFilenames` setting of Rollup. I didn’t realize that `name` was needed to let users override this. Was accidentally using `fileName` for no reason.

This doesn’t allow `[hash]` syntax directly in `extract.name`, but it’s better because it just respects Rollup settings and behaves just like any other asset (which should mean less config/less confusion for folks).

drwpow:
Pushed fixes to all the comments! Fixed the duplication issue and allow users to set hashes via [output.assetFilenames](https://rollupjs.org/configuration-options/#output-assetfilenames).

site/docs/integrations/rollup.md
```diff
@@ -90,3 +90,13 @@ Each integration will set a default value based on the configuration options pas
...
+
+### extract
+
+Extract all generated `.css` into one bundle. Specify filename relative to `output.dir`.
+
+```ts
+vanillaExtractPlugin({
+  extract: 'app.css'
+});
+```
```
askoufis: Looks like the config options need to be updated to reflect the new config. Additionally, should the fact that `output.assetFileNames` is respected also be mentioned?

drwpow: Oh good catch. Added that note as well!

.changeset/twelve-suns-hope.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/rollup-plugin': minor
+---
+
+Add "extract" option which bundles CSS into one bundle. Removes .css imports.
```
askoufis, line 5:
```suggestion
Add "extract" option which bundles CSS into one bundle. Removes .css imports.

**EXAMPLE USAGE**:
```ts
vanillaExtractPlugin({
  extract: {
    name: 'bundle.css',
    sourcemap: false
  }
});
```

askoufis (approved):
Really appreciate the effort on this! I'll take a look at getting #1616 merged as well and the cut a release.

PR #1605, opened 2025-06-09
Update vite-node to v3.2.2

sapphi-red:
Vite 7 (currently in beta) includes a small breaking change that affects vite-node. Because of that, currently the plugin cannot be used with Vite 7.

This PR updates vite-node to v3.2.2 so that the plugin can be used with Vite 7. vite-node 3.2 is compatible with Vite 5 and Vite 6 so this change should not affect older versions.
It's still a beta version, but merging this PR will help keeping the ecosystem-ci green and helps us find any regression faster (it was caught indirectly via waku).

### DIFF
--- a/.changeset/old-dragons-sort.md
+++ b/.changeset/old-dragons-sort.md
@@ -1,0 +1,5 @@
+---
+"@vanilla-extract/compiler": patch
+---
+
+Update `vite-node` dependency to `^3.2.2`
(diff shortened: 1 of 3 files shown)


askoufis:
@sapphi-red This update was released in [`@vanilla-extract/vite-plugin@5.0.5`](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fvite-plugin%405.0.5).

PR #1614, opened 2025-06-16
`vite`: Add `inlineCssInDev` option to fix FOUC in SSR dev

askoufis:
Best reviewed [with hidden whitespace](https://github.com/vanilla-extract-css/vanilla-extract/pull/1614/files?diff=split&w=1).

During development, Vite SSR/SSG applications typically call [`ssrLoadModule`](https://vite.dev/guide/ssr#setting-up-the-dev-server) to load some sort of `render` entrypoint that's used to generate some initial HTML. The Vite plugin handles compilation of VE styles during `ssrLoadModule`, so all classnames are present in the DOM, but the CSS generated by VE is not.

Until Vite gains first-class support for injecting CSS during SSR, it looks like the common solution is to inject it yourself https://github.com/vitejs/vite/issues/2282. By iterating of the `cssCache` in the compiler, we can get all CSS generated during `ssrLoadModule`, and inject it into the `head`.

Consumers could further optimize their dev critical CSS if they want to with something like [`beasties`](https://github.com/danielroe/beasties).

In my eyes this replaces `unstable_mode: 'transform'`, at least during development. `transform` also suffered from falling back to the old compilation method, where as `inlineCssInDev` at least uses the `vite-node` compiler.

## Tests

While I wanted to add an SSR fixture to the repo, in its current state I think this would involve a decent refactor/some hacks to get working. I've tested this in [`sku`](https://github.com/seek-oss/sku) and am happy with the behaviour, so I'm ok with not having test coverage for the time being.

### DIFF
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -125,7 +125,10 @@ export function vanillaExtractPlugin({
           {
             tag: 'style',
             children: allCss,
-            attrs: { type: 'text/css' },
+            attrs: {
+              type: 'text/css',
+              'data-vanilla-extract-inline-dev-css': true,
+            },
             injectTo: 'head-prepend',
           },
         ];


.changeset/shiny-emus-guess.md
```diff
@@ -0,0 +1,28 @@
...
+```ts
+// vite.config.ts
+import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
+
+export default {
+  plugins: [
+    vanillaExtractPlugin({
+      unstable_mode: {
+        inlineCssInDev: true,
+      },
```
DanDroryAu: **Question:** Should this be `unstable_mode: 'inlineCssInDev'`.

The `Options` interface has the following key/value `unstable_mode?: 'transform' | 'emitCss' | 'inlineCssInDev';` for `unstable_mode`.

askoufis: yes, thanks copilot 🤦

.changeset/shiny-emus-guess.md
```diff
@@ -0,0 +1,28 @@
+---
+'@vanilla-extract/vite-plugin': minor
+---
+
+Add new `inlineCssInDev` option to `unstable_mode` configuration property
+
+Setting `unstableMode: inlineCssInDev` will result in all CSS generated by Vanilla Extract being inlined into a `style` element at the top of the document `head`. This feature is useful for preventing [FOUC] (Flash of Unstyled Content) when server-rendering your initial HTML. Without this, calling `ssrLoadModule` on your server entrypoint will not include the CSS in the HTML, leading to a FOUC when the client-side JavaScript takes over.
```
DanDroryAu: **Suggestion:** Missing quotes in the pre-block.

```suggestion
Setting `unstableMode: 'inlineCssInDev'` will result in all CSS generated by Vanilla Extract being inlined into a `style` element at the top of the document `head`. This feature is useful for preventing [FOUC] (Flash of Unstyled Content) when server-rendering your initial HTML. Without this, calling `ssrLoadModule` on your server entrypoint will not include the CSS in the HTML, leading to a FOUC when the client-side JavaScript takes over.
```

tests/compiler/compiler.test.ts
```diff
@@ -551,4 +555,24 @@ describe('compiler', () => {
...
     `);
   });
+
+  test('getAllCss should return all CSS that has been processed', async () => {
+    const compiler = compilers.getAllCss;
+
+    const buttonCssPath = './fixtures/class-composition/button.css.ts';
+    await compiler.processVanillaFile(buttonCssPath);
+
+    expect(compiler.getAllCss()).toMatchInlineSnapshot(`
```
DanDroryAu, line 567: **Suggestion:** Can we add a second test that shows all CSS processed from multiple files?

packages/vite-plugin/src/index.ts
```diff
@@ -109,162 +109,187 @@ export function vanillaExtractPlugin({
...
+          // The user is using a vite-based framework that has a custom config file
+          else {
+            configForViteCompiler = config.inlineConfig;
+          }
 
-      if (compiler) {
-        const absoluteId = getAbsoluteId(validId);
+          const viteConfig = {
+            ...configForViteCompiler,
+            plugins: configForViteCompiler?.plugins
```
DanDroryAu, line 183: **Praise:** Nice work. This should stop the plugins being initialized twice inside the compiler.

tests/compiler/compiler.test.ts
askoufis: Ended up adding a second file to the existing test. https://github.com/vanilla-extract-css/vanilla-extract/pull/1614/commits/757f55e1906571bfbcb1202956f2896d61d16152

PR #1616, opened 2025-06-18
allow plugin to work in rolldown

radnan:
rolldown does not expose importedIdResolutions similar to rollup and so we need to manually get it from importedIds and this.getModuleInfo


packages/rollup-plugin/src/index.ts
```diff
@@ -74,8 +74,9 @@ export function vanillaExtractPlugin({
     },
     // Emit .css assets
     moduleParsed(moduleInfo) {
-      moduleInfo.importedIdResolutions.forEach((resolution) => {
-        if (resolution.meta.css) {
+      moduleInfo.importedIds.forEach((id) => {
```
radnan, line 77: currently tying to use `rolldown` gives the following error:

image

but with this patched version we are able to successfully run the plugin in `rolldown`:

image

drwpow:
(not a maintainer, just interested in Vanilla Extract + Rollup support) This seems like a reasonable change, especially since it seems this is an intentional API departure on Rolldown’s end.

Would you be able to add a Rolldown test to ensure this a) works as intended and b) doesn’t regress? (again, not a maintainer, just also interested in this fix)

packages/rollup-plugin/test/rollup-plugin.test.ts
```diff
@@ -75,4 +79,76 @@ describe('rollup-plugin', () => {
       ]),
     ).toMatchSnapshot();
   });
+
+  describe('should be compatible with rolldown', () => {
```
radnan, line 187: added unit tests for `rolldown` @drwpow - thanks for the suggestion!

askoufis:
@radnan I just merged #1604 which introduced conflicts into your branch. I'm happy to take over this PR if you don't have bandwidth the work on it, but if you're happy to keep working on it I'll take a look once the branch is in order.

packages/rollup-plugin/src/lib.ts
```diff
@@ -16,17 +14,18 @@ export function generateCssBundle({
...
+      cssFiles[id] = buildImportChain(id, plugin);
     }
   }
 
   // 2. build bundle from import order
   for (const id of sortModules(cssFiles)) {
-    const { importedIdResolutions } = getModuleInfo(id) ?? {};
-    for (const resolution of importedIdResolutions ?? []) {
-      if (resolution.meta.css && !extractedCssIds.has(resolution.id)) {
+    const { importedIds } = plugin.getModuleInfo(id) ?? {};
```
radnan, line 25: thanks for offering to take it up @askoufis! but i resolved the conflicts - didn't look like there were too many of them.

i also duplicated the same extract test under rolldown describe section

and thanks again for taking the time to review this

packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap
```diff
@@ -1220,3 +1220,1270 @@ export { Radio as default };
...
+.utility_mt300__1vyatv82 {
+  margin-top: var(--size-space-300);
+}
+.utility_mt400__1vyatv83 {
+  margin-top: var(--size-space-400);
+}",
+  ],
+  [
+    "app.js",
+    "import "./styles/reset.css.js";
```
askoufis, line 1475: It looks like rolldown isn't stripping side-effect imports for some reason.

Here's the equivalent section of the rollup snapshot in `master`: https://github.com/vanilla-extract-css/vanilla-extract/blob/50f12348894a9e77a36ec269fcdf2ad159e5211d/packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap#L1107-L1115

Any thoughts @drwpow?

drwpow: Hm yeah we go in and remove those after-the-fact if someone is extracting `.css`. That happens [here, in generateCssBundle()](https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/rollup-plugin/src/lib.ts#L19):

```ts
  for (const id of getModuleIds()) {
    if (cssFileFilter.test(id)) {
      cssFiles[id] = buildImportChain(id, { getModuleInfo, warn });
    }
  }
```

Ultimately, if they get added to that `cssFiles` mapping, then we’ll strip the side effect imports later.

I haven’t debugged locally, but what is likely happening is Rolldown doesn’t have a complete graph of `this.getModuleIds()` at this point in time, so we don’t know that these are side-effects we generated (we obviously don’t touch user code as that would be Bad; we only will remove imports we know Vanilla Extract generated).

If, say, we debugged what the current status is of `this.getModuleIds()` at this point in time:

```diff
+   console.log(getModuleIds());
    for (const id of getModuleIds()) {
```

I’d expect to see that for whatever reason, Rollup and Rolldown are returning different things here.

If so, maybe it’s because Rolldown hasn’t yet built this graph, or it’s possibly partial, or something. All that tracks, because Rolldown has intentionally departed from some of Rollup’s API which led to this PR in the first place. I may not be able to take a look this week, but it could be as simple as calling `generateCssBundle()` at some other point in time, maybe.

radnan: I think I remember the file output is not exactly the same in rolldown as it is in rollup

So the file we're looking for might just have been renamed or something internally by rolldown

I guess the strategy to get extractedCssIds has to be looked at

drwpow-figma: Oh! Yeah if that’s the case then we can just expand our RegEx to look for both Rollup and Rolldown versions. That still should be pretty safe/I wouldn’t imagine any user code getting accidentally stripped from it (we have multiple checks to ensure we’re not just deleting user code, that’s only one layer).

radnan: ok so i see what's going on - we're stripping side effects manually - and previously in rollup it was kinda collapsing the modules where needed whereas in rolldown it's not doing that anymore

so rollup produces something like this:

```
- app.js
- reset.css
```

and `app.js` imports `reset.css` directly removing the interim `reset.css.ts` file since i guess rollup sees that the interm js file is not needed as it's just side-effects and doesn't export any actual js modules

whereas in rolldown we have:

```
- app.js
- reset.js
  - reset.css 
```

`app.js` is importing `reset.js` (which was `reset.css.ts`) and that in turn is importing `reset.css`. the side effects is now getting stripped from `reset.js` as you can see here:

https://github.com/radnan/vanilla-extract/blob/d385965a8057dff0ba8ecc9b8fe1de8e780db267/packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap#L1578-L1581

but we're left with an empty `reset.js` file that does nothing. i guess we'd have to do a second pass to manually remove these empty js modules - or just live with it

drwpow: Oh good find. Yeah I wonder given Rolldown’s current state and active iteration, if we just leave it as-is because I would strongly imagine others run into this, and this gets fixed on Rolldown’s side? Maybe?

drwpow: So I dug in more and it’s trickier than I thought. I’m pretty confident **it’s a bug on Rolldown’s end**—when modules are deleted, it doesn’t remove them from the graph and they stick around as imports. There’s not a good way for us to “clean up” on our end in our plugin, because it’s really Rolldown’s responsibility and something it should manage. Trying to detect empty files and deeply remove them from the module graph is not something a plugin should do.

So all that said, I think **this plugin in its current form is good enough for now!** I can file a bug on Rolldown’s end and follow up with that. I’m pretty confident this problem will go away on its own with Rolldown updates.

packages/rollup-plugin/test/rollup-plugin.test.ts
drwpow: Nit: these tests are nested under the `stripSideEffectImports` describe, which is unexpected. Maybe move these to the top level or somewhere else

packages/rollup-plugin/test/rollup-plugin.test.ts
```diff
@@ -215,4 +220,156 @@ export default function Button() {
...
+      const output = await build({
+        cwd,
+        extract: { name: 'app.css', sourcemap: true },
+        rollup: {
+          input: { app: path.join(cwd, 'src/index.ts') },
+          external: ['clsx', 'react', 'react/jsx-runtime', 'react-dom'],
+          output: {
+            assetFileNames: '[name][extname]',
+            format: 'esm',
+            preserveModules: true, // not needed for output, just makes some assertions easier
```
drwpow, line 239: The problem goes away, and Rolldown does fine if we just remove `preserveModules: true`. There’s some very very minute behavioral difference between Rollup + Rolldown that I’m not able to pinpoint 🙃. I spent some time today trying to reproduce it in Rolldown, but it seems to be a very hard to recreate issue where Rolldown generates an empty chunk. In many other scenarios it does fine.

I think it doesn’t degrade the test at all having `preserveModules: false` because as the comment states, that was never a central test goal. In fact, this is “alternate” bundler behavior, so I think we could even remove it for both Rollup and Rolldown tests.

Anyway, all that said, I don’t know _what_ the Rolldown bug is, but I can’t pinpoint it, and I still think it’s unrelated to this PR / this work.

packages/rollup-plugin/test/rollup-plugin.test.ts
radnan: oops - must've been from the rebase after that feature was merged - fixed

PR #1624, opened 2025-07-09
Support Vite 7

BPScott:
[Vite v7 was released on June 24th](https://github.com/vitejs/vite/releases/tag/v7.0.0). This PR adds support for the new major version by widening the peerDependencies range.

- Update `vitest` to `3.2.4` (vitest's vite v7 support was added in [v3.2.0](https://github.com/vitest-dev/vitest/releases/tag/v3.2.0).
- Update `vite-plugin-inspect` to `11.3.0` for vite v7 support (this version drops support for vite v5 but this is only used in test content and doesn't seem to affect anything)
- Add `|| ^7.0.0` to the `vite` peerDependencies.
- Use Vite v7 as the installed version when developing vanilla-extract.

vite-node was already bumped to a version that supports Vite v7 in #1605.

### Guidance requested

- The jest unit tests are failing, as the `"1/4": "styles_opacity_1/4__jteyb17"` key is no longer present in the output. I strongly suspect that this is because this change updates the installed version of rollup from v4.30.1 to v4.44.2 - thus including the changes to improved tree-shaking of unused object keys in [v4.34.0](https://github.com/rollup/rollup/releases/tag/v4.34.0). Adding a usage of the `1/4` class into the test fixture causes the unit tests to pass but then requires the playwright tests to be updated. What would you like to do here? Should we update the jest snapshot to remove that class or add an example of `1/4` usage in the fixture so that the classname is not removed (and then update the playwright test snapshots)? I'm leaning towards adding a usage of 1/4 in the test fixture - as currently these tests cease validating the behaviour that `fallbackVar` is working as expected (I don't know if there's other places where that is already covered)

- There is no `vite-plugin-inspect` version that supports Vite 5, 6 and 7. Our choices are `^0.8.x` which supports Vite 5 and 6, or `^11.3.0` which supports Vite 6 and 7. What should we do here? Do you have a preferred version? This is only used in test harnesses and both versions seem to work fine.

### DIFF
--- a/test-helpers/package.json
+++ b/test-helpers/package.json
@@ -42,8 +42,8 @@
     "prettier": "^2.8.8",
     "serve-handler": "^6.1.3",
     "style-loader": "^2.0.0",
-    "vite": "^5.0.0 || ^6.0.0",
-    "vite-plugin-inspect": "^0.8.1",
+    "vite": "^5.0.0 || ^6.0.0 || ^7.0.0",
+    "vite-plugin-inspect": "^11.3.0",
     "webpack": "^5.90.0",
     "webpack-dev-server": "^5.0.4",
     "webpack-merge": "^6.0.1"
--- a/.changeset/giant-apples-know.md
+++ b/.changeset/giant-apples-know.md
@@ -1,0 +1,6 @@
+---
+'@vanilla-extract/vite-plugin': patch
+'@vanilla-extract/compiler': patch
+---
+
+Include `^7.0.0` in `vite` dependency range
--- a/examples/remix/package.json
+++ b/examples/remix/package.json
@@ -21,6 +21,6 @@
     "@remix-run/dev": "^2.8.0",
     "@types/react": "^18.2.55",
     "@vanilla-extract/vite-plugin": "workspace:*",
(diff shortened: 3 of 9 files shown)


test-helpers/package.json
```diff
@@ -42,8 +42,8 @@
     "prettier": "^2.8.8",
     "serve-handler": "^6.1.3",
     "style-loader": "^2.0.0",
-    "vite": "^5.0.0 || ^6.0.0",
-    "vite-plugin-inspect": "^0.8.1",
+    "vite": "^5.0.0 || ^6.0.0 || ^7.0.0",
+    "vite-plugin-inspect": "^11.3.0",
```
askoufis, line 46: The `vite` version range is really only there to keep [`manypkg`](https://github.com/Thinkmill/manypkg) happy. I'm happy to stick with `^11.3.0` as we'll only actually use Vite@7 in the repo.

askoufis:
> The jest unit tests are failing

I agree that adding some usage of the `1/4` class is the better option.

askoufis (approved):
Thanks for the PR!

BPScott:
@askoufis thank you for the feedback!

PR #1627, opened 2025-07-31
fix: include content in css sourcemap when extracting

stipsan:
Hi 👋 

Without source map contents included [(which is the default behavior of rollup sourcemaps)](https://rollupjs.org/configuration-options/#output-sourcemapexcludesources) it makes them less useful. 

For example if you explore [the `app.css` and `app.css.map`](https://github.com/user-attachments/files/21537058/before.zip) for the `fixtures/react-library-example` workspace on [https://evanw.github.io/source-map-visualization] the mappings lead to empty files:
image

[With this PR the mappings now preserve the CSS that were initially omitted](https://github.com/user-attachments/files/21537218/after.zip):

image

Richer sourcemaps are especially useful when vanilla extract is the first step, and later on `lightningcss`, `cssnano` or `postcss` is operating on the extracted CSS asset. In those scenarios it's helpful to know what the CSS looked like before it was optimized and possibly deduped 😌 

Thanks so much for #1604 and #1616 btw. At Sanity we've decided to move away entirely from `styled-components`; `vanilla-extract` is shaping up to become its replacement. Since our tooling is leaning heavily on `rollup`, with the long term goal of moving to `rolldown`, these recent improvements were helpful and appreciated ❤

### DIFF
--- a/.changeset/breezy-poets-matter.md
+++ b/.changeset/breezy-poets-matter.md
@@ -2,4 +2,4 @@
 '@vanilla-extract/rollup-plugin': patch
 ---
 
-include content in css sourcemap when extracting
+Include content in CSS sourcemap when bundling with the `extract` option
--- a/.changeset/breezy-poets-matter.md
+++ b/.changeset/breezy-poets-matter.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/rollup-plugin': patch
+---
+
+include content in css sourcemap when extracting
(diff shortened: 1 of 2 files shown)


.changeset/breezy-poets-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/rollup-plugin': patch
+---
+
+include content in css sourcemap when extracting
```
askoufis:
```suggestion
Include content in CSS sourcemap when bundling with the `extract` option
```

askoufis (approved):
Thanks for the fix! Appreciate the example and explanation.

.changeset/breezy-poets-matter.md
askoufis: Ah, was going to commit this change from the github UI and merge immediately but I don't have permissions to push to your fork. Happy to merge once this change is committed.

PR #1564, opened 2025-03-23
Allow `packageInfo` to be undefined

emonadeo:
Fixes #1563.

IIRC, the name is only relevant for packages that publish to npm (#20).
This PR relaxes that requirement.

PR #1631, opened 2025-09-12
fix: support deno packages

marvinhagemeister:
This PR adds support for Deno projects which don't necessarily have a `package.json`. When a Deno environment is detect it will check for `deno.json` or `deno.jsonc` first before falling back to `package.json`.

Fixes https://github.com/denoland/fresh/issues/3392


jakubzet:
Is any contribution needed here? ;) Looks like this one also solves the issue: https://github.com/vanilla-extract-css/vanilla-extract/pull/1564

Issue #1636, opened 2025-10-04
Vite's `server.watch.ignore` patterns are not inherited by the compiler


PsychoLlama:
## Problem

If the project contains a huge directory or symlinks to a huge directory, a "hello world" project can take ~45 seconds or even minutes to launch `vite dev`.

Vite supports ignoring directories through the `server.watch.ignore` option. The Vanilla Extract compiler [doesn't pass it through](https://github.com/vanilla-extract-css/vanilla-extract/blob/e2c439ffa96dce570f2d472d2ca6ee40c69e3ee9/packages/compiler/src/compiler.ts#L113-L113). The parent process starts quickly but chokes as it waits for the child compiler to initialize.

> [!NOTE]
> There appears to be a race condition. I can repro this ~75% of the time with `bun`, and almost never with `node`. Maybe it has something to do with load order?

## Example

I ran into this because I use [Nix](https://nixos.org/) and [direnv](https://direnv.net/), although the bug isn't specific to either tool. It creates a `.direnv/` directory which includes a symlink to [nixpkgs](https://github.com/NixOS/nixpkgs). It's a big honkin' repo with ~50k files.

I can ignore that directory just fine (instant startup) until I add `@vanilla-extract/vite-plugin`. Then everything slows to a crawl.

```ts
import { defineConfig } from 'vite';
import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';

export default defineConfig({
  plugins: [vanillaExtractPlugin()],
  server: {
    watch: {
      ignored: ['**/.direnv/**'], // Not respected by vanilla extract.
    },
  },
});
```

This happens for any large directory or symlink not covered by the [default ignore patterns](https://vite.dev/config/server-options.html#server-watch).

## Hacky Workaround

[...]

PR #1637, opened 2025-10-05
Add bun as a package manager option

PsychoLlama:
I recently submitted [a new issue](https://github.com/vanilla-extract-css/vanilla-extract/issues/1636) relating to the bun runtime which has its own package manager. There was no option for bun in the template.

Template tested on my fork:

2025-10-04T22:26:39-04:00


askoufis:
Thank you!

PR #1638, opened 2025-10-05
[compiler] Inherit vite server watch options

### LINKED ISSUE #1636: Vite's `server.watch.ignore` patterns are not inherited by the compiler

PsychoLlama:
This fixes #1636 by inheriting the user's `server.watch` configuration, notably ignored watch patterns.

Does this break other use cases? No idea. Scrutiny required.

### DIFF
--- a/packages/compiler/src/compiler.ts
+++ b/packages/compiler/src/compiler.ts
@@ -110,7 +110,7 @@ const createViteServer = async ({
     root,
     server: {
       hmr: false,
-      watch: enableFileWatcher ? undefined : null,
+      watch: enableFileWatcher ? viteConfig.server?.watch : null,
     },
     logLevel: 'silent',
     optimizeDeps: {


askoufis:
> Does this break other use cases?

I don't think so. Inheriting the entire `server` config has been proposed elsewhere, and I'm more hesitant about that. Inheriting just the `watch` config makes sense to me.

PR #1639, opened 2025-10-08
turbopack support

### LINKED ISSUE #1367: Cannot build a next app with Turbopack (Styles were unable to be assigned to a file)
### Describe the bug

I'm trying to run my `next` app with the `--turbo` flag as it drastically improves build times.
This builds the project using [Turbopack](https://turbo.build/pack). This tool is still in beta, but more than 99% of its tests are passing: https://areweturboyet.com/. I never got any issues in projects that are not using vanilla extract.

Unfortunately running `next dev --turbo` will cause the app to fail with the following error:
image

I created this simple reproduction repo, where I just ran `create-next-app` and installed vanilla extract:
https://github.com/nicolassanmar/vanilla-extract-turbopack-issue/tree/main

### Reproduction

https://github.com/nicolassanmar/vanilla-extract-turbopack-issue/tree/main

### System Info

```shell
System:
    OS: Linux 5.15 Zorin OS 16.3
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 4.26 GB / 23.27 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.18.2 - ~/.nvm/versions/node/v18.18.2/bin/node
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.2/bin/npm
    pnpm: 8.13.1 - ~/.local/share/pnpm/pnpm
  Browsers:
    Chrome: 122.0.6261.94
  npmPackages:
    @vanilla-extract/css: ^1.14.1 => 1.14.1 
    @vanilla-extract/next-plugin: ^2.3.7 => 2.3.7
```

### Used Package Manager

pnpm

### Logs

```shell
✓ Compiled / in 899ms
 ⨯ Error: Styles were unable to be assigned to a file. This is generally caused by one of the following:

- You may have created styles outside of a '.css.ts' context
- You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started
    at Object.getFileScope (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/@vanilla-extract+css@1.14.1/node_modules/@vanilla-extract/css/fileScope/dist/vanilla-extract-css-fileScope.cjs.dev.js:35:11)
    at generateIdentifier (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/@vanilla-extract+css@1.14.1/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.cjs.dev.js:175:49)
    at Object.style (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/@vanilla-extract+css@1.14.1/node_modules/@vanilla-extract/css/dist/vanilla-extract-css.cjs.dev.js:374:19)
    at /home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[root of the server]__87cd9b._.js:125:86
    at [project]/styles/vanilla-extract.css.ts [ssr] (ecmascript) (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[root of the server]__87cd9b._.js:129:3)
    at instantiateModule (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:488:23)
    at getOrInstantiateModuleFromParent (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:539:12)
    at esmImport (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:113:20)
    at /home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[root of the server]__87cd9b._.js:141:132
    at [project]/pages/index.tsx [ssr] (ecmascript) (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[root of the server]__87cd9b._.js:430:3)
    at instantiateModule (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:488:23)
    at getOrInstantiateModuleFromParent (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:539:12)
    at esmImport (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:113:20)
    at /home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/node_modules__pnpm_af8e7b._.js:6683:112
    at [project]/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/esm/build/templates/pages.js/(INNER_PAGE)/[project]/pages/index.tsx [ssr] (ecmascript) (ecmascript) (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/node_modules__pnpm_af8e7b._.js:6718:3)
    at instantiateModule (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:488:23)
    at instantiateRuntimeModule (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:547:12)
    at Object.getOrInstantiateRuntimeModule (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/chunks/[turbopack]_runtime.js:562:12)
    at Object.<anonymous> (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/.next/server/pages/index.js:6:26)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at mod.require (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/require-hook.js:65:28)
    at require (node:internal/modules/cjs/helpers:119:18)
    at requirePage (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/require.js:109:84)
    at /home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/load-components.js:98:84
    at async loadComponentsImpl (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/load-components.js:98:26)
    at async DevServer.findPageComponentsImpl (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/next-server.js:664:36)
    at async DevServer.findPageComponents (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/dev/next-dev-server.js:565:20)
    at async DevServer.renderPageComponent (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/base-server.js:1832:24)
    at async DevServer.renderToResponseImpl (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/base-server.js:1882:32)
    at async DevServer.pipeImpl (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/base-server.js:895:25)
    at async NextNodeServer.handleCatchallRenderRequest (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/next-server.js:269:17)
    at async DevServer.handleRequestImpl (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/base-server.js:791:17)
    at async /home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/dev/next-dev-server.js:331:20
    at async Span.traceAsyncFn (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/trace/trace.js:151:20)
    at async DevServer.handleRequest (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/dev/next-dev-server.js:328:24)
    at async invokeRender (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/lib/router-server.js:174:21)
    at async handleRequest (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/lib/router-server.js:353:24)
    at async requestHandlerImpl (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/lib/router-server.js:377:13)
    at async Server.requestListener (/home/nicolassanmar/trabajo/repos/vanilla-extract-turbopack-issue/node_modules/.pnpm/next@14.1.4_@babel+core@7.24.3_react-dom@18.2.0_react@18.2.0/node_modules/next/dist/server/lib/start-server.js:140:13) {
  page: '/'
}
```

### Validations

RJWadley:
adds turbopack support to `@vanilla-extract/next-plugin`

turbopack doesn't have any sort of external API yet, so in order to evaluate the contents of our `*.css.ts` files we need to compile it separately using `@vanilla-extract/compiler`. To avoid import resolution mismatches, we defer module resolution to turbopack via its webpack-compatible api.

`next/font` doesn't work out of the box because turbopack has custom logic for it, so we transform any modules that use it in a css context to contain static references to the generated font family, style, and weight. `next/image` gets a similar treatment. most other next modules won't work in `css.ts` files for now.

fixes #1367

### DIFF
--- a/site/docs/integrations/next.md
+++ b/site/docs/integrations/next.md
@@ -15,7 +15,8 @@ npm install --save-dev @vanilla-extract/next-plugin
 
 ## Setup
 
-If you don't have a `next.config.js` or `next.config.ts` file in the root of your project, create one. Add the plugin to your Next.js config file.
+If you don't have a `next.config.js` or `next.config.ts` file in the root of your project, create one.
+Then add the plugin to your Next.js config file:
 
 ```ts
 // next.config.ts
@@ -50,7 +51,7 @@ export default withVanillaExtract(withMDX(nextConfig));
 ## Version Support
 
 - **Next.js >= 16.x**: Both Turbopack and Webpack are supported
-- **Next.js <= 15.x**: Webpack-only support
+- **Next.js <= 15.x**: Only Webpack is supported
 
 ## Configuration
 
@@ -79,24 +80,20 @@ Each integration will set a default value based on the configuration options pas
 
 > ⚠️&nbsp;&nbsp;Turbopack support is experimental. Its API is unstable and may undergo breaking changes in non-major versions. Additionally, it may not handle all features supported by Next.js.
 
-Turbopack-related options are grouped under the `unstable_turbopack` key.
+You can control Turbopack autoconfiguration using `unstable_turbopack.mode`:
 
-#### unstable_turbopack.mode
-
-You can control Turbopack autoconfiguration using `mode`:
-
-- `auto` (default): enable Turbopack config only when Next >= 16.0.0
+- `off`: (default) never configure Turbopack (i.e. use Webpack)
+- `auto`: enable Turbopack config only when Next >= 16.0.0
 - `on`: force-enable Turbopack config
-- `off`: never configure Turbopack (Webpack-only)
 
-For example, to disable Turbopack integration explicitly:
+For example:
 
 ```ts
 // next.config.ts
 import { createVanillaExtractPlugin } from '@vanilla-extract/next-plugin';
 
 const withVanillaExtract = createVanillaExtractPlugin({
-  unstable_turbopack: { mode: 'off' }
+  unstable_turbopack: { mode: 'auto' }
 });
 
 export default withVanillaExtract({});
@@ -104,9 +101,8 @@ export default withVanillaExtract({});
 
 If you already manage `turbopack.rules` yourself for the same file globs, the plugin may throw to avoid rule conflicts. In that case, set `mode: 'off'` and apply your Turbopack config manually.
 
-#### unstable_turbopack.glob
-
-By default Turbopack integration processes `**/*.css.{js,cjs,mjs,jsx,ts,tsx}`. You can override this via `glob`:
+By default Turbopack integration processes `**/*.css.{js,cjs,mjs,jsx,ts,tsx}` files.
+You can override this with `unstable_turbopack.glob`:
 
 ```ts
 // next.config.ts
@@ -114,6 +110,7 @@ import { createVanillaExtractPlugin } from '@vanilla-extract/next-plugin';
 
 const withVanillaExtract = createVanillaExtractPlugin({
   unstable_turbopack: {
+    mode: 'auto',
     glob: ['**/*.css.ts', '**/*.css.tsx']
   }
 });
@@ -151,4 +148,4 @@ const nextConfig: NextConfig = {
 export default withVanillaExtract(nextConfig);
 ```
 
-[`transpilepackages`]: https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages
+[`transpilePackages`]: https://nextjs.org/docs/app/api-reference/next-config-js/transpilePackages


csorfab:
@mattcompiles @askoufis Can anyone take a look at this? This is a pretty important feature for next.js users. Or should we consider the project dead? Last PR was merged in early August, basically no activity from maintainers since then. Not meaning to bash anyone, I know it's thankless work maintaining open source software, but it would be nice to know what we can expect from this project. Thanks in advance!

askoufis:
> Or should we consider the project dead? Last PR was merged in early August, basically no activity from maintainers since then.

A PR was merged in September. Reviews take time and ongoing work is often slow. All current maintainers have professional commitments and/or families.

I try to look at PRs when I have spare time/motivation, which is often quite sporadic. I'm currently on holiday, but I intend to make some headway on open PRs and repo maintenance in when I get back in early November.

csorfab:
Thanks for your reply! I was a bit too dramatic, sorry about that. Enjoy your holiday, and let me know if I can help with anything to get this PR going.

callum-gander:
I see there have been some recent commits pushed up to this PR. Are there any bits that you need help with that could be provided to speed up getting this merged in?

Appreciate that it's been a holiday period, but a lot of apps have performance issues with Next which has various potential solutions but the main recommendation seems to boil down to "just use Turbopack" https://github.com/vercel/next.js/issues/48748. Obviously, the lack of support for Turbopack with Vanilla is effectively blocking us from getting those gains, which would be greatly appreciated ahead of Black Friday.

It would be great if there were a list of tasks that I or people within my company could help out with to speed this up, whether that's commits or just QAing. Cheers and thanks for the work on this so far!

RJWadley:
If you're interested in trying this out, I've been using a `preinstall` script to download and install this branch:
```
"preinstall": "pnpm dlx gitpick https://github.com/RJWadley/vanilla-extract/tree/turbo-loader/ /tmp/vanilla-temp -f && cd /tmp/vanilla-temp && pnpm i && pnpm build"
```
You can install that temporary build with a link:
```
"@vanilla-extract/next-plugin": "link:/tmp/vanilla-temp/packages/next-plugin",
```

Although, you probably want to pin to a specific commit instead of the branch for the sake of stability:
```
"preinstall": "pnpm dlx gitpick RJWadley/vanilla-extract -b f5163f1 /tmp/vanilla-temp -f && cd /tmp/vanilla-temp && pnpm i && pnpm build"
```
Feedback or issues appreciated! I haven't seen many stats yet on speed or size of builds (compared to webpack) so if you have those I'd love to see some details.

----

The biggest issue with this at the moment is `next` specific modules not playing nicely. Since we must process styles outside of next.js there's not much that can be done. Think `next/font`, `next/image`, etc. Next has decent webpack loader support, but there are [a few things missing](https://x.com/timneutkens/status/1983161149379227805) that make it much harder (and in some cases impossible) for us to preprocess styles.

I'm gonna do a quick cleanup pass this week to prep this for merge, it's basically ready at this point but is still a bit messy while I'm experimenting with some stuff.

TLDR:
- if your styles don't use next modules at all everything will work fine
- next/font partially works
- next/image imports don't work at all
- other next modules may or may not work

callum-gander:
Hmmm, strange, I followed your steps and got this error
```
@frontend/app:dev:    ▲ Next.js 15.5.0
@frontend/app:dev:    - Local:        http://localhost:3000
@frontend/app:dev:    - Network:      http://10.100.22.2:3000
@frontend/app:dev:    - Environments: .env
@frontend/app:dev:    - Experiments (use with caution):
@frontend/app:dev:      ⨯ reactCompiler
@frontend/app:dev:      · optimizePackageImports
@frontend/app:dev: 
@frontend/app:dev:  ✓ Starting...
@frontend/app:dev: Warning: total number of custom routes exceeds 1000, this can reduce performance. Route counts:
@frontend/app:dev: headers: 5
@frontend/app:dev: rewrites: 2
@frontend/app:dev: redirects: 1246
@frontend/app:dev: See more info: https://nextjs.org/docs/messages/max-custom-routes-reached
@frontend/app:dev: Error: > Couldn't find a `pages` directory. Please create one under the project root
@frontend/app:dev:     at Object.webpack (../../../../../../../../private/tmp/vanilla-temp/packages/next-plugin/dist/vanilla-extract-next-plugin.cjs.dev.js:130:47)
```
That seemed to point at this specific line in the dist
```
 const findPagesDirResult = findPagesDir.findPagesDir(dir, ((_resolvedNextConfig$e = resolvedNextConfig.experimental) === null || _resolvedNextConfig$e === void 0 ? void 0 : _resolvedNextConfig$e.appDir) ?? false);
```
It's probably worth saying that we don't have a pages dir at all, we're fully on app dir. To be fair as well, I'm not sure that this would work for us anyway as we make heavy use of `next/image`, `next/font`, as well as other next modules. 

When you say it's basically ready, do you mean basically ready _minus_ support for next modules due to dependency on things turbopack isn't yet providing to help you preprocess styles? 

If you have a solution to the above error, more than happy to give it a whirl

RJWadley:
The error looks like resolution error in the webpack plugin setup. Should've seen that coming.
Switch back to the release version of next-plugin, then try configuring the loader manually:
```
"@vanilla-extract/turbopack-plugin": "link:/tmp/vanilla-temp/packages/turbopack-plugin",
```
```
  turbopack: {
    rules: {
      '**/*.css.{ts,tsx,js,jsx}': {
        loaders: [
          {
            loader: '@vanilla-extract/turbopack-plugin',
            options: { nextEnv: {} }, // include env vars set by next.config if any
          },
        ],
      },
    },
  },
  ```

----

As far as next/image support, you're free to style next images as this doesn't route through the plugin:
```
// page.ts
<Image className={myVanillaClass} />
```

The specific scenario that doesn't work is importing an image into a `css.ts` file and using it to generate styles:
```
// myStyles.css.ts
import coolThing from "./my-cool-image.png"

export const background = style({
  backgroundImage: `url(${coolThing.src})` // fails! use public dir instead
})
```

This is unlikely to change until we get better tools from turbopack.

For `next/font`, support is better. Anything under the `style` property is supported:
```
// myStyles.css.ts
import myFont from "./typography"

export const background = style({
  fontFamily: myFont.style.fontFamily
})
```

The specific scenario that doesn't work is extending font classNames in your styles:
```
// myStyles.css.ts
import myFont from "./typography"

export const background = style(myFont.className) // fails! set font family directly instead
```
This is also unlikely to change until we get better tools from turbopack.



TLDR using next modules is fine, but using next modules in your styles themselves is likely to fail unless I've specifically worked around it.

**Comment by Pnlvfx** [comment]
> The error looks like resolution error in the webpack plugin setup. Should've seen that coming. Switch back to the release version of next-plugin, then try configuring the loader manually:
> 
> ```
> "@vanilla-extract/turbopack-plugin": "link:/tmp/vanilla-temp/packages/turbopack-plugin",
> ```
> 
> ```
>   turbopack: {
>     rules: {
>       '**/*.css.{ts,tsx,js,jsx}': {
>         loaders: [
>           {
>             loader: '@vanilla-extract/turbopack-plugin',
>             options: { nextEnv: {} }, // include env vars set by next.config if any
>           },
>         ],
>       },
>     },
>   },
> ```
> 
> As far as next/image support, you're free to style next images as this doesn't route through the plugin:
> 
> ```
> // page.ts
> <Image className={myVanillaClass} />
> ```
> 
> The specific scenario that doesn't work is importing an image into a `css.ts` file and using it to generate styles:
> 
> ```
> // myStyles.css.ts
> import coolThing from "./my-cool-image.png"
> 
> export const background = style({
>   backgroundImage: `url(${coolThing.src})` // fails! use public dir instead
> })
> ```
> 
> This is unlikely to change until we get better tools from turbopack.
> 
> For `next/font`, support is better. Anything under the `style` property is supported:
> 
> ```
> // myStyles.css.ts
> import myFont from "./typography"
> 
> export const background = style({
>   fontFamily: myFont.style.fontFamily
> })
> ```
> 
> The specific scenario that doesn't work is extending font classNames in your styles:
> 
> ```
> // myStyles.css.ts
> import myFont from "./typography"
> 
> export const background = style(myFont.className) // fails! set font family directly instead
> ```
> 
> This is also unlikely to change until we get better tools from turbopack.
> 
> TLDR using next modules is fine, but using next modules in your styles themselves is likely to fail unless I've specifically worked around it.

For me is fine, thanks, I also had next font issues with webpack so I don't mind this approach, 

thank you very much for your work,

I'll try this ASAP

**Comment by callum-gander** [comment]
> The error looks like resolution error in the webpack plugin setup. Should've seen that coming. Switch back to the release version of next-plugin, then try configuring the loader manually:
> 
> ```
> "@vanilla-extract/turbopack-plugin": "link:/tmp/vanilla-temp/packages/turbopack-plugin",
> ```
> 
> ```
>   turbopack: {
>     rules: {
>       '**/*.css.{ts,tsx,js,jsx}': {
>         loaders: [
>           {
>             loader: '@vanilla-extract/turbopack-plugin',
>             options: { nextEnv: {} }, // include env vars set by next.config if any
>           },
>         ],
>       },
>     },
>   },
> ```
> 
> As far as next/image support, you're free to style next images as this doesn't route through the plugin:
> 
> ```
> // page.ts
> <Image className={myVanillaClass} />
> ```
> 
> The specific scenario that doesn't work is importing an image into a `css.ts` file and using it to generate styles:
> 
> ```
> // myStyles.css.ts
> import coolThing from "./my-cool-image.png"
> 
> export const background = style({
>   backgroundImage: `url(${coolThing.src})` // fails! use public dir instead
> })
> ```
> 
> This is unlikely to change until we get better tools from turbopack.
> 
> For `next/font`, support is better. Anything under the `style` property is supported:
> 
> ```
> // myStyles.css.ts
> import myFont from "./typography"
> 
> export const background = style({
>   fontFamily: myFont.style.fontFamily
> })
> ```
> 
> The specific scenario that doesn't work is extending font classNames in your styles:
> 
> ```
> // myStyles.css.ts
> import myFont from "./typography"
> 
> export const background = style(myFont.className) // fails! set font family directly instead
> ```
> 
> This is also unlikely to change until we get better tools from turbopack.
> 
> TLDR using next modules is fine, but using next modules in your styles themselves is likely to fail unless I've specifically worked around it.

Hmmm, so I gave this a try and I was getting `.readFile` related errors, which I believe was due to the fact that your fixture uses Next 16, whereas we were still on Next 15, so I bumped ours to Next 16 and followed the rest of your steps. It seemed to be fine until we actually tried to load a page and we're now getting these errors
```
@frontend/app:dev:  ○ Compiling /[store]/[...404] ...
@frontend/app:dev:  ⨯ Error: Styles were unable to be assigned to a file. This is generally caused by one of the following:
@frontend/app:dev: 
@frontend/app:dev: - You may have created styles outside of a '.css.ts' context
@frontend/app:dev: - You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started
@frontend/app:dev:     at module evaluation (src/lib/theme/theme.css.ts:15:46)
@frontend/app:dev:     at module evaluation (src/app/[store]/layout.tsx:2:1)
@frontend/app:dev:   13 | } from ".";
@frontend/app:dev:   14 |
@frontend/app:dev: > 15 | export const [themeClass, vars] = createTheme({
@frontend/app:dev:      |                                              ^
@frontend/app:dev:   16 |   color: { ...colors },
@frontend/app:dev:   17 |   space: { ...spacings },
@frontend/app:dev:   18 |   zIndex: { ...zIndexes }, {
@frontend/app:dev:   type: 'Error',
@frontend/app:dev:   page: '/en/theme-test'
@frontend/app:dev: }
@frontend/app:dev:  GET /theme-test 500 in 7.5s (compile: 7.4s, proxy.ts: 56ms, render: 63ms)
@frontend/app:dev:  ⨯ Error: Styles were unable to be assigned to a file. This is generally caused by one of the following:
@frontend/app:dev: 
@frontend/app:dev: - You may have created styles outside of a '.css.ts' context
@frontend/app:dev: - You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started
@frontend/app:dev:     at module evaluation (src/lib/theme/theme.css.ts:15:46)
@frontend/app:dev:     at module evaluation (src/app/[store]/layout.tsx:2:1)
@frontend/app:dev:   13 | } from ".";
@frontend/app:dev:   14 |
@frontend/app:dev: > 15 | export const [themeClass, vars] = createTheme({
@frontend/app:dev:      |                                              ^
@frontend/app:dev:   16 |   color: { ...colors },
@frontend/app:dev:   17 |   space: { ...spacings },
@frontend/app:dev:   18 |   zIndex: { ...zIndexes }, {
@frontend/app:dev:   type: 'Error',
@frontend/app:dev:   page: '/en'
@frontend/app:dev: }
@frontend/app:dev:  GET / 500 in 4.3s (compile: 4.3s, proxy.ts: 8ms, render: 5ms)
``` 
That's obviously this `import { createTheme } from "@vanilla-extract/css";` createTheme. Not sure if again this is due to a misconfiguration, followed your instructions to a tee. I pulled your PR and basically copied over the theme files and created a new test page in the `next-16-app-pages-router` and it seemed to be working fine so I'm not sure if it is somehow an issue with how I've been testing this

**Comment by RJWadley** [comment]
yes, I dropped support for next 15 in [c1146fe](https://github.com/vanilla-extract-css/vanilla-extract/pull/1639/commits/c1146fe61c8d8c00e2740d44ce34fedc88675ef6) because it dramatically simplifies dependency tracking when the source code has errors.

From the stack trace your code looks good, so it's more likely an issue with the plugin config than an issue with your code. most likely the plugin isn't running on the theme file. if you can manage to throw together a repo that does throw I'd be happy to investigate further. more minimal is obviously better but anything that fails would help diagnose the issue.

Thanks for testing this out!

**Comment by RJWadley** [comment]
cc @askoufis would love a review on this whenever you have time for it. if we can get an initial version of this published it'll make it much easier for folks to try it and report feedback.

**Comment by callum-gander** [comment]
<strike>So, following your steps I managed to get it all working, and the homepage of our website loaded styles and all, but as soon as we moved page, we get the below error
```
Parsing CSS source code failed
  3 |   width: 100%;
  4 |   position: relative;
> 5 |   screen and (min-width: 47.99rem) {
    |         ^
  6 |     min-height: 564px;
  7 |   }
  8 | }

Unexpected token Ident("and")

Import trace:
  Client Component Browser:
    ./workspaces/apps/frontend/src/packages/ui-shared/components/CarouselLoader/data:5da1d5 [Client Component Browser]
    ./workspaces/apps/frontend/src/packages/ui-shared/components/CarouselLoader/CarouselLoader.css.ts [Client Component Browser]
    ./workspaces/apps/frontend/src/packages/ui-shared/components/CarouselLoader/CarouselLoader.tsx [Client Component Browser]
    ./workspaces/apps/frontend/src/components/Basket/components/Carousels/Carousels.tsx [Client Component Browser]
    ./workspaces/apps/frontend/src/components/Basket/components/Basket.tsx [Client Component Browser]
    ./workspaces/apps/frontend/src/components/Basket/BasketWrapper.tsx [Client Component Browser]
    ./workspaces/apps/frontend/src/components/Basket/BasketWrapper.tsx [Server Component]
    ./workspaces/apps/frontend/src/app/[store]/basket/page.tsx [Server Component]
```
We declare our breakpoints as seen below that we then export them from the same file as our theme
```
export const mediaQueries = {
  mobile: `screen and (max-width: ${breakpoints?.sm})`,
  tablet: `screen and (min-width: ${breakpoints?.md}) and (max-width: ${breakpoints?.lg})`,
  smallDesktop: `screen and (min-width: ${breakpoints?.md})`,
  mdDesktop: `screen and (min-width: ${breakpoints?.mdlg})`,
  desktop: `screen and (min-width: ${breakpoints?.lg})`,
  desktopLarge: `screen and (min-width: ${breakpoints?.xl})`
};
```
I'm presuming before that vanilla was automatically appending `@media` to those queries, but using the turbo plugin from this PR it seems to not be? 

I got your PR running locally and moved the code over to validate this and we seem to get the same issue
```
 ⨯ ./fixtures/next-16-app-pages-router/src/components/data:fa392e:6:9
Parsing CSS source code failed
  4 |   position: relative;
  5 |   background-color: lightblue;
> 6 |   screen and (min-width: 47.99rem) {
    |         ^
  7 |     min-height: 200px;
  8 |     background-color: lightcoral;
  9 |   }

Unexpected token Ident("and")

Import trace:
  Client Component Browser:
    ./fixtures/next-16-app-pages-router/src/components/data:fa392e [Client Component Browser]
    ./fixtures/next-16-app-pages-router/src/components/MediaQueryTest.css.ts.js [Server Component]
    ./fixtures/next-16-app-pages-router/src/components/MediaQueryTest.tsx [Server Component]
    ./fixtures/next-16-app-pages-router/src/app/page.tsx [Server Component]


 GET / 500 in 1938ms (compile: 1870ms, render: 68ms)
```
The obvious fix here is just adding `@media` ourselves, and that works on your PR, the page renders as expected

However, in our repo, this doesn't happen and we get this, seemingly appending `@media` twice? 
```
@frontend/app:dev: Error evaluating Node.js code
@frontend/app:dev: Error: Invalid media query: "@media @media screen and (min-width: 47.99rem) and (max-width: 79.99rem)"
@frontend/app:dev: 
@frontend/app:dev: Expected media condition or media prefix
@frontend/app:dev: 
@frontend/app:dev: Read more on MDN: https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries
@frontend/app:dev:     [at createMediaQueryError (/private/tmp/vanilla-temp/packages/css/dist/transformCss-5ed3d195.cjs.dev.js:474:50)]
@frontend/app:dev:     [at validateMediaQuery (/private/tmp/vanilla-temp/packages/css/dist/transformCss-5ed3d195.cjs.dev.js:483:11)]
@frontend/app:dev:     [at Stylesheet.transformMedia (/private/tmp/vanilla-temp/packages/css/dist/transformCss-5ed3d195.cjs.dev.js:768:9)]
@frontend/app:dev:     [at Stylesheet.processCssObj (/private/tmp/vanilla-temp/packages/css/dist/transformCss-5ed3d195.cjs.dev.js:601:12)]
@frontend/app:dev:     [at Object.transformCss (/private/tmp/vanilla-temp/packages/css/dist/transformCss-5ed3d195.cjs.dev.js:958:16)]
@frontend/app:dev:     [at /private/tmp/vanilla-temp/packages/compiler/dist/vanilla-extract-compiler.cjs.dev.js:340:34]
@frontend/app:dev:     [at async queueFn (/private/tmp/vanilla-temp/packages/compiler/dist/vanilla-extract-compiler.cjs.dev.js:20:24)]
@frontend/app:dev:     [at async processQueue (/private/tmp/vanilla-temp/packages/compiler/dist/vanilla-extract-compiler.cjs.dev.js:41:3)]
@frontend/app:dev: 
```
I've double checked our config to see if anything else might be doing this and I can't see any reason why it would?

Beyond the turbo plugin, is there nothing else that needs to be added on my end? That's the only obvious thing I could think of.

Here's the changes https://github.com/callum-gander/vanilla-extract-turbo-test/commit/80e23e4ee367f233b60f7f4096a3ed02b2c8c6e3</strike>

Ignore the above, there was some... unique uses of css within our codebase that seemingly broke this, once they were fixed, this all worked great! Thanks again for your work on this and all the help!

**Comment by gavinthomas-valtech** [comment]
When is this is likely to be merged? more testing required?

**Comment by RJWadley** [comment]
Try out the prerelease using overrides:

It's built from [this branch](https://github.com/RJWadley/vanilla-extract/commits/turbo-merge-1) and is a merge of this PR with #1659

```jsonc
// package.json
"pnpm": {
  "overrides": {
    "@vanilla-extract/compiler": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/compiler@c1d4a88",
    "@vanilla-extract/css": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/css@c1d4a88",
    "@vanilla-extract/integration": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/integration@c1d4a88",
    "@vanilla-extract/next-plugin": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/next-plugin@c1d4a88",
    "@vanilla-extract/turbopack-plugin": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/turbopack-plugin@c1d4a88",
    "@vanilla-extract/webpack-plugin": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/webpack-plugin@c1d4a88"
  },
}
```

If you have issues, do let me know! As far as merging, next step is a review from a maintainer.

@askoufis let me know if there's anything I can do to make review easier! Happy to answer questions, revert bits, pull bits into their own PRs, etc. where possible for easier review. I know by nature it's a bit of a large PR so happy to do what I can to make it easier for you :)

exequielbc:
> The specific scenario that doesn't work is importing an image into a css.ts file and using it to generate styles:
>
> ```// myStyles.css.ts
> import coolThing from "./my-cool-image.png"
>
> export const background = style({
>  backgroundImage: `url(${coolThing.src})` // fails! use public dir instead
> })
> ```
> This is unlikely to change until we get better tools from turbopack.

@RJWadley the Panda CSS library gets around this by telling users to use the `style` prop on elements, [their justification](https://panda-css.com/docs/overview/faq#imported-image-is-not-working-in-vite-app) seems _good enough_ to me. Built-in support _could_ be added later if Turbo APIs become more open?

Sorry in advance if this wouldn't work, I thought it was a potential way to unblock this particular issue. It'd need to be a breaking change for `@vanilla-extract/next-plugin` if this was previously supported.

gavinthomas-valtech:
btw. Wanted to call out the fantastic work of this library. I love how it works

RJWadley:
@exequielbc ended up finding an approach that works for now. Imports from images imports now work as expected:

```ts
import coolThing from "./my-cool-image.png"

export const background = style({
  backgroundImage: `url(${coolThing.src})`, // works ✅
  width: `${coolThing.width}px`, // works ✅
  height: `${coolThing.height}px`, // works ✅
})
```

Perhaps worth noting, this is just a stub for `next/image`. Very advanced cases will still fail. For example, if you configured a fancy custom turbopack loader in `turbopack.rules` to import powerpoint files, that won't work in `css.ts` files (and I'm not planning to implement this until turbopack supports [module loading](https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack#missing-webpack-loader-features:~:text=Module%20loading)).

Try out the updated prerelease using overrides: 
(built from [this branch](https://github.com/RJWadley/vanilla-extract/commits/turbo-merge-1))
```jsonc
// package.json
"pnpm": {
  "overrides": {
    "@vanilla-extract/compiler": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/compiler@edaedbb",
    "@vanilla-extract/css": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/css@edaedbb",
    "@vanilla-extract/integration": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/integration@edaedbb",
    "@vanilla-extract/next-plugin": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/next-plugin@edaedbb",
    "@vanilla-extract/turbopack-plugin": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/turbopack-plugin@edaedbb",
    "@vanilla-extract/webpack-plugin": "https://pkg.pr.new/RJWadley/vanilla-extract/@vanilla-extract/webpack-plugin@edaedbb"
  },
}
```

serhat-m:
@RJWadley appreciate your work. Gave it a try with Bun and Turbopack and no issues so far  👍

hallatore:
@RJWadley test package seems to work great!
Any timeline for merging this into main? Vanilla-extract is the only package in our project not working with turbopack atm.

IsaacAndela:
I've tested the above mentioned prerelease branch in a project with over 8000 LOC in `.css.tsx` files.

I asked our QA guy to do a regression test and everything looks good.

I appreciate all the effort you are putting into this! 😃

Danny-A:
What needs to be done to get this thing merged? We are depending on this to move forward and could assist to make sure it gets in.

talipatpace:
Any updates on this?

We would also like to have this merged so we can switch to turbopack.

We are also willing to help out, if there are still open points that need to be addressed.

nandahibatullah:
Hello, when is this planned on being approved and release?

This would be so good to get in soon, tested it out with this branch and it seemed to be working great.

nandahibatullah (approved):

dy0gu:
I've also tested these changes in a relatively complex Turbopack Next project and everything worked flawlessly.

dasutyuzh:
> I've also tested these changes in a relatively complex Turbopack Next project and everything worked flawlessly.

I have not but am invested in this thread :-) how can I persuade a merge?

jksaunders:
Also very invested in this thread!

gardsa:
Along with many others, this PR is a blocker for us moving fully onto Next with Turbopack (we’re otherwise on Webpack due to lack of support here).

Could a maintainer share a rough status or timeline for review/merge, or point to anything still needed from contributors?

We’re happy to help with rebases, fixes, or testing to get this ready. Thanks

package.json
```diff
@@ -16,8 +16,9 @@
     "test:unit": "pnpm test:jest && pnpm test:vitest",
     "test:jest": "cross-env NODE_OPTIONS=--experimental-vm-modules jest",
     "test:vitest": "vitest --watch=false",
-    "test:playwright": "pnpm test:build-next && playwright test",
-    "test:build-next": "node scripts/copy-next-plugin.ts && pnpm --filter=@fixtures/next-* clean-build",
+    "test:playwright": "pnpm test:clean-next && pnpm test:build-next && playwright test",
+    "test:clean-next": "pnpm --filter=@fixtures/next-* run '/^clean.*/'",
```
askoufis, line 20: I'm a bit confused about `test:clean-next` because it looks like you've removed the `clean` scripts from the next fixtures. Should they have been removed?

fixtures/next-16-app-pages-router/src/styles/fonts/Inter-Regular.woff2
askoufis, line 1: Did a bit of research into the licensing of Inter. It seems we'd either need to commit a separate license file alongside the font, or commit a font file containing the license in its metadata. See https://openfontlicense.org/how-to-use-ofl-fonts/.

Do you know if this version embeds the license in its metadata? If not then the license should be committed alongside it, or alternatively a font with a more permissive license could be used.

fixtures/next-16-app-pages-router/next.config.ts
```diff
@@ -0,0 +1,13 @@
+import { NextConfig } from 'next';
+import { createVanillaExtractPlugin } from './next-plugin/dist/vanilla-extract-next-plugin.cjs.js';
```
askoufis, line 2: Should this import be using the package specifier rather than relative path?

```suggestion
import { createVanillaExtractPlugin } from '@vanilla-extract/next-plugin';
```

packages/turbopack-plugin/src/next-font/plugin.ts
```diff
@@ -0,0 +1,16 @@
+import { transformNextFont } from './transform';
+import type { Plugin } from 'vite';
+
+export function createNextFontVePlugin(): Plugin {
+  return {
+    name: 've-next-font-stub',
+    enforce: 'pre',
+    async transform(code: string, id: string) {
+      if (!/\.(?:[cm]?[jt]s|[jt]sx)$/.test(id) || !code.includes('next/font')) {
```
askoufis: Does this regex match the same files as `/\.(m|c)?(js|ts)x?$/` (used in `inject.js`)?

packages/turbopack-plugin/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "@swc/core": "^1.13.5",
+    "@vanilla-extract/compiler": "workspace:^",
+    "@vanilla-extract/integration": "workspace:^"
+  },
+  "devDependencies": {
+    "next": "12.3.4",
+    "vite": "^5.0.0 || ^6.0.0 || ^7.0.0"
+  },
+  "peerDependencies": {
+    "next": ">=12.1.7"
```
askoufis, line 33: Given the docs say only `next>=16` supports the turbopack plugin, should this version range reflect that too, or is this purely so package managers don't complain when a user that doesn't want to use turbopack installs `@vanilla-extract/next-plugin`?

.github/workflows/deploy-site.yml
```diff
@@ -1,6 +1,9 @@
 name: Deploy site
 
-on: push
+on:
+  push:
+    branches:
+      - master
```
askoufis: Any particular reason to change this? We deploy test sites for all branches to check docs changes.

.github/workflows/validate.yml
```diff
@@ -1,6 +1,10 @@
 name: Validate
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
```
askoufis: Same here - any particular reason to change this?

trekinbami:
It's been almost six months. I think it's safe to say this is not going to get merged. It might be a good idea to see if we can turn this PR into a plugin, or "just" fork?

malechite:
Maybe @mattcompiles and @askoufis should promote @RJWadley or some other contributors to be Maintainers of Vanilla Extract if they don't have the time

askoufis:
I started reviewing this over the weekend. It's a large PR so it will take some time.

zecka:
This might be an edge case, but I am hitting a failure when `outputFileTracingRoot` is set outside of the Next.js project root.

For example:

```ts
//next.config.ts
export const config: NextConfig = {
  outputFileTracingRoot: path.join(__dirname, '../../'),
};
```

With this configuration, Turbopack log the following error:
```
FATAL: An unexpected Turbopack error occurred. A panic log has been written to /var/folders/0t/x_bh86jn0cv7mvtwwdxp_plr0000gp/T/next-panic-6b92f89cf755c87a1be69f2c01527830.log.
```

Here is the complete log
```log
---------------------------
Resource path "nextjs/app/page.css.ts" need to be on project filesystem "apps/nextjs"

Debug info:
- Execution of hmr_update_with_issues_operation failed
- Execution of PlainIssue::from_issue failed
- Execution of PlainSource::from_source failed
- Execution of <WebpackLoadersProcessedAsset as Asset>::content failed
- Execution of WebpackLoadersProcessedAsset::process failed
- Resource path "nextjs/app/page.css.ts" need to be on project filesystem "apps/nextjs"
---------------------------
Resource path "nextjs/app/page.css.ts" need to be on project filesystem "apps/nextjs"
```
Here is a minimal reproduction on the try/outputFileTracingRoot branch:
https://github.com/zeckaissue/vanilla-extract-turbopack-test-yarn/tree/try/outputFileTracingRoot

Notes:
- the issue doesn't appear [with css module](https://github.com/zeckaissue/vanilla-extract-turbopack-test-yarn/commit/75f9e35898d92ef7b8b5c9c0486f496e98032a45)
- _(My goal with outputFileTracingRoot is to work locally with a UI library linked via yarn link. It is possible that this workflow is not supported and that I need to adjust how I work locally, but I wanted to report the behavior in case this is unintended.)_

kelseyway:
+1 on this, would be a massive win to have this in. currently blocked by this, and my company would love to use vanilla extract out the box with our turbopack next app

kelseyway (approved):

RJWadley:
> Here is a minimal reproduction on the try/outputFileTracingRoot branch:
[zeckaissue/vanilla-extract-turbopack-test-yarn@try/outputFileTracingRoot](https://github.com/zeckaissue/vanilla-extract-turbopack-test-yarn/tree/try/outputFileTracingRoot?rgh-link-date=2026-02-23T23%3A42%3A00Z)

@zecka I wasn't able to reproduce the issue on this repository when I cloned it. I would presume the issue is caused by the way things are linked on your filesystem.

Just guessing here, but one potential issue [that I have seen before](https://github.com/vercel/next.js/issues/84342) is that you may need to configure `turbopack.root` as turbopack is a bit more picky with how it resolves modules than webpack is. Try setting it to match your outputFileTracingRoot configuration.

Since you're using `yarn link`, you may also need to set the root higher (potentially as high as the filesystem root) to get turbopack to resolve the modules.


[here's the relevant bit of documentation for turbopack.root
](https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack#root-directory)

zecka:
> > Here is a minimal reproduction on the try/outputFileTracingRoot branch:
> > [zeckaissue/vanilla-extract-turbopack-test-yarn@try/outputFileTracingRoot](https://github.com/zeckaissue/vanilla-extract-turbopack-test-yarn/tree/try/outputFileTracingRoot?rgh-link-date=2026-02-23T23%3A42%3A00Z)
> 
> @zecka I wasn't able to reproduce the issue on this repository when I cloned it. I would presume the issue is caused by the way things are linked on your filesystem.

You’re right, it’s probably related to my local system configuration.
I’m not able to reproduce the issue on Codespaces, for example.
However, I can still reproduce it on my Mac, even with a fresh clone of the repository.
I’ll drop it for now, this is likely too much of an edge case to consider.

packages/turbopack-plugin/src/next-font/transform.ts
```diff
@@ -0,0 +1,274 @@
...
+  if (isLocal) {
+    parts.push(formatFontFamily(`${baseName} Fallback`));
+    if (fallbacks) parts.push(...fallbacks);
+  } else {
+    if (fallbacks && fallbacks.length > 0) {
+      parts.push(...fallbacks);
+    } else {
+      parts.push(formatFontFamily(`${baseName} Fallback`));
+    }
+  }
```
askoufis: I think this can be simplified a bit:

```suggestion
  const hasUserFallbacks = fallbacks && fallbacks.length > 0;
  
  if (isLocal || !hasUserFallbacks) {
    parts.push(formatFontFamily(`${baseName} Fallback`));
  }

  if (hasUserFallbacks) {
    parts.push(...fallbacks);
  }
```

packages/turbopack-plugin/src/next-font/transform.ts
```diff
@@ -0,0 +1,274 @@
...
+
+  return parts.join(', ');
+}
+
+function parseWeight(weight: any): number | undefined {
```
askoufis:
```suggestion
function parseWeight(weight: unknown): number | undefined {
```

packages/turbopack-plugin/src/next-font/transform.ts
```diff
@@ -0,0 +1,274 @@
...
+    let weight = parseWeight(opts.weight);
+    let style = parseStyle(opts.style, isGoogle);
+
+    // Local fonts only stub weight/style if src is a simple string
+    if (!isGoogle && typeof opts.src !== 'string') {
+      weight = undefined;
+      style = undefined;
+    }
+
+    decl.init = await createStubNode(family, weight, style);
```
askoufis: I'd prefer this function to return a new `init` rather than modify the function input `decl`.

.changeset/fresh-wombats-stick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@vanilla-extract/turbopack-plugin': minor
+'@vanilla-extract/next-plugin': minor
+---
+
+turbopack support for next.js >= 16.0.0
```
askoufis:
```suggestion
Add turbopack support for next.js >= 16.0.0
```

packages/compiler/src/compiler.ts
```diff
@@ -258,7 +259,17 @@ export interface CreateCompilerOptions {
...
+    filePath: string,
+    css: string,
+  ) => string | Promise<string>;
+  /**
+   * When true, generates one CSS import per rule instead of one per file.
+   * This can help bundlers like Turbopack deduplicate shared CSS more effectively.
+   *
+   * @default false
+   */
+  splitCssPerRule?: boolean;
```
askoufis: I'm a bit confused about this. If you're already sharing the same VE style in multiple places, isn't there no need to deduplicate anything?

packages/next-plugin/src/index.ts
```diff
@@ -12,10 +12,22 @@ import type {
...
+type PluginOptions = ConstructorParameters<typeof VanillaExtractPlugin>[0] & {
+  turbopackGlob?: string[];
+  /**
+   * controls whether we attempt to configure turbopack
+   * auto: enable only when Next >= 16.0.0
+   * on: force enable regardless of detected Next version
+   * off: never configure turbopack, webpack only
+   */
+  turbopackMode?: 'auto' | 'on' | 'off';
+};
```
askoufis, line 32: I'd prefer these options to be nested under an `unstable_turbopack` property for 2 reasons:
- I'm not confident that this integration won't require breaking changes in the near future
- Grouping `turbopack`-related options together seems a bit neater than prefixing everything with `turbopack`

packages/css/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "exports": {
     "./package.json": "./package.json",
+    "./vanilla.virtual.css?*": "./vanilla.virtual.css?*",
```
askoufis, line 15: Is the `?*` necessary here? AFAIK query parameters should be preserved during module resolution, and they don't need to be specified in export conditions.

```suggestion
    "./vanilla.virtual.css": "./vanilla.virtual.css",
```

packages/turbopack-plugin/src/index.ts
```diff
@@ -0,0 +1,219 @@
...
+
+  sharedCompiler = createCompiler({
+    root: loaderContext.rootContext,
+    identifiers,
+    enableFileWatcher: false,
+    splitCssPerRule: true,
+    cssImportSpecifier: async (_filePath, css) => {
+      return `@vanilla-extract/css/vanilla.virtual.css?ve-css=${encodeURIComponent(
+        await serializeCss(css),
+      )}`;
```
askoufis, line 74: Can the `vanilla.virtual.css` file live inside the turbopack plugin package instead of the css package? I'd rather it be associated with the plugin it's relevant to, rather than being included in the css package.

site/docs/integrations/next.md
```diff
@@ -60,13 +65,54 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
...
 const withVanillaExtract = createVanillaExtractPlugin({
   identifiers: ({ hash }) => `prefix_${hash}`
 });
 ```
 
 Each integration will set a default value based on the configuration options passed to the bundler.
 
+### turbopackMode
+
+You can control Turbopack autoconfiguration using `turbopackMode`:
```
askoufis: I'd appreciate a warning of some kind here that informs consumers that this API is experimental and that it doesn't handle all use cases.
```suggestion
### turbopackMode

> ⚠️&nbsp;&nbsp;Turbopack support is experimental. Its API is unstable and may undergo breaking changes in non-major versions. Additionally, it may not handle all features supported by Next.js.

You can control Turbopack autoconfiguration using `turbopackMode`:
```

package.json
RJWadley: the cleaning is all still in there, I just removed the `clean-build` shortcut that calls clean + build.

this is just because we're testing webpack + turbopack and I wanted them to build in parallel, and it seemed like the easiest way at the time.

I would rather remove the cleaning entirely, as it shouldn't be needed, but thought that would be best for a future PR if at all.

fixtures/next-16-app-pages-router/src/styles/fonts/Inter-Regular.woff2
RJWadley: ~~the exact font doesn't matter, I just needed to load *something* and grabbed the first font I thought of. I'll replace it with a font in the public domain.~~

nvm, the license is embedded in the metadata (it's the [SIL Open Font License 1.1](http://scripts.sil.org/OFL)) so we're good.

fixtures/next-16-app-pages-router/next.config.ts
RJWadley: that would be better IMO, but neither of the existing next fixtures do, and we'll experience type errors if we try. see `scripts/copy-next-plugin.ts` for a bit more context.

this is another thing I'd like to clean up soon, but opted to just match the status quo for the first revision. happy to take another look if you want it in this PR.

packages/turbopack-plugin/src/next-font/plugin.ts
RJWadley: this regex matches any javascript/typescript file, just to skip processing of non-js files like JSON.

full disclosure: the next-font transform itself was AI-generated (you can probably tell). most of this PR was hand-crafted by yours truly, but not that.

I'll swap it out for the far superior version you referenced.

edit: done.

packages/turbopack-plugin/package.json
RJWadley: this is purely so that package managers don't complain. users don't install turbopack-plugin directly, it's installed by the next plugin. so any users on old next version would see peer errors.

Turbopack will only configure itself if next is >= 16. The user could still attempt to manually configure turbopack on next 15, but it would always crash if they did. It would be trivial to warn in this case, which I think is a good idea.

edit: added

.github/workflows/deploy-site.yml
RJWadley: this was temporary, will revert (I forgot I had done this)
edit: done

.github/workflows/validate.yml
RJWadley: this was temporary, will revert (I forgot I had done this)
edit: done

packages/compiler/src/compiler.ts
RJWadley: I can't confirm how webpack or any other integration works, but before adding this we were seeing massive amounts of duplication across reused files. For example, every file that used our color library would re-bundle the entire color library (and it would be included hundreds of times)

I added a test that catches some common duplication issues, and splitting each rule out per import was the easiest way to deduplicate them all.

[turbopack preserves the import order when bundling the CSS](https://nextjs.org/docs/app/api-reference/turbopack#css-module-ordering)

packages/next-plugin/src/index.ts
RJWadley: this is actually great idea, I might also do this to the compiler option I added to reset the cache (since it's only there to work around a bug in the compiler)

edit: done. also renamed the two experimental compiler apis.

packages/css/package.json
RJWadley: I did not originally include it, but iirc I added it because turbopack does require it. will double check.

packages/turbopack-plugin/src/index.ts
RJWadley: turbopack must be able to resolve the import from the source file itself - so if you're compiling styles across packages in a monorepo (like how, for example, our test fixtures do) turbopack would not be able to resolve `@vanilla-extract/turbopack-plugin`. Turbopack searches for that module from the importing file, so the only package that is guaranteed to exist is `@vanilla-extract/css`. Turbopack refused to resolve relative imports here, and it doesn't support importing the absolute path.

Originally I used data imports here but those are not deduplicated by turbopack.

Resolution and compilation across package boundaries is something I'd like to revisit at some point, as every integration seems to handle this slightly differently.

packages/next-plugin/src/index.ts
RJWadley: do you also want turbopack support to disable to "off" instead of "auto"? auto mode configures turbopack if we're on next >=16

packages/css/package.json
RJWadley: confirmed. turbopack does not strip query strings before matching against the exports field. without the wildcard, every CSS import will fail.

packages/turbopack-plugin/src/next-font/transform.ts
RJWadley: done. cleaned up a few other things in this file too.

RJWadley:
@askoufis thanks for the review! addressed feedback and replied to comments, I will leave resolving them up to you at your satisfaction.

packages/next-plugin/src/index.ts
askoufis: It should default to `off`. I'd much prefer new features like this to be opt-in, especially if they're experimental.

fixtures/next-16-app-pages-router/next.config.ts
askoufis: All good, happy for cleanup to be separate to this PR.

package.json
askoufis: Oh right, there are still `clean:dev` and `clean:prod` scripts, which will be called by `test:clean-next`. Cool.

packages/turbopack-plugin/src/index.ts
askoufis: Ah, that makes sense. Every bundler has its quirks, and all the itegrations have evolved rather organically and at different points in time, so it'd be great if they could eventually align with each other to some extent.

test-helpers/src/startFixture/index.ts
```diff
@@ -69,7 +69,11 @@ export async function startFixture(
     });
   }
 
-  if (type === 'next-pages-router' || type === 'next-app-router') {
+  if (
+    type === 'next-12-pages-router' ||
+    type === 'next-13-app-router' ||
+    type === 'next-16-app-pages-router'
```
askoufis, line 75: Probably worth addressing in a separate PR, but this change alone isn't enough to run the next 16 fixture.

`next.config.js` is hardcoded as the config file because next 12 and 13 don't support TypeScript config, but next 16 does.

But even when I hardcode `.ts` config, the next 16 fixture starts but navigating to the sprinkles page results in `Styles were unable to be assigned to a file` errors.

Another thing I noticed when running the built next 16 app was that navigating to a fixture page then going back and navigating to another page accumulated stylesheets from both pages. You can keep doing this for all the pages in the app, resulting in some broken styles.

The most obvious way to see this to go to the duplication test page, then back, and then the sprinkles back and then back again. You end up with `body` styles from both pages.

RJWadley: I actually had lots of issues trying to use the start-fixture CLI with next. Since I don't think it's really representative of the way folks run next in reality, I didn't invest much time in trying to fix it and opted instead to use the official tooling. See `tests/servers.ts` for more details. I actually quite liked doing it this way, and am considering reworking the fixture system in a follow-up to make testing VE more consistent across all the supported bundlers.

With respect to the global styles "leaking" across pages, this is currently expected behavior per the [next docs](https://nextjs.org/docs/app/getting-started/css#global-styles:~:text=Good%20to%20know%3A%20Global,needed) and the [react docs](https://react.dev/reference/react-dom/components/style#:~:text=React%20may%20leave%20the%20style%20in%20the%20DOM%20even%20after%20the%20component%20that%20rendered%20it%20has%20been%20unmounted). React makes no attempt to clean up stylesheets when components that use them unmount. I have invested a lot of time and energy into trying to fix that in upstream react, but it is actually quite complex to do universally with good performance. I imagine it won't be fixed until some youtuber makes a clickbait video with a title like "I made react's styles 40 billion times faster" at which point the facebook geniuses will take notice and write some really good code to fix it for real

I wouldn't mind attempting to fix this for vanilla extract specifically, but it definitely belongs in a future PR (if at all).

packages/next-plugin/src/index.ts
RJWadley: adjusted

site/docs/integrations/next.md
```diff
@@ -60,13 +65,62 @@ Different formatting of identifiers (e.g. class names, keyframes, CSS Vars, etc)
...
+
+Turbopack-related options are grouped under the `unstable_turbopack` key.
+
+#### unstable_turbopack.mode
+
+You can control Turbopack autoconfiguration using `mode`:
+
+- `auto` (default): enable Turbopack config only when Next >= 16.0.0
+- `on`: force-enable Turbopack config
+- `off`: never configure Turbopack (Webpack-only)
```
askoufis: Given the default value change, this should be updated:

```suggestion
- `off` (default): never configure Turbopack (i.e. use Webpack)
- `auto`: enable Turbopack config only when Next >= 16.0.0
- `on`: force-enable Turbopack config
```

test-helpers/src/startFixture/index.ts
askoufis: Fair enough. I think it makes more sense to handle Next.js separately from the rest of the bundler integrations, so happy for more effort to be put into that in the future.

I do want to make the fixture CLI much better for all bundlers at some point.

Re: react - I'm not sure how much framework-specific stuff we want to put into Vanilla Extract, if any. If there's a need for it then maybe it's worth working on in the future.

askoufis:
Turbopack support has been released in [v2.5.0 of the next plugin](https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fnext-plugin%402.5.0).

plexey:
Thank you so much for taking the time to add Turbopack support @askoufis @RJWadley & others.

Among other benefits, this work has significantly improved our build times in CI from 6 minutes down to 2 minutes, which is a huge win.

Your hard work doesn't go unnoticed and is super appreciated ❤️

nandahibatullah:
Echoing @plexey's appreciation here @RJWadley and @askoufis + everyone else involved.

Thank you for all the time you put into this. I happen to work with @plexey and we are enjoying the improved build times - so huge for us. Mega effort 🔥

PR #1643, opened 2025-10-29
Add container type `scroll-state` to type definitions

maurer2:
Hello, there is a new container type for scroll state queries that is currently missing in the custom type definitions. More info: https://developer.mozilla.org/en-US/docs/Web/CSS/container-type

Cheers


askoufis:
@maurer2 Thanks for raising the PR.

Recent versions of `csstype` actually support `scroll-state`, so I opted to remove this custom type instead in #1666.

PR #1648, opened 2025-11-09
fix: use base64url encoding to prevent esbuild ?source= corruption

sallustfire:
Base64-encoded CSS data in virtual file paths as the source query parameter can contain `/` and `//` characters, which are susceptible to path normalization in bundler plugin systems. While the source parameter should be opaque it is not URL safe and esbuild configurations can collapse `//` → `/`, which corrupts the base64 data preventing compressed source data from being read.

Switching to base64url encoding replaces `/` with `_` and `+` with `-`, making the encoded data safe for use in file paths and resilient to this class of error.


sallustfire:
We encountered this in production using @vanilla-extract/esbuild-plugin with a CSS file where gzip+base64 produced 
a `//` sequence:

- onResolve receives: `...?source=#H4sI...4//feL...` (509 chars)
- onLoad receives: `...?source=#H4sI...4/feL...` (508 chars) ← corrupted
- Result: `Error: incorrect header check` from gunzip

.changeset/honest-lamps-complain.md
```diff
@@ -0,0 +1,5 @@
+---
+'@vanilla-extract/integration': patch
+---
+
+base64url encode generated CSS source to avoid errors in virtual file paths
```
askoufis:
```suggestion
`serializeCss|deserializeCss`: Use `base64url` encoding instead of `base64` to avoid errors in virtual file paths
```

askoufis:
@sallustfire Thanks for the PR!

askoufis:
@sallustfire This fix is now available in the latest version of all the bundler plugins.

PR #1650, opened 2025-11-18
Merge Vite server config with existing settings

alechoey:
Trying to launch a vite server with `middlewareMode` doesn't get propagated to the compiler vite server.

Without middlewareMode, the vite server directly handles `SIGTERM` and immediately exits the process. This breaks the parent server behavior (such as Fastify graceful shutdown)

### DIFF
--- a/packages/compiler/src/compiler.ts
+++ b/packages/compiler/src/compiler.ts
@@ -109,6 +109,7 @@ const createViteServer = async ({
     configFile: false,
     root,
     server: {
+      ...(viteConfig.server ?? {}),
       hmr: false,
       watch: enableFileWatcher ? undefined : null,
     },


askoufis:
@alechoey Thanks for the PR! Would it be possible for you to make a minimal reproduction of your issue? It would help me understand the problem better.

As it stands, I'm hesitant to blindly spread the entire `server` config into the compiler's vite server. This isn't because some options may cause problems (they may, I just haven't validated this), but primarily because future changes to vite's config options could cause incompatibilities. I'd rather be explicit about which configuration options are being forwarded to the compiler, i.e. only forward `server.middlewareMode`.

alechoey:
@askoufis Thanks for the feedback! I've created a minimal reproduction:

**Repro: https://github.com/alechoey/vanilla-extract-middleware-repro**

To run:
```bash
git clone https://github.com/alechoey/vanilla-extract-middleware-repro
cd vanilla-extract-middleware-repro
npm install
./test-sigterm.sh
```

The test clearly shows the issue with timestamps:

**Control case (NO vanilla-extract)** - graceful shutdown works:
```
[19:14:37.186] Received SIGTERM, starting graceful shutdown...
[19:14:37.186] Draining in-flight requests...
[19:14:37.188] Vite server closed
[19:14:46.196] Responding to /slow request     ← 9 seconds later, request completes!
[19:14:46.198] Graceful shutdown complete
```

**With vanilla-extract** - process exits immediately:
```
[19:14:55.470] Received SIGTERM, starting graceful shutdown...
[19:14:55.471] Draining in-flight requests...
                                              ← Process exits immediately, request never completes
```

At a minimum, we need to forward `middlewareMode` to the internal compiler's Vite server:

```typescript
// In packages/compiler/src/compiler.ts
server: {
  hmr: false,
  watch: enableFileWatcher ? viteConfig.server?.watch : null,
  middlewareMode: viteConfig.server?.middlewareMode,  // Add this line
},
```

PR #1654, opened 2025-11-24
rollup: Add `unstable_injectFilescopes` flag

askoufis:
Some packages choose to ship transpiled VE modules (e.g. `.css.cjs`, `css.mjs`) rather than statically extracted CSS + JS variables containing classnames. In such cases, it's often beneficial to inject filescopes into Vanilla Extract modules as there's no guarantee that the consumer's bundling/test configuration will do this.

As an example, [moving to `tsdown` in `capsize`](https://github.com/seek-oss/capsize/pull/241) resulted in some consumers encountering `Styles were unable to be assigned to a file` errors as VE modules without filescopes were loaded. The new `unstable_injectFilescopes` flag injects filescopes where necessary, fixing this error for `capsize` consumers.

### DIFF
--- a/.changeset/two-chefs-hide.md
+++ b/.changeset/two-chefs-hide.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Update `csstype` dependency to `^3.2.3`
--- a/packages/css/package.json
+++ b/packages/css/package.json
@@ -120,7 +120,7 @@
     "@vanilla-extract/private": "workspace:^",
     "css-what": "^6.1.0",
     "cssesc": "^3.0.0",
-    "csstype": "^3.0.7",
+    "csstype": "^3.2.3",
     "dedent": "^1.5.3",
     "deep-object-diff": "^1.1.9",
     "deepmerge": "^4.2.2",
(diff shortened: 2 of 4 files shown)


site/docs/integrations/rollup.md
askoufis, line 1: Edited these docs a bit to ensure once sentence per line.

packages/rollup-plugin/test/__snapshots__/rollup-plugin.test.ts.snap
```diff
@@ -554,6 +554,303 @@ export { altTheme, responsiveTheme, theme, vars };
...
+
+export { button, container, opacity };
+",
+  ],
+  [
+    "src/themes.css.js",
+    "import { setFileScope, endFileScope } from '@vanilla-extract/css/fileScope';
+import { style, createGlobalTheme, createTheme, layer, assignVars } from '@vanilla-extract/css';
+
+setFileScope("src/themes.css.ts", "@fixtures/themed");
```
askoufis, line 742: This is the main effect of the filescope injection (as well as the `endFileScope` call at the end of the file).

PR #1657, opened 2025-11-25
Enhance RecipeStyleRule to support nested styles

JoseGalRe:
This pull request updates the `RecipeStyleRule` type definition in `packages/recipes/src/types.ts` to support more flexible and nested style rule configurations.

This change allows style rules to be defined as objects with arbitrary keys, each mapping to another `RecipeStyleRule`, enabling more complex and deeply nested style structures.

Type definition improvements:

* Expanded the `RecipeStyleRule` type to allow objects with string keys mapping to additional `RecipeStyleRule` values, supporting recursive/nested style rule definitions.

### DIFF
--- a/.changeset/enhance-recipe-style-rule-to-support-nested-styles.md
+++ b/.changeset/enhance-recipe-style-rule-to-support-nested-styles.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/recipes': minor
+---
+
+Enhance RecipeStyleRule to support nested styles


askoufis:
Modifying the types alone is not enough to implement this feature.

> to support more flexible and nested style rule configurations

I'm skeptical of the utility of actually doing this. I'd much prefer this be raised as a discussion with use cases/example code for a potential API before even attempting to implement anything.

Going to close this PR for now pending further discussion.

JoseGalRe:
@askoufis 🤔 Interesting, it's strange because it works for me, The only problem I have is the typing, this doesn't allow nesting.

By modifying the type, TypeScript no longer reports any issues.

image

Here is a repository with minimal reproducible (need pnpm). https://github.com/JoseGalRe/vanilla-extract-recipes-css-nesting-types

askoufis:
@JoseGalRe This is exactly why a discussion before trying to implement anything would have been useful.

Your usage happens to result in correct CSS, but only because our serializer happens to output nested CSS in this situation. However, this is more of an accidental implementation detail than anything else, and not an intentional feature, which is why the current types stop you from doing this.

The regular `style` API, and by extension `recipe`, doesn't support nested styles, however you can use the `selectors` API to generate a subset of nested CSS, specifically the selector must target the current style ([docs](https://vanilla-extract.style/documentation/styling/#complex-selectors)). This is a scoped style.

For example:

```ts
export const foo = recipe({
  base: {
    display: 'flex',
    selectors: {
      'h1 &': {
        backgroundColor: '#212121',
      },
      'p &': {
        textDecoration: 'underline',
      },
      '.nesting-class h1 &': {
        color: 'white',
      },
      '.nesting-class p &': {
        fontSize: '40px',
        backgroundColor: 'greenyellow',
      },
    },
  },
});
```

This generates the following CSS:

```css
.styles_foo__spi1fre {
  display: flex;
}
h1 .styles_foo__spi1fre {
  background-color: #212121;
}
p .styles_foo__spi1fre {
  text-decoration: underline;
}
.nesting-class h1 .styles_foo__spi1fre {
  color: white;
}
.nesting-class p .styles_foo__spi1fre {
  font-size: 40px;
  background-color: greenyellow;
}
```

This isn't quite the same as what you're trying to do. In your case, you'll want to use `globalStyle` which gives you the freedom to write unscoped styles using any selector you want. You can tie this to the recipe base class like this:

```ts
export const foo = recipe({
  base: {
    display: 'flex',
  },
});

globalStyle(`${foo.classNames.base} h1`, {
  backgroundColor: '#212121',
});
globalStyle(`${foo.classNames.base} p`, {
  textDecoration: 'underline',
});
globalStyle(`${foo.classNames.base} .nesting-class h1`, {
  color: 'white',
});
globalStyle(`${foo.classNames.base} .nesting-class p`, {
  fontSize: '40px',
  backgroundColor: 'greenyellow',
});
```

This would generate the following CSS, which is equivalent to the nested CSS you're trying to generate with `recipe`:

```
.styles_foo__spi1fre {
  display: flex;
}
.styles_foo__spi1fre h1 {
  background-color: #212121;
}
.styles_foo__spi1fre p {
  text-decoration: underline;
}
.styles_foo__spi1fre .nesting-class h1 {
  color: white;
}
.styles_foo__spi1fre .nesting-class p {
  font-size: 40px;
  background-color: greenyellow;
}
```

A CSS preprocessor may then be able to convert this into actual nested CSS if you want.

Obviously having to call `globalStyle` a bunch of times and template in the styles you want is more verbose, but it sends a clear signal that the styles being generated are global styles, which means you're opting out of the benefits of scoped styles.

JoseGalRe:
@askoufis oh nice, it's clear to me, thank you for the explanation.

PR #1669, opened 2025-12-11
Consolidate release workflows for OIDC publishing

askoufis:
- NPM packages have been configured for OIDC publishing
- You can only specify 1 workflow file for trusted publishing, so the snapshot and release workflows have been consolidated
- OIDC publishing also needs at least `npm@11.5.1`, so I've updated the node version to LTS 24, which installs `npm@11.6.1` at the moment


askoufis:
Followed up by #1670 which actually made things work.

PR #1718, opened 2026-04-09
fix(vite-plugin): normalize absolute Vite ids on Windows

### LINKED ISSUE #1717: No CSS for file when workspace .css.ts modules are resolved from Vite ids shaped like @id/C:/...
### Describe the bug

Hi, we're seeing a dev-mode issue with `@vanilla-extract/vite-plugin` when a workspace `*.css.ts` module is resolved through a Vite id shaped like `@id/C:/...`.

We first observed this naturally on Windows in a pnpm monorepo. We then built a minimal repro and found that the same underlying failure can also be triggered on macOS/Linux by simulating that same Vite id shape. So this no longer looks Windows-only; Windows just seems to be where that id form shows up naturally in real projects.

Versions:
- `@vanilla-extract/vite-plugin`: `5.2.1`
- `@vanilla-extract/compiler`: `0.6.0`
- `vite`: reproduced in minimal repro with `6.3.3`, originally observed in a real app with `8.0.8`
- `node`: reproduced locally with Node `20.17.8`; originally observed with Node `24.14.1`

Context:
- pnpm monorepo
- Vite app imports `*.css.ts` from a sibling workspace package
- happens in dev mode
- production build works

What happens:
`@vanilla-extract/vite-plugin` eventually calls `getCssForFile(...)` with a file path that does not match the compiler cache key and throws:

```txt
Error: No CSS for file: .../@id/C:/.../block.css.ts
    at Object.getCssForFile (...@vanilla-extract/compiler...)
    at ResolveIdContext.resolveId (...@vanilla-extract/vite-plugin...)
```

Example stack:

```txt
at Object.getCssForFile (.../node_modules/@vanilla-extract/compiler/dist/vanilla-extract-compiler.cjs.dev.js:415:15)
at ResolveIdContext.resolveId (.../node_modules/@vanilla-extract/vite-plugin/dist/vanilla-extract-vite-plugin.cjs.dev.js:256:81)
at EnvironmentPluginContainer.resolveId (.../node_modules/vite/dist/node/...)
at async EnvironmentModuleGraph._resolveUrl (...)
at async EnvironmentModuleGraph.getModuleByUrl (...)
```

Why we think `@id/C:/...` is the key part:
- In the real app, the failure appeared when Vite/TanStack-style dev resolution produced ids shaped like `@id/C:/...`
- In the minimal repro, we can trigger the same failure on macOS by simulating:

```txt
/@id/C:/.../block.css.ts.vanilla.css
```

Expected:
`@vanilla-extract/vite-plugin` should normalize or otherwise handle Vite ids shaped like `@id/C:/...` consistently in dev mode, so they map to the same compiler cache entry as the original `*.css.ts` file.

Actual:
The plugin looks up CSS using a mismatched path and throws `No CSS for file`.

Current workaround:
We can work around this by adding a small Vite `resolveId` normalization step before vanilla-extract sees the id, rewriting values shaped like:

```txt
@id/C:/foo/bar.css.ts
```

back to:

```txt
C:/foo/bar.css.ts
```

That avoids the mismatch.

See the minimal reproduction repo as well.

### Reproduction

https://github.com/MagneH/vanilla-extract-vite-windows-bug-repro

### System Info

```shell
System info:
- Reproduced locally on macOS 26.3.1 (build 25D2128), arm64
- Node.js: 24.14.0
- pnpm: 10.33.0
- vite: 6.3.3 (minimal repro)
- @vanilla-extract/vite-plugin: 5.2.1
- @vanilla-extract/compiler: 0.6.0

Also originally observed in a real app on:
- Windows
- Node.js: 24.14.1
- vite: 8.0.8
```

### Used Package Manager

pnpm

### Logs

```shell
at Object.getCssForFile (.../node_modules/@vanilla-extract/compiler/dist/vanilla-extract-compiler.cjs.dev.js:415:15)
at ResolveIdContext.resolveId (.../node_modules/@vanilla-extract/vite-plugin/dist/vanilla-extract-vite-plugin.cjs.dev.js:256:81)
at EnvironmentPluginContainer.resolveId (.../node_modules/vite/dist/node/...)
at async EnvironmentModuleGraph._resolveUrl (...)
at async EnvironmentModuleGraph.getModuleByUrl (...)
```

### Validations

MagneH:
Fixes #1717

## Summary

Fix a Windows dev-mode issue in `@vanilla-extract/vite-plugin` where Vite can resolve workspace virtual CSS modules as absolute `@id/...` ids, for example:

- `@id/C:/...`
- `/@id/C:/...`

Previously these ids could be treated as root-relative paths and joined with `config.root`, producing an invalid path and eventually causing `Error: No CSS for file: ...`.

This change:

- normalizes absolute Vite `@id/...` ids before resolving virtual CSS files
- preserves existing behavior for monorepo absolute paths outside `config.root`
- preserves existing SSR-style root-relative paths like `/app/styles.css.ts`
- adds tests covering the Windows `@id/C:/...` case and the existing path behaviors

## Testing

- `pnpm build`
- `pnpm test:unit -- packages/vite-plugin/src/ids.test.ts`

### DIFF
--- a/.changeset/chilly-bees-hug.md
+++ b/.changeset/chilly-bees-hug.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Normalize absolute Windows `@id/...` module ids before resolving virtual CSS files in Vite dev mode.
--- a/.changeset/chilly-bees-hug.md
+++ b/.changeset/chilly-bees-hug.md
@@ -2,4 +2,6 @@
 '@vanilla-extract/vite-plugin': patch
 ---
 
-Normalize absolute Windows `@id/...` module ids before resolving virtual CSS files in Vite dev mode.
+Fix `No CSS for file` error in Vite dev mode when virtual CSS modules are resolved through Vite's `@id/` id wrapper
+
+This was most commonly hit with workspace `.css.ts` imports in pnpm monorepos on Windows, where Vite produces ids shaped like `@id/C:/...`, but it can also occur on macOS/Linux. The plugin now unwraps these ids before resolving the virtual CSS file so they map to the correct compiler cache entry.
--- a/packages/vite-plugin/src/ids.test.ts
+++ b/packages/vite-plugin/src/ids.test.ts
@@ -1,0 +1,52 @@
+import { describe, expect, it } from 'vitest';
+
+import { getAbsoluteId } from './ids';
+
+describe('getAbsoluteId', () => {
+  it('unwraps Windows absolute Vite ids', () => {
+    expect(
+      getAbsoluteId({
+        root: 'C:/Users/nit/projects/nettflater/apps/spkno',
+        filePath:
+          '@id/C:/Users/nit/projects/nettflater/packages/shared/src/PortableText/block/block.css.ts.vanilla.css',
+      }),
+    ).toBe(
+      'C:/Users/nit/projects/nettflater/packages/shared/src/PortableText/block/block.css.ts.vanilla.css',
+    );
+  });
+
+  it('unwraps Windows absolute Vite ids with a leading slash', () => {
+    expect(
+      getAbsoluteId({
+        root: 'C:/Users/nit/projects/nettflater/apps/spkno',
+        filePath:
+          '/@id/C:/Users/nit/projects/nettflater/packages/shared/src/PortableText/block/block.css.ts.vanilla.css',
+      }),
+    ).toBe(
+      'C:/Users/nit/projects/nettflater/packages/shared/src/PortableText/block/block.css.ts.vanilla.css',
+    );
+  });
+
+  it('keeps absolute monorepo paths outside the app root', () => {
+    expect(
+      getAbsoluteId({
+        root: '/Users/nit/projects/vanilla-extract/packages/vite-plugin',
+        filePath:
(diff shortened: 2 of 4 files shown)


MagneH:
Quick update: I refined the [repro](https://github.com/MagneH/vanilla-extract-vite-windows-bug-repro) so it now exercises `@vanilla-extract/vite-plugin` with an `@id`-wrapped absolute path on macOS, without relying on fake `C:/...` path semantics on POSIX.

Results:

- Published `@vanilla-extract/vite-plugin@5.2.1`: still fails
- Local fork with the normalization fix: succeeds

Published plugin result:

```txt
500 Internal Server Error
No CSS for file: .../@id/Users/.../block.css.ts
```

Fixed fork result:

```json
{
  "source": "/@id//Users/.../block.css.ts.vanilla.css",
  "resolvedId": "/Users/.../block.css.ts.vanilla.css",
  "loadedCss": true
}
```

So the repro now gives a clean before/after signal on macOS:
- current release reproduces the bug
- the forked fix resolves it

askoufis (approved):
Thanks for the PR. I made a few changes:
- Some refactors to the `getAbsoluteId` code for clarity
- Removed Windows UNC path support as this is a niche use case and complicates the regex a bit
- Anonymized the file paths used in the test cases

PR #1720, opened 2026-04-11
add scope at-rule support

cahnory:
Add support for @scope rules


cahnory:
I'm having some trouble running the e2e tests locally on my machine, and I'd really appreciate a bit of help if anyone has an idea what I might be missing. I'm on macOS, in case that makes a difference.

askoufis:
> I'm having some trouble running the e2e tests locally on my machine, and I'd really appreciate a bit of help if anyone has an idea what I might be missing. I'm on macOS, in case that makes a difference.

They can be a bit temperamental. Usually running a `pnpm build` before `pnpm playwright` should be all that's necessary. Happy to help out further on discord (or elsewhere) if you want.

cahnory:
> > I'm having some trouble running the e2e tests locally on my machine, and I'd really appreciate a bit of help if anyone has an idea what I might be missing. I'm on macOS, in case that makes a difference.
> 
> They can be a bit temperamental. Usually running a `pnpm build` before `pnpm playwright` should be all that's necessary. Happy to help out further on discord (or elsewhere) if you want.

I just tried running the e2e tests locally on `master` and got the same result, so it doesn’t seem to be caused by my changes (at least not all errors ^^).

The main issue I’m seeing locally is that the tests appear to be picking up development class names in a production context, for example `.themes_responsiveTheme__cdwe0ve` instead of `.cdwe0ve`.

What makes this harder to reason about is that CI is also failing, but with different errors:
https://github.com/cahnory/vanilla-extract/actions/runs/24511603102/job/71643966254

I really don't know what to do :/

cahnory:
@askoufis The e2e test issues were caused by parallelization. Since the tests write files, multiple workers were stepping on each other, and limiting the worker count to 1 fixed everything.

I was then able to generate the correct snapshots/screenshots, and the tests are now passing:
https://github.com/cahnory/vanilla-extract/actions/runs/24556048277

Thanks again for your help on Discord!

tomahl:
@cahnory glad to see this PR. 👍  Does it support scope limit selectors `@scope (scope root) to (scope limit)`?

cahnory:
> @cahnory glad to see this PR. 👍 Does it support scope limit selectors `@scope (scope root) to (scope limit)`?

Yes it does as the "query" part (what follows `@scope `) is not validated. The same treament as for selectors is applied (TL;DR: you can use classes generated by vanilla-extract and the `&` character).

packages/css/src/transformCss.ts
```diff
@@ -542,6 +548,58 @@ class Stylesheet {
...
+      const transformedScopeBounds: Record<string, string> = {};
+      this.currConditionalRuleset?.addConditionPrecedence(
+        parentConditions,
+        Object.keys(rules).map((bounds) => {
+          transformedScopeBounds[bounds] = `@scope ${this.transformSelector(
+            bounds.replace(
+              RegExp('(?<=^(?:[^"\']*|"[^"]*"|\'[^\']*\')*)&', 'g'),
+              root.selector,
+            ),
+          )}`;
```
askoufis, line 564: I'm a bit confused by this. What is this regex doing? Is there a reason why `this.transformSelector(selector.replace(RegExp('&', 'g'), root.selector))` wouldn't work here?

packages/sprinkles/src/index.ts
```diff
@@ -281,6 +286,14 @@ export function defineProperties(options: any): any {
             };
           }
 
+          if (condition['@scope']) {
```
askoufis, line 289: Usage of this should be added to the sprinkles fixture.

.changeset/thirty-suns-show.md
```diff
@@ -0,0 +1,18 @@
...
+
+```ts
+import { style } from '@vanilla-extact/css';
+export const styleWithScopeRule = style({
+    '@scope': {
+      '(body)': {
+        ':after': { content: '"Scoped to body"' },
+      },
+    },
+  });
```
askoufis: Indentation seems inconsistent here.
```suggestion
  '@scope': {
    '(body)': {
      ':after': { content: '"Scoped to body"' },
    },
  },
});
```

packages/css/src/transformCss.test.ts
```diff
@@ -1167,6 +1167,509 @@ describe('transformCss', () => {
...
+      }
+      @scope (.scope) {
+        .otherClass {
+          color: red;
+        }
+      }
+    `);
+  });
+
+  it('should handle complexe @scope query', () => {
```
askoufis:
```suggestion
  it('should handle complex @scope query', () => {
```

.changeset/thirty-suns-show.md
```diff
@@ -0,0 +1,18 @@
+---
+'@vanilla-extract/css': minor
+---
+
+`style`: Add support for `@scope` rules
```
askoufis: `globalStyle` should also support `@scope` now too. Worth adding that to the changeset.

packages/css/src/transformCss.ts
cahnory: The regex is there to avoid replacing & characters that appear inside quoted strings.

For example, if a selector contains something like `[data-value="foo&bar"]`, a plain `selector.replace(/&/g, root.selector)` would also replace the _&_ inside `"foo&bar"`, which is not what we want.

I do see a fix to make though. It should probably use:
`/(?<=^(?:[^"']*|"(?:\\.|[^"\\])*"|'(?:\\.|[^'\\])*')*)&/g`

That way it only matches & outside quoted strings, while still handling escaped quotes correctly.

I should probably add a few tests for that as well.

You can play with a few examples here:
https://regex101.com/?regex=%28%3F%3C%3D%5E%28%3F%3A%5B%5E%22%5C%27%5D*%7C%22%28%3F%3A%5C%5C.%7C%5B%5E%22%5C%5C%5D%29*%22%7C%5C%27%28%3F%3A%5C%5C.%7C%5B%5E%27%5C%5C%5D%29*%5C%27%29*%29%26&testString=%26%3Abefore%2C+%26%3Aafter%0A%26%5Btitle%3D%22this+%26+that%22%5D%2C+%26%3Aafter%0A%26%5Btitle%3D%22this+%27%26%27+that%22%5D%2C+%26%3Aafter%0A%26%5Btitle%3D%22this+%5C%22%26%5C%22+that%22%5D%2C+%26%3Aafter%0A%26%5Btitle%3D%22thi%5Cs+%5C%5C%26+that%22%5D%2C+%26%3Aafter%0A%0A%26%3Abefore%2C+%26%3Aafter%0A%26%5Btitle%3D%27this+%26+that%27%5D%2C+%26%3Aafter%0A%26%5Btitle%3D%27this+%22%26%22+that%27%5D%2C+%26%3Aafter%0A%26%5Btitle%3D%27this+%5C%27%26%5C%27+that%27%5D%2C+%26%3Aafter%0A%26%5Btitle%3D%27thi%5Cs+%5C%5C%26+that%27%5D%2C+%26%3Aafter&flags=gm&flavor=javascript&delimiter=%2F

packages/sprinkles/src/index.ts
cahnory: I’ve addressed your other requests locally. Since I’m not using sprinkles, this one will take me a bit more time—I’ll try to get it done quickly!

askoufis (approved):

PR #1730, opened 2026-04-28
fix(css): preserve empty string as a valid fallbackVar fallback

DoriAlagem:
## Problem

`fallbackVar('var(--foo)', '')` returns `'var(--foo)'` — the empty string fallback is silently discarded.

This breaks patterns where an empty CSS custom property fallback is intentional, e.g. when using Tailwind utility classes that rely on `var(--ring-color, )` producing a blank value rather than inheriting a default.

## Root cause

The implementation uses an empty string `''` as the sentinel to detect "not yet initialised":

```ts
let finalValue = '';

values.reverse().forEach((value) => {
  if (finalValue === '') {     // ← also true when the last value is ''
    finalValue = String(value);
  } else { ... }
});
```

When the last argument is `''`, the first loop iteration correctly sets `finalValue = ''`. But on the second iteration, `finalValue === ''` is still `true`, so the next variable is treated as the initialiser instead of being wrapped — the empty string fallback is lost.

## Fix

Switch the sentinel to `null`:

```ts
let finalValue: string | null = null;

values.reverse().forEach((value) => {
  if (finalValue === null) {   // unambiguously "not yet initialised"
    finalValue = String(value);
  } else { ... }
});
```

A test covering the empty-string case is included.

## Before / After

```ts
// Before
fallbackVar('var(--foo-bar)', '')  // => 'var(--foo-bar)'   ← wrong

// After
fallbackVar('var(--foo-bar)', '')  // => 'var(--foo-bar, )' ← correct
```

### DIFF
--- a/.changeset/fix-fallbackvar-empty-string.md
+++ b/.changeset/fix-fallbackvar-empty-string.md
@@ -1,0 +1,5 @@
+---
+"@vanilla-extract/css": patch
+---
+
+Fix `fallbackVar()` discarding empty string `""` as a valid CSS fallback value.
--- a/packages/css/src/vars.test.ts
+++ b/packages/css/src/vars.test.ts
@@ -45,6 +45,12 @@ describe('fallbackVar', () => {
     );
   });
 
+  it('supports an empty string as the final fallback', () => {
+    expect(fallbackVar('var(--foo-bar)', '')).toMatchInlineSnapshot(
+      `"var(--foo-bar, )"`,
+    );
+  });
+
   it('should throw with invalid vars', () => {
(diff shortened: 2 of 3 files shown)


DoriAlagem:
Closing — repo activity has been low and no maintainer engagement. Fix is available here if needed.

askoufis:
@DORI2001 It has barely been 2 days since you opened this PR. I have a full time job, amongst other commitments, so I can't always respond to things immediately. The fact that you didn't link this to the relevant issue makes it less likely that I'll refer to it when I get around to addressing that issue.

Thanks anyway.

PR #1731, opened 2026-04-30
fix: Escape regex metacharacters in composedClassLists class names

sashank-gogula-glean:
Class names containing characters like (, ), +, ., etc. were being interpreted as regex syntax when building the substitution pattern, causing malformed matches. Escape the class list before constructing the RegExp.


askoufis (approved):
Thanks for the PR!

PR #1742, opened 2026-05-21
fix: replace dots with underscores in dev mode identifiers

### LINKED ISSUE #1489: Development mode identifiers should replace . with _
### Describe the bug

To differentiate our vanilla extract css files to scss, we have to use a prefix to `.css.ts` so that we can handle the compiled vanilla css files differently than our compiled scss files. We've gone with `.vanilla.css.ts` but that means our development mode class names now have a `.` in them, which needs to be escaped in the CSS files. This is a vector for instability that we would rather avoid, so it would be nice if:
- vanilla extract would automatically replace non-word/non-whitespace characters in the file page with _, or 
- vanilla extract exposed the classname in the `identifiers` function so we could do basically the same thing ourselves 

Right now the classname is not exposed so all our classnames are the same except for the hash, which is not helpful for understanding which styles are fighting.

### Reproduction

No repro

### System Info

```shell
System:
    OS: macOS 14.7
    CPU: (12) arm64 Apple M2 Pro
    Memory: 57.23 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 4.1.1 - ~/.nvm/versions/node/v18.12.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 129.0.6668.101
    Firefox: 131.0.2
    Safari: 17.6
  npmPackages:
    @vanilla-extract/css: ^1.16.0 => 1.16.0 
    @vanilla-extract/css-utils: ^0.1.4 => 0.1.4 
    @vanilla-extract/jest-transform: ^1.1.9 => 1.1.9 
    @vanilla-extract/webpack-plugin: ^2.3.13 => 2.3.13 
    webpack: ^5.91.0 => 5.93.0
```

### Used Package Manager

yarn

### Logs

_No response_

### Validations

ATOM00blue:
## Summary
Dev mode debug identifiers are derived from the source file name. For file names with extra extensions (e.g. `foo.vanilla.css.ts`), the generated class names contained a literal `.`, a CSS selector metacharacter that requires escaping and is a source of instability. Dots in the debug file name segment are now replaced with `_`.

- Only the debug/dev identifier path is affected; production hashes are unchanged.
- Added unit tests asserting dotted file/dir names sanitize to `_` and that generated identifiers contain no `.`.

Closes #1489


packages/css/src/getDebugFileName.ts
```diff
@@ -48,7 +48,10 @@ const _getDebugFileName = (path: string): string => {
 
   const debugFileName = file !== 'index' ? file : dir;
 
-  return debugFileName;
+  // Dots are not valid in CSS class names without escaping, so replace them
+  // with underscores to keep dev mode identifiers stable. This is relevant for
+  // file names with additional extensions, e.g. `foo.vanilla.css.ts`.
```
askoufis:
```suggestion
  // Dots are not valid in CSS class names without escaping, so replace them
  // with underscores to keep debug identifiers stable.
```

.changeset/dev-identifier-replace-dots.md
```diff
@@ -0,0 +1,7 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Replace dots with underscores in development mode identifiers
+
+Dev mode debug identifiers are derived from the source file name. For file names with additional extensions (e.g. `foo.vanilla.css.ts`), the resulting class names contained a literal `.`, which is a CSS selector metacharacter that requires escaping and is a source of instability. Dots in the debug file name segment are now replaced with `_`. This only affects the debug/dev identifier path and not production hashes.
```
askoufis:
```suggestion
Replace dots with underscores in `debug` identifiers

`debug` identifiers are derived from the source file path. For file paths with additional dots beyond the `.css.ts` extension (e.g. `foo.vanilla.css.ts`), the resulting class names contained a literal `.`, which is a CSS selector metacharacter that requires escaping and is a source of instability. Dots in debug identifiers are now replaced with `_`. This only affects `debug` identifiers and not `short` identifiers.
```

askoufis (approved):
Thanks for the PR!

PR #1748, opened 2026-06-16
fix(vite-plugin): await compiler creation in transform to fix race with concurrently-emitted entries

### LINKED ISSUE #1747: vite-plugin: transform can run before buildStart creates the compiler, leaving .css.ts untransformed (race)
### Describe the bug

`@vanilla-extract/vite-plugin`'s `transform` hook bails out when the compiler isn't ready yet:

```ts
// packages/vite-plugin/src/index.ts — transform()
if (!compiler) {
  return null; // .css.ts passes through untransformed
}
```

`compiler` is created in the plugin's **async `buildStart`** (which spins up a child vite-node compiler). Rollup normally finishes `buildStart` before `transform` runs, but when another plugin emits an **additional entry** whose module graph is transformed concurrently, a `transform` for a `.css.ts` file can run **before** `buildStart` has created the compiler. It then hits `if (!compiler) return null`, the `.css.ts` is left untransformed, its `style()`/`globalStyle()` calls remain in the output bundle, and at runtime the code throws `Styles were unable to be assigned to a file`.

This surfaces with [`@module-federation/vite`](https://github.com/module-federation/vite), which emits each **exposed** module as its own Rollup chunk. It is **non-deterministic** — roughly 3 in 10 clean builds fail in the reproduction; the rest extract CSS correctly.

I instrumented the plugin's `transform` hook to confirm the cause: failing builds call `transform()` for the `.css.ts` with `compiler === undefined`, passing builds have `compiler` set. The outcome tracks exactly with whether compiler creation won the race.

**I intend to submit a PR**: memoize compiler creation in an `ensureCompiler()` helper and `await` it from `transform` instead of returning `null`. `buildStart` still triggers it eagerly, so behaviour is unchanged in the common case. Verified in the reproduction: 27/27 clean builds extract CSS after the change (vs ~70% before), with no runtime `getFileScope` calls left in the exposed chunk. Full diff: [`proposed-fix.md`](https://github.com/smitev/vanilla-extract-mf-vite-repro/blob/main/proposed-fix.md).

### Reproduction

https://github.com/smitev/vanilla-extract-mf-vite-repro

```sh
cd remote && npm install && cd ..
./repro.sh 10   # runs N clean builds, reports extracted vs broken (runtime) counts
```

`repro.sh` does a clean build (`rm -rf dist node_modules/.vite`) each iteration, so the only variable is plugin scheduling.

### System Info

```shell
System:    macOS 26.5.1, arm64
Binaries:  Node v24.16.0, npm 11.13.0
npmPackages:
  @vanilla-extract/css: 1.20.1
  @vanilla-extract/vite-plugin: 5.2.2
  @vanilla-extract/compiler: 0.7.0
  @vanilla-extract/integration: 8.0.10
  vite: 7.3.5
  @module-federation/vite: 1.16.8
  @vitejs/plugin-react: 5.1.4
  react: 19.2.7
```

### Used Package Manager

npm

### Logs

```shell
Error: Styles were unable to be assigned to a file. This is generally caused by one of the following:

- You may have created styles outside of a '.css.ts' context
- You may have incorrect configuration. See https://vanilla-extract.style/documentation/getting-started
    at getFileScope (.../assets/Widget-<hash>.js)
    at generateIdentifier (...)
    at style (...)
```

smitev:
Fixes #1747

## Problem

`@vanilla-extract/vite-plugin` creates its compiler in an **async `buildStart`**, and `transform` bails out if the compiler isn't ready yet:

```ts
if (!compiler) {
  return null; // .css.ts passes through untransformed
}
```

Rollup normally finishes `buildStart` before `transform`, but when another plugin emits an **additional entry** whose module graph is transformed concurrently, a `transform` for a `.css.ts` can run **before** `buildStart` has created the compiler. It then returns `null`, the `.css.ts` is left untransformed, its `style()`/`globalStyle()` calls remain in the output, and at runtime the code throws `Styles were unable to be assigned to a file`.

This reproduces with [`@module-federation/vite`](https://github.com/module-federation/vite), which emits each **exposed** module as its own Rollup chunk. It is **non-deterministic** — roughly 3 in 10 clean builds fail.

Reproduction: https://github.com/smitev/vanilla-extract-mf-vite-repro (`./repro.sh 10`).

## Fix

Memoize compiler creation in an `ensureCompiler()` helper and `await` it from `transform` instead of returning `null`. `buildStart` still triggers it eagerly, so behaviour is unchanged in the common case — the change only matters when `transform` wins the race. Because `transform` no longer depends on `buildStart` having already resolved, the outcome is deterministic regardless of plugin ordering or which plugin emitted the entry.

## Verification

Using this branch built via `pnpm build` and dropped into the reproduction:

- **Before:** ~7/10 clean builds extract CSS, ~3/10 leave runtime `style()` calls.
- **After:** 15/15 clean builds extract CSS to a static file, with no runtime `getFileScope` calls in the exposed chunk.

`pnpm lint:tsc` reports no new errors in `packages/vite-plugin/src/index.ts` (the two pre-existing "Cannot find module" errors on the import lines are present on `master` too and require a prior `pnpm build`). `prettier --check` passes on the changed files. A changeset is included.

### DIFF
--- a/.changeset/vite-plugin-compiler-race.md
+++ b/.changeset/vite-plugin-compiler-race.md
@@ -1,0 +1,9 @@
+---
+'@vanilla-extract/vite-plugin': patch
+---
+
+Fix a race condition that could leave `.css.ts` files untransformed
+
+The compiler is created in `buildStart`, and `transform` previously bailed out (`return null`) if it ran before the compiler was ready. When another plugin emits an additional entry whose module graph is transformed concurrently (e.g. a module federation plugin exposing a module as its own chunk), a `transform` for a `.css.ts` could run before `buildStart` finished creating the compiler, leaving the file untransformed and producing runtime `style()` calls that throw `Styles were unable to be assigned to a file`.
+
+Compiler creation is now memoized and awaited from `transform`, so it no longer depends on `buildStart` having already resolved.
--- a/.changeset/vite-plugin-compiler-race.md
+++ b/.changeset/vite-plugin-compiler-race.md
@@ -2,8 +2,6 @@
 '@vanilla-extract/vite-plugin': patch
 ---
 
-Fix a race condition that could leave `.css.ts` files untransformed
+Fixed an issue where `.css.ts` files could be left untransformed
 
-The compiler is created in `buildStart`, and `transform` previously bailed out (`return null`) if it ran before the compiler was ready. When another plugin emits an additional entry whose module graph is transformed concurrently (e.g. a module federation plugin exposing a module as its own chunk), a `transform` for a `.css.ts` could run before `buildStart` finished creating the compiler, leaving the file untransformed and producing runtime `style()` calls that throw `Styles were unable to be assigned to a file`.
-
-Compiler creation is now memoized and awaited from `transform`, so it no longer depends on `buildStart` having already resolved.
+This was most commonly hit when using module federation plugins that expose modules as their own chunks, resulting in untransformed runtime `style()` calls throwing `Styles were unable to be assigned to a file`.
--- a/packages/vite-plugin/src/index.ts
+++ b/packages/vite-plugin/src/index.ts
@@ -67,6 +67,7 @@ export function vanillaExtractPlugin({
   let server: ViteDevServer;
   let packageName: string;
   let compiler: Compiler | undefined;
+  let compilerReady: Promise<void> | undefined;
   let isBuild: boolean;
   const vitePromise = import('vite');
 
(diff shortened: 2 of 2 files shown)


smitev:
@askoufis I was able to track the issue with vanilla-extract used in vite module-federation setup.
can you check this fix?

askoufis:
I refactored the IIFE (just personal preference) and updated the changeset. Thanks for the PR!

askoufis (approved):

PR #1754, opened 2026-06-24
fix(rollup-plugin): use basename for emitFile name to avoid rolldown rejecting relative paths

### LINKED ISSUE #1715: rollup-plugin: emitFile rejects relative paths in name field with rolldown 1.0.0-rc.12
## Description

The `renderChunk` hook in `@vanilla-extract/rollup-plugin` passes `moduleInfo.id` directly as the `name` field in `this.emitFile()`:

https://github.com/vanilla-extract-css/vanilla-extract/blob/master/packages/rollup-plugin/src/plugin.ts#L155

```js
const assetId = this.emitFile({
  type: 'asset',
  name: moduleInfo.id,  // can be a relative path like "../styles/src/focusRing.css.ts.vanilla.css"
  source: moduleInfo.meta.css
});
```

When a `.css.ts` file imports from another package (e.g. `@vanilla-extract/rollup-plugin` processing a component that imports tokens from a shared styles package), `moduleInfo.id` resolves to a relative path like `../styles/src/focusRing.css.ts.vanilla.css`.

This worked with rollup/older rolldown versions, but **rolldown 1.0.0-rc.12** (shipped with tsdown 0.21.x) now validates that `name` is neither an absolute nor relative path:

```
Error: The "fileName" or "name" properties of emitted chunks and assets must be strings
that are neither absolute nor relative paths, received "../styles/src/focusRing.css.ts.vanilla.css".
```

## Suggested Fix

Use `path.basename(moduleInfo.id)` instead of `moduleInfo.id` for the `name` field:

```js
const assetId = this.emitFile({
  type: 'asset',
  name: path.basename(moduleInfo.id),
  source: moduleInfo.meta.css
});
```

The `name` field is only used for generating the output filename (via `assetFileNames`), so stripping the directory prefix should be safe.

## Reproduction

1. Create a package with a `.css.ts` file that imports from another package's `.css.ts` file
2. Build with tsdown >= 0.21.0 (which bundles rolldown 1.0.0-rc.12)
3. Build fails with the error above

## Versions

- `@vanilla-extract/rollup-plugin`: 1.5.1 (also reproducible on 1.5.3)
- `tsdown`: 0.21.7
- `rolldown`: 1.0.0-rc.12

smitev:
## Summary

When a `.css.ts` file imports from another package, `moduleInfo.id` in `renderChunk` resolves to a relative path like `../thirdparty-dep/styles.css.ts.vanilla.css`. This was passed directly as the `name` field to `emitFile`. Recent rolldown versions (rc.12+) reject relative or absolute paths in `name`, throwing:

```
Error: The "fileName" or "name" properties of emitted chunks and assets must be strings
that are neither absolute nor relative paths, received "../thirdparty-dep/styles.css.ts.vanilla.css".
```

Stripping `moduleInfo.id` to its basename keeps the resolved filename identical (via `assetFileNames`) and avoids the rejected prefix.

Closes #1715

## Test plan

- [x] Added a regression test under the existing rolldown describe block that builds the `@fixtures/thirdparty` fixture (which imports `.css.ts` across package boundaries) and asserts emitted CSS asset names contain no `/` separator. Verified the test fails on master and passes with this fix.
- [x] All 16 `packages/rollup-plugin` tests pass.
- [x] Existing rolldown snapshots updated to reflect basenames replacing prefixed paths in emitted asset filenames (the `src/` prefix dropped).

### DIFF
--- a/.changeset/rollup-plugin-emitfile-relative-path.md
+++ b/.changeset/rollup-plugin-emitfile-relative-path.md
@@ -1,0 +1,7 @@
+---
+'@vanilla-extract/rollup-plugin': patch
+---
+
+Fixed an issue where builds using rolldown could fail when a `.css.ts` file imports across packages
+
+`renderChunk` passed `moduleInfo.id` directly as the `name` for `emitFile`, which can be a relative path (e.g. `../styles/foo.css.ts.vanilla.css`) for cross-package imports. Recent rolldown versions reject relative or absolute paths in the `name` field, throwing `The "fileName" or "name" properties of emitted chunks and assets must be strings that are neither absolute nor relative paths`. The `name` is now stripped to its basename, which rolldown accepts and which still produces the same final asset filename via `assetFileNames`.
--- a/packages/rollup-plugin/src/index.ts
+++ b/packages/rollup-plugin/src/index.ts
@@ -16,7 +16,7 @@ import {
   tryGetPackageName,
 } from './lib';
 
-const { relative, normalize, dirname } = posix;
+const { relative, normalize, dirname, basename } = posix;
 
 export interface Options {
   /**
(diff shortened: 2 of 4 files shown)


askoufis:
Closing in favour of https://github.com/vanilla-extract-css/vanilla-extract/pull/1767.

PR #1758, opened 2026-06-29
fix: treat boxOrdinalGroup as a unitless property

mahirhir:
`boxOrdinalGroup` is missing from the `UNITLESS` map in `transformCss`, even though its two siblings from the same legacy flexbox box model, `boxFlex` and `boxFlexGroup`, are both there.

Because of that, a numeric value gets a `px` suffix it shouldn't:

```ts
style({ boxOrdinalGroup: 2 });
// box-ordinal-group: 2px;   <- invalid, the browser drops the declaration
```

`box-ordinal-group` takes an `<integer>`, so the output should be `box-ordinal-group: 2`. csstype types it as a number-accepting property (`Property.BoxOrdinalGroup = Globals | number | string`), the same as `boxFlex` and `boxFlexGroup`, so it is reachable from the public `CSSProperties` type.

Added `boxOrdinalGroup: true` next to the other two and extended the existing "should cast property values to pixels when relevant" test to cover it.

### DIFF
--- a/.changeset/unitless-box-ordinal-group.md
+++ b/.changeset/unitless-box-ordinal-group.md
@@ -1,0 +1,5 @@
+---
+'@vanilla-extract/css': patch
+---
+
+Add `boxOrdinalGroup` to the list of unitless properties so numeric values are no longer suffixed with `px`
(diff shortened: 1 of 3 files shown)


askoufis (approved):
Thanks!

Issue #1574, opened 2025-04-30
'Invalid exports' in vite-plugin-vanilla-extract when loading external library containing style files with file extensions `.css.ts`


tkgroot:
### Describe the bug

vite-plugin-vanilla-extract in an application context in combination with a "third-party" library using web-components with LitElement throws the `invalid exports` error when the component loads its own styles from an separate file, named `styles.css.ts` which clashes with the `.css.ts` files from vanilla-extract.

```
[vite-plugin-vanilla-extract] Invalid exports.

You can only export plain objects, arrays, strings, numbers and null/undefined.
```

Creating a style file containing the styles for the web-component called `styles.css.ts` within the component library for the web-component in `MyCustomElement.ts` will produce the issue in the application.

Adding an additional option for the vanillaExtractPlugin which offers to set the file extension which is used for files which contains "vanilla-extract" content would most likely solve this issue.

```js
vanillaExtractPlugin({
  fileExtension: '.css.ts',
});
```

### Reproduction

https://github.com/tkgroot/vanilla-extract-issue/tree/master

### System Info

```shell
System:
    OS: macOS 13.6.4
    CPU: (10) arm64 Apple M1 Max
    Memory: 127.14 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.5.1 - ~/.nvm/versions/node/v22.5.1/bin/node
    npm: 10.9.0 - ~/.nvm/versions/node/v22.5.1/bin/npm
    pnpm: 10.4.1 - ~/.nvm/versions/node/v22.5.1/bin/pnpm
  Browsers:
    Chrome: 132.0.6834.110
    Safari: 17.5
```

### Used Package Manager

npm

### Logs

```shell

```

### Validations

- [x] Check that there isn't [already an issue](https://github.com/vanilla-extract-css/vanilla-extract/issues) that reports the same bug to avoid creating a duplicate.
- [x] The provided reproduction is a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) of the bug.

jimthedev:
I am running into this need too but in a bit different context.  

[...]

PR #1760, opened 2026-07-01
fix: add exclude

jimthedev:
## Problem

When consuming libraries that publish pre-compiled vanilla-extract CSS (`.css.js`/`.css.mjs` files), the vite plugin attempts to reprocess these files during the consumer's build. This works fine for simple exports like strings from `style()`, but fails for libraries that export **recipe functions** from `@vanilla-extract/recipes`.

Recipe functions use `createRuntimeFn` which returns a function—not serializable data. When the plugin tries to serialize these exports, it throws:

```
Error: Invalid exports.
```

This is a breaking issue for design system libraries that publish pre-compiled vanilla-extract CSS with recipe exports. Consumers have no way to exclude these files from reprocessing.

## Solution

Add `include` and `exclude` options to the vite plugin using `@rollup/pluginutils` (the standard approach used by most Vite/Rollup plugins). These options filter which files pass through the `transform` hook.

```ts
vanillaExtractPlugin({
  exclude: [/node_modules\/my-design-system/]
})
```

### Why not fix the serialization instead?

Recipe functions are intentionally non-serializable—they're runtime code that maps variant props to class names. The pre-compiled `.css.js` file already contains the correct output; reprocessing it is unnecessary and causes this failure. The `exclude` option lets consumers tell the plugin "this file is already compiled, skip it."

## Changes

- Add `@rollup/pluginutils` dependency for `createFilter`
- Add `include`/`exclude` options to the `Options` interface  
- Apply filter in `transform` hook before processing
- Add documentation to `site/docs/integrations/vite.md`
- Add test fixture with non-serializable recipe export
- Add unit tests and integration test

## Testing

The new `precompiled-exclude` fixture contains a simulated pre-compiled library with a recipe export. Tests verify:

1. Build **fails** without the exclude option (confirming the problem exists)
2. Build **succeeds** with `exclude: ['**/precompiled-lib/**']`

Also tested successfully in a real production app consuming a design system library with recipe exports.

While this works. I think one other way to do this is for files to be marked as already compiled at the top of them and then you wouldn't need such a thing as your consumers could just filter out already processed files. In that world you'd still need to read the files, so it would perhaps be less performant. Open to ideas and feedback. Relates to #1574

### DIFF
--- a/.changeset/add-include-exclude-options.md
+++ b/.changeset/add-include-exclude-options.md
@@ -1,0 +1,20 @@
+---
+'@vanilla-extract/vite-plugin': minor
+---
+
+Add `include` and `exclude` options to control which files the plugin processes
+
+These options use the same filter syntax as `@rollup/pluginutils` and allow you to:
+
+- Skip pre-compiled `.css.js` files from libraries that export non-serializable values (like recipe functions)
+- Restrict processing to specific directories
+
+Example:
+
+```ts
+vanillaExtractPlugin({
+  exclude: [/node_modules\/my-design-system/]
+});
+```
+
+This is particularly useful when consuming design system libraries that publish pre-compiled vanilla-extract CSS, which would otherwise cause "Invalid exports" errors during build.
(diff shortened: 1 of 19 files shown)


jimthedev:
Closing in favor of #1763

PR #1763, opened 2026-07-02
fix: add fileExtension, DEFAULT_FILE_EXTENSIONS

### LINKED ISSUE #1574: 'Invalid exports' in vite-plugin-vanilla-extract when loading external library containing style files with file extensions `.css.ts`

jimthedev:
## Problem

When consuming pre-compiled vanilla-extract libraries that publish `.css.js` files, the vite plugin would reprocess those library files unnecessarily. Users needed a way to use a different extension (like `.ve.ts`) for their source files so only their code gets processed.

## Solution

Added a `fileExtension` option to `@vanilla-extract/vite-plugin` that lets users specify exactly which file extensions should be processed.

### `@vanilla-extract/integration`

- Added `DEFAULT_FILE_EXTENSIONS` constant — a readonly array of the 6 standard extensions:
  - `.css.ts`, `.css.tsx`, `.css.js`, `.css.jsx`, `.css.mjs`, `.css.cjs`
- Refactored `cssFileFilter` to be **derived** from `DEFAULT_FILE_EXTENSIONS` rather than a separate hand-written regex. This prevents the two from drifting out of sync.

### `@vanilla-extract/vite-plugin`

- Added `fileExtension` option accepting a string or array of strings
- Re-exports `DEFAULT_FILE_EXTENSIONS` from integration for convenience
- When `fileExtension` is provided, only files matching those exact extensions are processed
- When omitted, behavior is unchanged (uses `cssFileFilter` from integration)

### Replace default extensions entirely

```ts
import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';

export default {
  plugins: [
    vanillaExtractPlugin({
      // Only process .ve.ts and .ve.tsx files
      fileExtension: ['.ve.ts', '.ve.tsx'],
    }),
  ],
};
```

### Keep defaults and add custom extensions

```ts
import { vanillaExtractPlugin, DEFAULT_FILE_EXTENSIONS } from '@vanilla-extract/vite-plugin';

export default {
  plugins: [
    vanillaExtractPlugin({
      // Process all default extensions PLUS .ve.ts and .ve.tsx
      fileExtension: [...DEFAULT_FILE_EXTENSIONS, '.ve.ts', '.ve.tsx'],
    }),
  ],
};
```

Fixes #1574

### DIFF
--- a/.changeset/full-file-extension-option.md
+++ b/.changeset/full-file-extension-option.md
@@ -1,0 +1,40 @@
+---
+"@vanilla-extract/vite-plugin": minor
+"@vanilla-extract/integration": minor
+---
+
+Add `fileExtension` option to vite-plugin and export `DEFAULT_FILE_EXTENSIONS` constant from integration.
+
+`DEFAULT_FILE_EXTENSIONS` is now the single source of truth for which file extensions vanilla-extract processes. The `cssFileFilter` regex is derived from this array, preventing drift between them.
+
+This allows specifying exact file extensions (e.g., `.ve.ts`) for vanilla-extract files instead of the default `.css.ts`, `.css.tsx`, etc. This is useful when consuming pre-compiled vanilla-extract libraries that publish `.css.js` files - you can use a different extension for your source files to avoid reprocessing library files.
+
+Replace default extensions with custom ones:
+
+```ts
+import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
+
+export default {
+  plugins: [
+    vanillaExtractPlugin({
+      // Process only .ve.ts and .ve.tsx files (NOT .css.ts, etc.)
+      fileExtension: ['.ve.ts', '.ve.tsx'],
+    }),
+  ],
+};
+```
+
+Or keep the defaults while adding custom extensions using `DEFAULT_FILE_EXTENSIONS`:
+
+```ts
+import { vanillaExtractPlugin, DEFAULT_FILE_EXTENSIONS } from '@vanilla-extract/vite-plugin';
+
+export default {
+  plugins: [
+    vanillaExtractPlugin({
+      // Process default extensions PLUS .ve.ts and .ve.tsx
+      fileExtension: [...DEFAULT_FILE_EXTENSIONS, '.ve.ts', '.ve.tsx'],
+    }),
+  ],
+};
+```
--- a/.changeset/full-file-extension-option.md
+++ b/.changeset/full-file-extension-option.md
@@ -1,6 +1,7 @@
 ---
 "@vanilla-extract/vite-plugin": minor
 "@vanilla-extract/integration": minor
+"@vanilla-extract/compiler": minor
 ---
 
 Add `fileExtension` option to vite-plugin and export `DEFAULT_FILE_EXTENSIONS` constant from integration.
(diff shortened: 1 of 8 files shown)


jimthedev:
@askoufis I think this is ready for at least a prelim look

o-alexandrov:
Please consider to merge #1773 instead: no new abstraction, one `cssFileFilter?: RegExp` option threaded through plugin + compiler, defaults to the **existing** filter

askoufis:
> Please consider to merge #1773 instead: no new abstraction, one `cssFileFilter?: RegExp` option threaded through plugin + compiler, defaults to the **existing** filter

@o-alexandrov I'm quite confused by your PR. I don't quite understand the practical difference between a `css.ts` and `*.css.ts` extension. Are they not the same? If you could present a better explanation of your problem (potentially in a separate issue if it's different to #1574) I'd be happy to help troubleshoot.

Just to make it quite clear, I'm strongly against allowing customization of the file filter/file extension unless absolutely necessary, which is why I'm trying to help @jimthedev with his issue _without_ exposing an API for customizing the file extension (if possible).

o-alexandrov:
Thank you for maintaining this project, and the willingness to come up with the best way moving forward.

---

@askoufis no, not the same. Letting to modify the pattern is a natural decision that also every similar plugin has.

Currently, we have to add a prefix, ex. “button.” to make it “button.css.ts”. Just “css.ts” won’t be used by the plugin because of “*.” before the “css.ts” in the file pattern.

It’s an unnecessary mental effort (waste of developers’ time) to require adding a prefix with zero value. A simpler option is to require always naming “css.ts” and then it results in a necessity to group all components in separate directories which is extremely popular.

Just “css.ts” also avoids another waste of time, when a component is renamed. As prefix typically has naming related to the name of the component.

---

side note:
This PR is 600 lines and increases API scope the plugin needs to maintain, while offering less flexibility. That PR is 80 lines

jimthedev:
@askoufis thanks for the info. Apologies for the delay, I've been out. I'll compare what you linked to our set up and get back to you.

Issue #590, opened 2022-02-14
createSprinkles with multiple collections that have the same property throws an error


moroshko:
### Describe the bug

Passing multiple collections of properties to `createSprinkles` that share the same property (`backgroundColor` in this case) results in an error:

**sprinkles.css.ts**
```ts
import { defineProperties, createSprinkles } from "@vanilla-extract/sprinkles";

const responsiveProperties = defineProperties({
  conditions: {
    mobile: {},
    tablet: { "@media": "(min-width: 768px)" }
  },
  defaultCondition: "mobile",
  properties: {
    backgroundColor: ["red", "blue"]
  }
});

const stateProperties = defineProperties({
  conditions: {
    default: {},
    focus: { selector: "&:focus" }
  },
  defaultCondition: "default",
  properties: {
    backgroundColor: ["red", "blue"]
  }
});

export const sprinkles = createSprinkles(responsiveProperties, stateProperties);
```

**App.css.ts**
```ts
import { sprinkles } from "./sprinkles.css";

export const container = sprinkles({
  // backgroundColor: { default: "red", focus: "blue" } // works, but TypeScript complains
  backgroundColor: { mobile: "red", tablet: "blue" } // doesn't work, runtime error
});

```

```
ERROR in ./src/App.css.ts
Module build failed (from ./node_modules/@vanilla-extract/webpack-plugin/loader/dist/vanilla-extract-webpack-plugin-loader.cjs.js):
NonErrorEmittedError: (Emitted value instead of an instance of Error) SprinklesError: "backgroundColor" has no condition named "mobile". Possible values are "default", "focus"
    at processResult (/sandbox/node_modules/webpack/lib/NormalModule.js:673:12)
    at /sandbox/node_modules/webpack/lib/NormalModule.js:778:5
    at /sandbox/node_modules/loader-runner/lib/LoaderRunner.js:399:11
    at /sandbox/node_modules/loader-runner/lib/LoaderRunner.js:195:20
    at context.callback (/sandbox/node_modules/loader-runner/lib/LoaderRunner.js:124:13)
    at /sandbox/node_modules/@vanilla-extract/webpack-plugin/loader/dist/vanilla-extract-webpack-plugin-loader.cjs.dev.js:77:5
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

### Link to reproduction

[CodeSandbox](https://codesandbox.io/s/sprinkles-mb3k2?file=/src/App.css.ts)

[...]

PR #1764, opened 2026-07-04
fix(sprinkles): reject duplicate property definitions

cyphercodes:
## Summary
- detect duplicate sprinkles property/shorthand names across `createSprinkles` property collections before they can be overwritten
- throw a clear `SprinklesError` that explains properties must be unique across collections
- add regression coverage for duplicate `backgroundColor` definitions with different condition sets

Refs #590

## Testing
- `pnpm build`
- `pnpm test:unit tests/sprinkles/sprinkles.test.ts`
- `pnpm exec oxlint packages/sprinkles/src/createSprinkles.ts tests/sprinkles/index.css.ts tests/sprinkles/sprinkles.test.ts`
- `pnpm lint:prettier`
- `git diff --check`

Note: `pnpm lint:tsc` currently fails on pre-existing site typing/generated-docs issues (for example `site/src/docs-store.ts` cannot find `../docs-manifest.json`, plus several implicit-any errors in `site/src/DocsPage/DocsPage.tsx` and `site/src/HomePage/HomePage.tsx`).

### DIFF
--- a/packages/sprinkles/src/createSprinkles.ts
+++ b/packages/sprinkles/src/createSprinkles.ts
@@ -67,11 +67,36 @@ export type SprinklesFn<Args extends ReadonlyArray<SprinklesProperties>> = ((
   props: SprinkleProps<Args>,
 ) => string) & { properties: Set<keyof SprinkleProps<Args>> };
 
+class SprinklesError extends Error {
+  constructor(message: string) {
+    super(message);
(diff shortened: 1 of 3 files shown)


askoufis:
#590 never proposed an agreed-upon solution, so I'd rather not introduce a breaking change unless that happens. Thanks for the PR, but I'm going to close this.

PR #1783, opened 2026-08-26
Skip module graph invalidation in production builds

mpsq:
Production builds spend most of their compile time re-evaluating .css.ts files that have not changed.

The loader invalidates the shared compiler's entire module graph before every file. Dev needs that, since the file watcher is disabled and a file can change between invocations. A production build never mutates its inputs, so the invalidation only forces each file's imports to be re-evaluated from scratch.

Measured on an app with 634 .css.ts files, Next 16 and Turbopack, cold builds:

| | before | after |
| --- | --- | --- |
| time in loader | 256.3s | 83.6s |
| wall compile | 266s | 177s |

CSS output is identical: same 311 files, same 2,594,429 bytes, same content hash.

Dev is unchanged, the invalidation still runs there. The stress test in index.test.ts still covers it, since vitest sets NODE_ENV to test.

Happy to gate this some other way if you would rather not read NODE_ENV in the loader, though identifiers just above it does the same.

### DIFF
--- a/.changeset/lucky-pandas-listen.md
+++ b/.changeset/lucky-pandas-listen.md
@@ -2,8 +2,4 @@
 '@vanilla-extract/turbopack-plugin': patch
 ---
 
-Stop invalidating the compiler's module graph on every file during production builds
-
-Production builds now reuse the shared compiler's module graph across files instead of dropping it before each one. On an app with 634 `.css.ts` files, this cut time spent in the loader from 256s to 84s and total Turbopack compile time by a third, with byte-identical CSS output.
-
-Development builds are unaffected: the file watcher is disabled there, so the manual invalidation is still required to pick up edits between loader invocations.
+perf: Don't invalidate the compiler's module graph each time a file is processed during production builds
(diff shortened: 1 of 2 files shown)


.changeset/lucky-pandas-listen.md
```diff
@@ -0,0 +1,9 @@
+---
+'@vanilla-extract/turbopack-plugin': patch
+---
+
+Stop invalidating the compiler's module graph on every file during production builds
+
+Production builds now reuse the shared compiler's module graph across files instead of dropping it before each one. On an app with 634 `.css.ts` files, this cut time spent in the loader from 256s to 84s and total Turbopack compile time by a third, with byte-identical CSS output.
+
+Development builds are unaffected: the file watcher is disabled there, so the manual invalidation is still required to pick up edits between loader invocations.
```
askoufis:
```suggestion
perf: Don't invalidate the compiler's module graph each time a file is processed during production builds
```

askoufis (approved):
Thanks for the PR!

askoufis:
@mpsq This fix was released in https://github.com/vanilla-extract-css/vanilla-extract/releases/tag/%40vanilla-extract%2Fturbopack-plugin%400.1.4.

mpsq:
Thank you!

PR #1790, opened 2026-09-23
fix(css): upgrade modern-ahocorasick to v3.3.0

sonofmagic:
## Summary

Upgrade `@vanilla-extract/css` from the `modern-ahocorasick` v1 API to `modern-ahocorasick@3.3.0`.

The v3 matcher returns independent match objects with UTF-16 half-open ranges instead of the former grouped tuple shape. This PR migrates selector replacement to that API while preserving vanilla-extract's overlap and longest-match behavior.

## Changes

- Update `modern-ahocorasick` to `^3.3.0` and refresh the lockfile.
- Consume `{ start, end, pattern }` match objects using exclusive UTF-16 end offsets.
- Preserve the existing selector semantics by selecting the longest match at each end position.
- Correct overlap handling so adjacent class names are still replaced.
- Document the Node.js 16.0.0+ runtime requirement inherited from `Intl.Segmenter`.
- Add regression coverage for UTF-16 ranges, overlapping and adjacent class names, and emoji.
- Add a patch changeset for `@vanilla-extract/css`.

## Performance report

Benchmarks compare the same `transformCss` workload on Apple Silicon with Node 24.11.1/25.x. Each run used 2,048 class names and 2,500 selectors, started in a fresh process, warmed up the code, ran 30 measured iterations, and verified identical CSS output.

| Comparison | Result |
| --- | ---: |
| v1.x → v3.3.0, complete `transformCss` | 19.87 → 16.78 ms/op, about 15.6% faster |
| v1.x → v3.3.0, repeated Node 24 runs | about 15–27% faster on the large-dictionary workload |
| Pre-built matcher, `search()` only | 63.3 → 61.2 µs/op, about 3.5% faster |
| Matcher construction + search | about 2.66x faster |

The main v1-to-v3 gain comes from faster automaton construction and lower retained matcher memory. For 2,048 patterns, retained matcher heap dropped from roughly 3.3–4.5 MB to about 0.1 MB per matcher. Small dictionaries (32–512 patterns) were between roughly 4% and 13% slower in the tested transform workloads, and short selector searches can also be slower because v3 creates independent match objects.

The results do not imply a universal speedup for every consumer. The improvement is most visible when matcher construction dominates.

## Runtime and package-size considerations

- The original `@vanilla-extract/css@1.21.2` package does not declare an `engines.node` minimum; the repository history targets Node.js 14 compatibility for its generated tooling output.
- v3 uses ES2022 features and `Intl.Segmenter` for grapheme-aware matching. Node.js support therefore starts at 16.0.0, where `Intl.Segmenter` is available.
- Removing the fallback narrows the effective runtime baseline from the historical Node.js 14-compatible range to Node.js 16.0.0+.
- The v3.3.0 npm tarball is about 90 KB versus about 3.3 KB for the v1 package. This includes the additional runtime and optional entrypoints.

## Validation

- `modern-ahocorasick@latest` resolves to `3.3.0`.
- Focused migration test: 1 passed, covering UTF-16 ranges, overlap behavior, adjacent matches, and emoji.
- Benchmark CSS output equality: passed for all compared workloads.
- The repository CI should run the full unit, type, build, and package checks before merge.


tony-scio:
FYI, this was a hotspot for a large private repo I'm working on. https://github.com/vanilla-extract-css/vanilla-extract/pull/1791 is a complimentary optimization here. Should be interesting to see how things add up with both

